PR #3, opened 2026-04-12
Accept optional focus_topic during compression

Tosko4:
This makes the engine accept an optional `focus_topic` hint during compression and pass it through the summarization path.

Why this helps:
- callers can use guided/manual compression without tripping over a signature mismatch
- the hint now reaches both leaf summarization and condensation
- existing callers keep working because the new argument stays optional

Tests:
- added coverage for `focus_topic` reaching the summarizer

This is intentionally small and backwards-compatible.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -117,7 +117,8 @@ class LCMEngine(ContextEngine):
         return rough >= self.threshold_tokens
 
     def compress(self, messages: List[Dict[str, Any]],
-                 current_tokens: int = None) -> List[Dict[str, Any]]:
+                 current_tokens: int = None,
+                 focus_topic: Optional[str] = None) -> List[Dict[str, Any]]:
         """Main compaction entry point.
 
         1. Ingest any new messages into the store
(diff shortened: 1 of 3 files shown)


Tosko4:

I am linking it here because Hermes already has guided and manual compression paths that can pass a focus hint, so this upstream fix helps keep that compatibility path clean.

stephenschoettler (approved):
## Review: Approve

Clean, minimal, backwards-compatible fix for issue #5. `focus_topic` is correctly threaded through `compress()` → `_maybe_condense()` → `summarize_with_escalation()` → L1/L2 prompts. Test verifies the value reaches the summarizer.

### Minor notes

1. **L1 prompt ordering** — `{focus_guidance}` is placed *after* the "End with: Expand for details about:" instruction. Moving it *before* that line would keep the expand hint as the last instruction the LLM sees, which may improve compliance. Cosmetic, not a blocker.

2. **No test for condensation path** — The test covers leaf compression (step 4) but not `_maybe_condense` (step 6) receiving `focus_topic`. Harder to trigger in tests since it requires hitting `condensation_fanin`, so acceptable for now.

3. **Extra blank line when `focus_topic` is empty** — `focus_guidance` resolves to `""`, leaving a double blank line in the prompt. Harmless but worth noting.

LGTM — ready to merge. Fixes #5.

PR #6, opened 2026-04-12
Bind tool calls to the active engine session

Tosko4:
This removes the global tool-engine singleton and routes tool calls through the active engine instance instead.

What changed:
- `handle_tool_call()` now passes `engine=self` into the tool handlers
- tool handlers use the provided engine instance instead of shared global state
- `lcm_describe` and `lcm_expand` now reject nodes outside the current session
- overview output no longer assumes depths are contiguous

Why this matters:
- long-lived processes can have more than one active engine/session over time
- global tool state makes cross-session leaks a lot more likely
- node inspection/expansion should stay scoped to the active session

Tests:
- added coverage for per-engine tool dispatch
- added coverage for session-scoped describe/expand
- added coverage for sparse high-depth overview output

This should make the plugin behave much better inside Hermes gateway-style runtimes.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -64,9 +64,6 @@ class LCMEngine(ContextEngine):
         # genuinely new messages (appended after compaction) get ingested.
         self._ingest_cursor: int = 0
 
-        # Wire tool handlers
-        lcm_tools.set_engine(self)
-
         # State required by ContextEngine ABC and run_agent.py compatibility
         self.model = ""
         self.base_url = ""
@@ -263,7 +260,7 @@ class LCMEngine(ContextEngine):
         }
         handler = handlers.get(name)
         if handler:
-            return handler(args)
+            return handler(args, engine=self)
         return json.dumps({"error": f"Unknown LCM tool: {name}"})
 
     def get_status(self) -> Dict[str, Any]:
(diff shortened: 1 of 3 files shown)


Tosko4:

I am linking it here because this session scoping fix matters for longer lived Hermes runtimes where more than one session can exist over time and tool state needs to stay bound to the active engine instance.

stephenschoettler (approved):
## Review: Approve

Solid fix that eliminates the global `_engine` singleton in `tools.py`, replacing it with explicit `engine=self` parameter passing. This prevents cross-engine contamination in multi-session/multi-engine processes.

### Highlights

- **Session-scoping added to describe/expand** — `_get_session_node()` helper correctly prevents cross-session data leaks
- **Sparse depth fix** — Old overview used `range(10)` with early `break` on empty depth, so d0+d2 (no d1) would miss d2. Now iterates actual depth set. Good catch.
- **Clean code improvements** — `elif` after `return` → `if`, descriptive variable names, consistent formatting
- **Good test coverage** — 3 new tests cover per-engine isolation, cross-session rejection, and sparse depth

### Minor notes

1. **`_require_engine` naming** — Returns `None` instead of raising, which is misleading for a "require" function. `_get_engine` would be more accurate. Not a blocker.

2. **Overview memory usage** — `get_session_nodes(session_id)` now fetches all nodes into memory (old code used `limit=100` per depth). Fine for current DAG sizes but worth revisiting if they grow large.

LGTM — ready to merge.

PR #7, opened 2026-04-12
Keep source mapping stable across repeated content and carry-over

Tosko4:
This tightens source-id mapping during compaction and adds a helper for moving retained summaries into a new session.

What changed:
- source-id matching now walks forward in store order instead of matching against any earlier duplicate content
- matching starts after the last compacted store id, so repeated content from older history can't hijack newer source rows
- added `reassign_session_nodes()` in the DAG
- added `carry_over_new_session_context()` on the engine for retained-summary handoff
- reset startup now clears the ingest/mapping cursors again

Why this matters:
- repeated content can otherwise point a later summary back at older rows
- retained summaries are a lot more useful if they can be moved into the new session cleanly

Tests:
- added repeated-content coverage for later-row mapping
- added carry-over coverage for retained nodes moving into a new session

This should make multi-compaction and session-reset flows much more predictable in Hermes-style runtimes.

### DIFF
--- a/dag.py
+++ b/dag.py
@@ -106,7 +106,7 @@ class SummaryDAG:
     def delete_below_depth(self, session_id: str, min_depth: int) -> int:
         """Delete all nodes for a session with depth < min_depth.
 
-        Returns the number of deleted nodes.  Used during session reset
+        Returns the number of deleted nodes. Used during session reset
         to retain only high-level summaries across sessions.
         """
         cur = self._conn.execute(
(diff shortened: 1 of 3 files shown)


Tosko4:

I am linking it here because this carry over and source mapping behavior lines up with the Hermes new session and reset flow, and it helps avoid drift between the host side and the plugin side.

stephenschoettler (approved):
## Review: Approve

Fixes a real bug where repeated identical messages (same role+content) get mapped to wrong store IDs. The old `_get_store_ids_for_messages()` used a content-keyed dict lookup, which breaks when the same content appears across compaction cycles.

### What changed

- **Source mapping rewrite** — Sequential forward-walk starting after `_last_compacted_store_id` instead of dict-based lookup. Correct and more predictable.
- **Cursor resets on session start** — `_ingest_cursor` and `_last_compacted_store_id` reset in `on_session_start()`. Without this, stale cursors from a previous session would corrupt mapping.
- **Carry-over support** — `reassign_session_nodes()` in DAG + `carry_over_new_session_context()` on engine for `/new` session transitions. Clean additive API.
- **Good test coverage** — `test_repeated_content_maps_to_later_store_rows` directly exercises the core bug, carry-over test validates the new path.

### Minor notes

1. **O(n*m) worst case** — Forward walk probes linearly for each message. O(n+m) amortized in practice since `store_idx` advances, but pathological cases (many non-matching synthetic messages) could be slower. Unlikely to matter at current scale.

2. **Candidates rebuilt each call** — `_get_store_ids_for_messages` fetches all session messages and filters in Python. Could push the `store_id > X` floor into the SQL query as a future optimization.

3. **`carry_over_new_session_context` not called from runtime** — Presumably integrated on the hermes-agent side in a separate PR. Worth confirming that's wired up.

LGTM — ready to merge.

PR #9, opened 2026-04-12
feat: add separate expansion path and query tool

Tosko4:
## Summary
- add a separate expansion path instead of reusing the summary-only path
- add `LCM_EXPANSION_MODEL`, `LCM_SUMMARY_TIMEOUT_MS`, and `LCM_EXPANSION_TIMEOUT_MS`
- add `lcm_expand_query` for answering a question from expanded LCM context
- keep this scoped to in-process expansion/synthesis for now, without introducing delegated sub-agent behavior yet

## Why
`lcm_expand` is useful for raw drill-down, but it is still a manual retrieval step. This adds a practical path for query-driven answer synthesis while keeping expansion configuration separate from summarization.

It also lets summary and expansion use different model/timeout settings.

## What changed
- config:
  - `expansion_model`
  - `summary_timeout_ms`
  - `expansion_timeout_ms`
- tool surface:
  - add `lcm_expand_query`
- engine:
  - expose the new tool
  - pass summary timeout into the summarization path
- docs/tests:
  - update README
  - add config + tool coverage

## Validation
Passed locally:

  - result: `54 passed`

Also dogfooded locally in Hermes using the external plugin runtime:
- verified Hermes loaded `/home/nabu/.hermes/plugins/hermes-lcm/engine.py`
- verified `lcm_expand_query` was exposed to `AIAgent`
- verified a live expansion query returned the correct answer using `LCM_EXPANSION_MODEL=gpt-5.4`
- verified this while `LCM_SUMMARY_MODEL` was intentionally set to a bogus value, to prove the answer path was not coupled to the summary model

## Non-goals
- no bundling into Hermes core
- no delegated/sub-agent expansion path in this PR
- no recursion/concurrency guard yet for delegated expansion, since delegation itself is not introduced here

Those can be layered on later if we want to get closer to `lossless-claw` parity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ into anything that was compacted.
 - **Immutable store** — every message persisted in SQLite, never modified
 - **Summary DAG** — hierarchical compaction (D0 minutes → D1 hours → D2 days)
 - **3-level escalation** — L1 detailed → L2 bullets → L3 deterministic truncate (guaranteed convergence)
-- **Agent tools** — `lcm_grep`, `lcm_describe`, `lcm_expand` for structured retrieval
+- **Agent tools** — `lcm_grep`, `lcm_describe`, `lcm_expand`, `lcm_expand_query` for structured retrieval
 - **Current-turn search** — live messages ingested before tool execution
 - **Profile-scoped** — separate DB per Hermes profile
 
@@ -94,7 +94,7 @@ Verify with `hermes plugins`:
 
 ```
 Plugins (1):
-  ✓ hermes-lcm v0.1.0 (3 tools, 0 hooks)
+  ✓ hermes-lcm v0.1.0 (4 tools, 0 hooks)
 
 Provider Plugins:
   Context Engine: lcm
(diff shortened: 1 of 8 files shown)


stephenschoettler (approved):
## Review: Approve

Adds `lcm_expand_query` — query-driven answer synthesis from expanded DAG nodes with separate model/timeout config from summarization.

Highlights:
- Expansion and summarization paths cleanly decoupled (different model, different timeout)
- Accepts either `query` (FTS search) or explicit `node_ids` — flexible
- Shared expansion logic refactored out of `lcm_expand` into `_expand_message_sources` and `_expand_child_nodes` — clean
- Test verifies model routing and timeout passthrough
- README and schemas updated

One note: `_invoke_summary_llm` in `escalation.py` uses `inspect.signature` at runtime to check if `_call_llm_for_summary` accepts `timeout`. Since both functions are in the same file and `timeout` is already added to `_call_llm_for_summary` in this PR, you could skip the introspection and just pass it directly. Minor — not a blocker.

CI green. LGTM.

PR #15, opened 2026-04-15
fix: clarify /lcm status before first session bind

Tosko4:
## Summary
- make `/lcm status` explicitly report an unbound runtime before the first session bind
- show useful database totals in that unbound state instead of misleading per-session zeros
- add a regression test covering the fresh-restart / pre-first-turn case

## Why
Right after a fresh Hermes restart, `/lcm` can run before any normal agent turn has called `on_session_start()`. In that state the command was showing `session_id: (none)`, `session_platform: (unknown)`, and `threshold_tokens: 0`, which looks broken even though the real issue is simply that no live session has bound the engine yet.

This change makes that state explicit and useful:
- `session_id: (unbound)`
- `session_platform: (unbound)`
- `threshold_tokens: (uninitialized)`
- global DB totals plus a note explaining that one normal message will populate live per-session runtime details

## Test Plan
- `python3 -m pytest tests/test_lcm_command.py -q`

### DIFF
--- a/command.py
+++ b/command.py
@@ -48,21 +48,42 @@ def _status_text(engine) -> str:
     db_path = Path(engine._store.db_path)
     db_exists = db_path.exists()
     db_size = db_path.stat().st_size if db_exists else 0
+    session_bound = bool(engine._session_id)
+
+    def _safe_scalar(conn, query: str) -> int | str:
+        try:
+            return int(conn.execute(query).fetchone()[0])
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Clean UX fix. Showing global DB stats and '(unbound)' labels before session bind is much better than misleading zeros. Test covers the new path.

PR #16, opened 2026-04-15
fix: add overflow recovery for assembly guardrails

Tosko4:
## Summary
- add forced overflow recovery when assembly-cap pressure is hit instead of effectively no-oping below the normal leaf threshold
- reserve observed provider overhead when deciding the effective recovery cap
- surface `overflow_recovery_failed` in engine status for irreducible cases and add regression coverage for guardrail edge cases

## Why
We already had hard assembly budget controls, but that only capped context assembly. This slice adds the actual recovery path for cases where live context pressure hits the cap and normal compaction thresholds would otherwise fail to respond usefully.

That closes the practical gap between “we know the budget is blown” and “we actively recover under budget when possible, and say so explicitly when not”.

## Validation
Local plugin validation:

Local host/runtime dogfood:
- confirmed Hermes loads the external plugin from `~/.hermes/plugins/hermes-lcm/engine.py`
- confirmed the active status surface includes `overflow_recovery_failed`

### DIFF
--- a/config.py
+++ b/config.py
@@ -7,6 +7,26 @@ def _parse_pattern_list(raw: str) -> list[str]:
     return [part.strip() for part in raw.split(",") if part.strip()]
 
 
+def _parse_int_env(key: str, default: int) -> int:
+    raw = os.environ.get(key)
+    if raw is None:
+        return default
+    try:
+        return int(raw)
+    except (TypeError, ValueError):
+        return default
+
+
+def _parse_float_env(key: str, default: float) -> float:
+    raw = os.environ.get(key)
+    if raw is None:
+        return default
+    try:
+        return float(raw)
+    except (TypeError, ValueError):
+        return default
+
+
 @dataclass
 class LCMConfig:
     """All tunables for the LCM engine."""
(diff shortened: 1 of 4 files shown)


Tosko4:
Quick heads-up: I also updated the Hermes host-support PR with a small generic lifecycle improvement for external context-engine plugins.

PR: https://github.com/NousResearch/hermes-agent/pull/8416

What changed there
- richer session-boundary hook context on `/new` / reset-style transitions
- old/new session ids plus prior-message context where the host already has it
- keeps default built-in Hermes behavior unchanged, but gives external plugins a cleaner rollover boundary contract

That host-side slice is directly relevant to the item-5/session-rollover work we’re doing here, and it should make external `hermes-lcm` integration cleaner without bundling plugin code into Hermes core.

stephenschoettler (approved):
Solid defensive work. The forced overflow recovery path handles the edge case where context exceeds assembly cap but normal compaction can't fire. Good details: provider overhead accounting, summary blob dedup to avoid doubles, irreducible overflow flagging, and the flag reset lifecycle. Config env parsing hardening is a nice bonus. 8 thorough tests.

PR #17, opened 2026-04-15
feat: add session rollover lifecycle helper

Tosko4:
## Summary
- add a small `rollover_session(...)` lifecycle helper to the LCM engine
- cover repeated `/new`-style session boundaries without duplicating retained nodes
- tighten Hermes-native session rollover behavior without pretending the engine already has full transcript bootstrap / rotation machinery

## Why
Session boundaries are exactly where long-lived context engines get weird if lifecycle handling is loose.

This PR takes the smallest useful step:
- flush old-session state
- rebind to the new session
- optionally carry retained summaries forward
- make repeated rollover boundaries deterministic

That gives Hermes LCM a cleaner rollover path right now, without overreaching into a much larger bootstrap/checkpoint design in the same slice.

## Notes
This is intentionally not a full bootstrap / checkpoint / transcript-rotation implementation yet.
That larger state model still needs its own design and review.

Related host-side compatibility follow-up: NousResearch/hermes-agent#8416.
That companion PR improves Hermes Agent's external context-engine lifecycle ergonomics without changing default built-in behavior or bundling hermes-lcm into Hermes Agent.

## Test Plan
- [x] `python3 -m pytest tests/test_lcm_engine.py -q`
- [x] `python3 -m pytest tests/ -q`
- [x] local host dogfood check confirmed Hermes is loading this plugin branch from `~/.hermes/plugins/hermes-lcm`

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -287,6 +287,35 @@ class LCMEngine(ContextEngine):
             return 0
         return self._dag.reassign_session_nodes(old_session_id, new_session_id)
 
+    def rollover_session(
+        self,
+        old_session_id: str,
+        new_session_id: str,
+        previous_messages: List[Dict[str, Any]] | None = None,
+        carry_over_context: bool = True,
+        **kwargs,
+    ) -> int:
+        """Complete a Hermes-style `/new` rollover for this engine.
+
+        This is a small helper for host/runtime integrations that need the
+        correct lifecycle ordering in one call:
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Simple and correct. Having the lifecycle ordering (flush → reset → bind → carry over) in one method prevents integration bugs. Repeated rollover test is a good edge case to cover.

PR #18, opened 2026-04-15
fix: harden sqlite migrations and recovery

Tosko4:
## Summary
- add a shared SQLite bootstrap path for store/DAG setup
- introduce a small versioned migration baseline (`schema_version = 2` + `lcm_migration_state`)
- repair malformed or drifted external-content FTS tables on startup
- fall back to `LIKE` search if FTS is unavailable at runtime
- raise SQLite `busy_timeout` to 30s and cover longer write-lock waits with regression tests

## Why
Existing `CREATE TABLE IF NOT EXISTS` bootstrap was fine for fresh DBs, but too weak for real upgrades and messy legacy state. This tightens the plugin-side DB layer so old or damaged local state is less likely to stay broken or fail under normal write contention.

## Notes
- this stays fully inside `hermes-lcm`; no Hermes Agent/core changes are involved
- startup repair and runtime fallback are both included, so degraded FTS state is less likely to crash search outright

## Test Plan
- `python3 -m pytest tests/test_lcm_core.py -q`

### DIFF
--- a/dag.py
+++ b/dag.py
@@ -23,7 +23,7 @@ from .db_bootstrap import (
     ExternalContentFtsSpec,
     configure_connection,
     ensure_external_content_fts,
-    set_schema_version,
+    run_versioned_migrations,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


stephenschoettler (approved):
Important reliability work. The shared db_bootstrap module with FTS health detection, auto-rebuild, versioned migrations, and LIKE fallback makes the plugin much more resilient to corrupted or legacy databases. Extended busy_timeout to 30s is a good call for real-world multi-process access. Thorough test coverage including legacy DB upgrades, FTS drift, and write lock contention.

PR #19, opened 2026-04-15
✨ feat: make lcm_grep recall much sharper

Tosko4:
## 🎉 This makes `lcm_grep` feel a lot less like “it found something” and a lot more like “it found the right thing”

This PR is a retrieval-quality pass for `lcm_grep`.

The short version:
- search ordering is now intentional instead of flat
- summary recall uses source-content recency, not just compaction-row time
- short CJK fragments and emoji-heavy queries no longer quietly fall through the cracks

That means `lcm_grep` gets noticeably better at the stuff that actually matters in real use:
- finding the newest relevant thing when you care about recency
- surfacing the strongest older hit when you care about relevance
- staying useful when the query is not nice clean ASCII/Latin text

### 1. Explicit search ordering
`lcm_grep`, message search, and summary search now support:
- `recency`
- `relevance`
- `hybrid`

These orderings are applied before `LIMIT`, so the ranking actually affects what survives the cut.

### 2. Summary source-window timestamps
Summary nodes now persist:
- `earliest_at`
- `latest_at`

This lets summary recall rank on when the summarized content happened, instead of only when the summary row was created.

Also included:
- in-place upgrade for older DAG DBs that do not have these columns yet
- leaf summaries derive their source window from source message timestamps
- condensed summaries derive their source window from child-node windows

### 3. CJK / emoji-aware fallback
Short CJK fragments and emoji-heavy queries should not blindly trust plain SQLite FTS tokenization.

This PR routes those queries through substring fallback for both:
- message search
- summary search

And the fallback path keeps the same ordering semantics:
- `recency`
- `relevance`
- `hybrid`

So CJK / emoji queries do not silently collapse into a different search experience.

### 4. Better tool guidance
The `lcm_grep` schema guidance now more clearly nudges:
- short focused FTS queries
- quoted multi-word phrases
- the fact that short CJK / emoji-heavy queries may use substring fallback

## Why this is worth it
Before this, `lcm_grep` was functional, but still had a few rough edges:
- summaries were effectively ranked by compaction timing more than source timing
- CJK / emoji retrieval could be incomplete or brittle
- fallback behavior could feel different from the main FTS path

After this, the retrieval story is a lot cleaner and more honest.

## Validation
Ran locally:
- `python3 -m pytest tests/test_lcm_core.py -k 'sort_modes_apply_before_limit' -q`
- `python3 -m pytest tests/test_lcm_engine.py -k 'grep_reports_sort_mode' -q`
- `python3 -m pytest tests/test_lcm_core.py -k 'source_window or search_sort_modes_apply_before_limit' -q`
- `python3 -m pytest tests/test_lcm_engine.py -k 'source_window' -q`
- `python3 -m pytest tests/test_lcm_core.py -k 'cjk_queries or emoji_queries' -q`

Latest full run:
- `python3 -m pytest tests/ -q` → `95 passed`

## Files
- `store.py`
- `dag.py`
- `engine.py`
- `tools.py`
- `schemas.py`
- `search_query.py`
- `tests/test_lcm_core.py`
- `tests/test_lcm_engine.py`

### DIFF
--- a/dag.py
+++ b/dag.py
@@ -28,6 +28,25 @@ from .db_bootstrap import (
 
 logger = logging.getLogger(__name__)
 
+AGE_DECAY_RATE = 0.001
+
+
+def _normalize_search_sort(sort: str | None) -> str:
+    normalized = (sort or "recency").strip().lower()
+    return normalized if normalized in {"recency", "relevance", "hybrid"} else "recency"
+
+
+def _build_search_order_by(sort: str | None, created_at_expr: str) -> str:
+    normalized = _normalize_search_sort(sort)
+    if normalized == "relevance":
+        return f"rank ASC, {created_at_expr} DESC"
+    if normalized == "hybrid":
+        return (
+            f"(rank / (1 + (((strftime('%s','now') - {created_at_expr}) / 3600.0) * {AGE_DECAY_RATE}))) ASC, "
+            f"{created_at_expr} DESC"
+        )
+    return f"{created_at_expr} DESC"
+
 
 @dataclass
 class SummaryNode:
@@ -42,6 +61,7 @@ class SummaryNode:
     source_type: str = "messages"  # "messages" or "nodes"
     created_at: float = 0.0
     expand_hint: str = ""  # "Expand for details about: ..."
+    search_rank: float | None = None
 
 
 class SummaryDAG:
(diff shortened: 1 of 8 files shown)


stephenschoettler (approved):
Nice search quality upgrade. The sort modes (recency/relevance/hybrid) give the agent real control over recall. CJK and emoji LIKE fallback is thoughtful — FTS5 tokenizers miss those. Source time windows on DAG nodes enable the hybrid scoring to work across condensation levels. Clean extraction into search_query.py. Heads up: this will need a rebase after #18 merges since both touch the search methods in dag.py and store.py.

stephenschoettler:
Approved and ready to merge, but there are conflicts with #18 (which just merged). The overlapping changes are in dag.py and store.py search methods. Can you rebase onto main? Should be straightforward since both PRs touch the same methods but in complementary ways.

Tosko4:
Rebased onto current main and resolved the overlap with the recently merged search changes. I reran the local test suite after the rebase (python3 -m pytest tests/ -q -> 116 passed). CI is running again on the updated head now.

stephenschoettler (approved):
Rebase looks clean. Ready to go.

PR #20, opened 2026-04-15
✨ feat: make compaction more adaptive and cache-friendly

Tosko4:
## Summary

This makes the current compaction path more adaptive and more resilient without pretending the plugin has cache metrics it does not actually receive.

The original policy slice is still here:

- dynamic leaf chunk sizing
- adaptive leaf rescue for retry-worthy summarization failures
- cache-friendly follow-on condensation gating

But the branch now also includes the hardening that fell out of local dogfooding and broader maturity review:

- hyphenated / operator-heavy retrieval fallback hardening for `lcm_grep` and `lcm_expand_query`
- ranking changes that demote raw tool-output noise below conversational hits
- bounded dynamic catch-up leaf passes when one leaf pass still leaves the active context under meaningful pressure
- reduced session-filter config log noise

## What changed

- add `dynamic_leaf_chunk_enabled` and `dynamic_leaf_chunk_max`
- compute a working leaf threshold from backlog pressure when dynamic mode is enabled
- compact the oldest bounded raw chunk instead of always taking the entire raw backlog outside the fresh tail
- add bounded adaptive rescue for retry-worthy timeout / context / token-limit style failures
- keep non-retry-worthy failures as hard failures instead of looping
- add `cache_friendly_condensation_enabled` and `cache_friendly_min_debt_groups`
- suppress single-group follow-on condensation after a leaf pass while still allowing one bounded follow-on pass when real debt is present
- add bounded dynamic catch-up leaf passes so dynamic mode can continue compacting when one pass still leaves the context above threshold
- harden retrieval fallback for risky query shapes like `hermes-lcm`, `plugin-only`, and operator-heavy query strings
- demote `tool`-role search hits below conversational hits in both store-level and combined `lcm_grep` relevance ordering
- reduce repeated filter-config diagnostics noise in the engine logs
- document the new point-8 knobs in the README

## Why this helps

The previous path was safe, but still blunt in a few practical ways:

- a large raw backlog could trigger an oversized leaf pass
- retry-worthy provider failures could repeat the same bad chunk choice
- dynamic mode could still stop after one pass even when pressure remained high
- retrieval could fall over on risky query shapes or return noisy tool-output-heavy hits
- repeated filter diagnostics added noise during runtime debugging

This keeps the default behavior intact when the new knobs are off, while making the opt-in path more resilient and more usable in the real runtime.

## Validation

- `python3 -m pytest tests/test_lcm_engine.py -k 'dynamic_leaf_chunk_sizing_runs_bounded_catchup_passes_when_pressure_remains_high or dynamic_leaf_chunk_sizing_compacts_only_oldest_bounded_raw_chunk or adaptive_leaf_rescue_retries_with_smaller_oldest_chunk or cache_friendly_gating' -q`
  - `6 passed`
  - `117 passed`
  - `130 passed`

Also rechecked locally against the real `~/.hermes/lcm.db` and the live Hermes runtime after restart:

- risky retrieval queries now return hits instead of surfacing the earlier SQLite parser failure path
- `lcm_expand_query` no longer reproduces the earlier `no such column: lcm` / `no such column: only` failure
- top message hits for `hermes-lcm`, `plugin-only`, and `vendoring` are no longer dominated by raw tool JSON noise
- no fresh post-restart journald hits appeared for `database is locked`, `overflow_recovery_failed`, or the old parser-failure path

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,10 +140,14 @@ Environment variables (all optional):
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `LCM_FRESH_TAIL_COUNT` | `64` | Recent messages protected from compaction |
-| `LCM_LEAF_CHUNK_TOKENS` | `20000` | Token threshold for leaf compaction |
+| `LCM_LEAF_CHUNK_TOKENS` | `20000` | Token threshold floor for leaf compaction |
+| `LCM_DYNAMIC_LEAF_CHUNK_ENABLED` | `false` | Opt-in dynamic oldest-chunk sizing instead of always compacting the full raw backlog outside the fresh tail |
+| `LCM_DYNAMIC_LEAF_CHUNK_MAX` | `50000` | Upper bound for dynamic leaf chunk sizing |
 | `LCM_CONTEXT_THRESHOLD` | `0.75` | Fraction of context window triggering compaction |
 | `LCM_INCREMENTAL_MAX_DEPTH` | `1` | Max condensation depth (`0` = disabled, `-1` = unlimited) |
 | `LCM_CONDENSATION_FANIN` | `4` | Same-depth nodes needed to trigger condensation |
+| `LCM_CACHE_FRIENDLY_CONDENSATION_ENABLED` | `false` | Opt-in suppression of low-value follow-on condensation after a leaf pass |
+| `LCM_CACHE_FRIENDLY_MIN_DEBT_GROUPS` | `2` | Debt threshold multiplier before cache-friendly gating allows a follow-on condensation pass |
 | `LCM_IGNORE_SESSION_PATTERNS` | *(empty)* | Comma-separated glob patterns for sessions to exclude from LCM storage entirely |
 | `LCM_STATELESS_SESSION_PATTERNS` | *(empty)* | Comma-separated glob patterns for sessions that stay read-only (`platform:session_id` matching supported) |
 | `LCM_SUMMARY_MODEL` | *(auxiliary)* | Override model for summarization |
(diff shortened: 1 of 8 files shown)


stephenschoettler (approved):
Two well-scoped opt-in features that complement each other nicely.

Dynamic leaf chunk sizing is the right approach — compacting the entire backlog in one shot doesn't scale, and the rescue retry with shrinking chunks is a pragmatic way to handle model context limits without failing the whole compaction.

Cache-friendly condensation gating is clever — no point burning an LLM call to condense 1 fanin group when another leaf pass is about to add more. The forced-overflow bypass ensures safety isn't compromised.

Good that both are disabled by default. 7 thorough tests.

stephenschoettler:
Approved! Conflicts with #19 after merge — needs a rebase onto main. Should be straightforward, mostly engine.py and tests.

Tosko4:
Follow-up after the earlier approval: I pushed one more hardening pass onto this branch and updated the PR description to match the current diff.

Extra changes now included:
- hardened retrieval fallback for risky query shapes like `hermes-lcm`, `plugin-only`, and operator-heavy query strings
- demoted raw `tool` output noise below conversational hits in relevance ordering
- added bounded dynamic catch-up leaf passes when one pass still leaves the active context above threshold
- reduced repeated filter-config diagnostics noise in the engine logs

Validation rerun on the updated head:
- `python3 -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q` -> `117 passed`
- `python3 -m pytest tests/ -q` -> `130 passed`

I also rechecked this against the real local runtime/DB after restart: the earlier `lcm_expand_query` parser failure path is no longer reproduced, and top hits for the noisy retrieval cases are no longer dominated by raw tool JSON output.

PR #21, opened 2026-04-16
feat: custom summary instructions and pre-compaction extraction

stephenschoettler:
## Summary

Two features inspired by [lossless-claw-vigil-recall](https://github.com/jamebobob/lossless-claw-vigil-recall), which identified these as the top production pain points with LCM-style context management:

- **Custom summary instructions** — configurable string injected into all summarization prompts (L1 and L2). Controls summary voice and style. e.g. `LCM_CUSTOM_INSTRUCTIONS="Write as a neutral documenter. Use third person. Report facts, not significance."`

- **Pre-compaction extraction** — before messages are summarized into DAG nodes, an optional LLM call extracts decisions, commitments, and outcomes into daily note files (`YYYY-MM-DD.md`). Best-effort: failures never block compaction. Disabled by default.

## New config

| Variable | Default | Description |
|----------|---------|-------------|
| `LCM_CUSTOM_INSTRUCTIONS` | `""` | Instructions injected into all summarization prompts |
| `LCM_EXTRACTION_ENABLED` | `false` | Enable pre-compaction decision extraction |
| `LCM_EXTRACTION_MODEL` | *(summary model)* | Override model for extraction |
| `LCM_EXTRACTION_OUTPUT_PATH` | `~/.hermes/lcm-extractions/` | Directory for daily extraction files |

## Test plan

- [x] Custom instructions appear in L1 and L2 prompts when set
- [x] Custom instructions omitted when empty (no prompt pollution)
- [x] Extraction writes daily files with session ID and timestamp headers
- [x] Extraction skips when LLM returns NOTHING_TO_EXTRACT
- [x] Extraction never blocks on LLM failure
- [x] Multiple extractions append to the same daily file
- [x] Config defaults and env var parsing

7 new tests, all passing. 1 pre-existing failure in `test_dynamic_leaf_chunk_sizing_runs_bounded_catchup_passes_when_pressure_remains_high` (from #20, not this PR).

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -6,6 +6,7 @@ with a DAG-based summarization system that preserves every message.
 
 import json
 import logging
+import os
 import re
 import time
 from pathlib import Path
--- a/engine.py
+++ b/engine.py
@@ -16,6 +16,7 @@ from agent.context_engine import ContextEngine
 from .config import LCMConfig
 from .dag import SummaryDAG, SummaryNode
 from .escalation import summarize_with_escalation
+from .extraction import extract_before_compaction
 from .schemas import LCM_DESCRIBE, LCM_DOCTOR, LCM_EXPAND, LCM_EXPAND_QUERY, LCM_GREP, LCM_STATUS
 from .session_patterns import (
     build_session_match_keys,
@@ -236,6 +237,7 @@ class LCMEngine(ContextEngine):
                     l2_budget_ratio=self._config.l2_budget_ratio,
                     l3_truncate_tokens=self._config.l3_truncate_tokens,
                     focus_topic=focus_topic or "",
+                    custom_instructions=self._config.custom_instructions,
                 )
                 return attempt_chunk, source_tokens, summary_text, level, attempt_number
             except Exception as exc:
@@ -334,6 +336,10 @@ class LCMEngine(ContextEngine):
 
             if not to_compact:
                 break
+
+            # Pre-compaction extraction: best-effort, never blocks compaction
+            if self._config.extraction_enabled:
+                self._run_pre_compaction_extraction(to_compact)
 
             compacted_chunk, source_tokens, summary_text, _level, _rescue_attempts = self._summarize_leaf_chunk_with_rescue(
                 to_compact,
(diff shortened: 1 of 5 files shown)


engine.py
```diff
@@ -684,6 +690,25 @@ def _get_store_ids_for_messages(self, messages: List[Dict[str, Any]]) -> List[in
 
     # -- Internal: summarization -------------------------------------------
 
+    def _run_pre_compaction_extraction(self, messages: List[Dict[str, Any]]) -> None:
+        """Best-effort extraction of decisions before compaction."""
+        try:
+            serialized = self._serialize_messages(messages)
+            output_path = self._config.extraction_output_path
+            if not output_path:
+                base = self._hermes_home or os.path.expanduser("~/.hermes")
```
Tosko4, line 749: Blocking issue: this default-path fallback uses `os.path.*`, but `engine.py` does not import `os`. Because `_run_pre_compaction_extraction()` swallows exceptions, `LCM_EXTRACTION_ENABLED=true` with the default/empty `LCM_EXTRACTION_OUTPUT_PATH` turns into a silent no-op instead of writing to `~/.hermes/lcm-extractions/`. I reproduced this locally and got `name 'os' is not defined` from this path.

Tosko4 (changes requested):
I like the shape of this overall — the custom-instructions wiring is straightforward, the extraction hook is nicely isolated, and the targeted tests cover the happy path well.

I did hit one blocking issue locally though:

- `LCM_EXTRACTION_OUTPUT_PATH` is documented as optional with a default of `~/.hermes/lcm-extractions/`
- but the fallback path in `LCMEngine._run_pre_compaction_extraction()` currently references `os.path.*` without importing `os`
- because the method wraps everything in a broad `try/except`, that turns into a silent no-op instead of writing the extraction file

So with `LCM_EXTRACTION_ENABLED=true` and the default output path (empty config), extraction never actually runs.

I reproduced it locally by calling `_run_pre_compaction_extraction()` with `extraction_enabled=True` and an empty `extraction_output_path`; the method logged:
`Pre-compaction extraction failed (non-blocking): name 'os' is not defined`

Once that default-path case is fixed, this looks close.

stephenschoettler:
Fixed the missing `os` import and added a test that exercises the default path case — catches the silent failure scenario you flagged. Ready for re-review when you get a chance.

Tosko4 (approved):
Looks good now. I re-checked the default extraction-path fix, reviewed the custom-instructions/extraction wiring against current main, and validated the merged shape locally by rebasing this branch onto current main.

Validation I ran locally on the rebased result:
- python3 -m pytest tests/test_lcm_core.py -q
- python3 -m pytest tests/ -q
- python3 -m compileall -q .

I also ran a synthetic compression probe to confirm:
- default extraction path resolves under hermes_home when config is empty
- custom instructions are passed through to summarization
- extraction remains non-blocking and uses the configured summary/extraction model path as expected

Happy to merge this.

PR #22, opened 2026-04-16
fix: accept api_mode in update_model

leavrcn:
## Summary
- accept api_mode in update_model
- preserve compatibility for callers that pass api_mode during model updates

## Notes
- opened as a draft to verify the fork/push/create-PR flow with the refreshed GitHub token

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -330,7 +330,7 @@ class LCMEngine(ContextEngine):
 
     def update_model(self, model: str, context_length: int,
                      base_url: str = "", api_key: str = "",
-                     provider: str = "") -> None:
+                     provider: str = "", api_mode: str = "", **kwargs) -> None:
         self.context_length = context_length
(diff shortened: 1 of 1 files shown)


Tosko4:
Closing this as superseded by main.

I validated it against the current hermes-agent call sites before closing:
- hermes-lcm main already accepts api_mode in LCMEngine.update_model
- current hermes-agent callers only pass model/context_length/base_url/api_key/provider, plus api_mode in the hot-switch path
- no current caller passes extra kwargs today

So the compatibility issue described in this PR is already resolved on main. The only remaining delta here is accepting arbitrary **kwargs for future-proofing, which is reasonable but not required for current compatibility.

If we want that extra tolerance later, it would be cleaner as a small follow-up PR with an updated title/body focused on forward-compatibility rather than api_mode support.

PR #24, opened 2026-04-16
feat: add lifecycle checkpoint substrate

Tosko4:
## Summary
- add a small durable lifecycle state store for session/conversation binding
- persist active and finalized frontier markers across rollover/rebind boundaries
- expose lifecycle metadata in engine status for inspection/debugging
- add regression coverage for bootstrap, rollover idempotence, frontier persistence, and restart recovery

## Why
Right now most lifecycle/checkpoint knowledge only exists in memory. That makes `/new`-style rollover and restart/rebind behavior harder to reason about, and it leaves no durable substrate for later maintenance/debt work.

This keeps the design deliberately small:
- one `lcm_lifecycle_state` table
- no transcript import/replay architecture
- no operator apply flows yet

## Validation
- `python3 -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q` → `123 passed`
- `python3 -m pytest tests/ -q` → `136 passed`
- `git diff --check`
- host/runtime dogfood against the real plugin path confirmed:
  - active frontier persists after compaction
  - rollover moves it into `last_finalized_frontier_store_id`
  - restart/rebind recovers the same lifecycle state

### DIFF
--- a/db_bootstrap.py
+++ b/db_bootstrap.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 import sqlite3
 from typing import Iterable, Sequence
 
-SCHEMA_VERSION = 2
+SCHEMA_VERSION = 3
 SQLITE_BUSY_TIMEOUT_MS = 30_000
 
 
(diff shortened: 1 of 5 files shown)


stephenschoettler (approved):
Clean minimal substrate — right scope for a first pass. Migration is safe, frontier advancement ordering is correct, idempotence guards are solid.

Minor: `bind_session` has dead branches after the early return for same-session rebind (lines computing `current_frontier`/`current_bound_at` with `existing.current_session_id == session_id` checks that can never be true at that point). Not a bug — values default correctly — just dead code worth cleaning up later.

6 well-targeted tests covering the key invariants. Ship it.

PR #25, opened 2026-04-16
feat: add deferred maintenance debt catch-up

Tosko4:
## Summary
- persist deferred maintenance debt in durable lifecycle state
- add bounded cross-turn catch-up for raw backlog debt
- surface debt state in status output so operators can see when maintenance work is still owed

## Details
- lifecycle state now stores debt kind/size and maintenance timestamps
- config adds deferred maintenance toggles and bounded pass limits
- raw backlog debt is preserved until backlog is actually reduced below threshold
- later turns can consume debt in bounded passes instead of turning foreground work into an unbounded cleanup sweep

## Validation
- python3 -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q
- python3 -m pytest tests/ -q
- python3 -m compileall -q .
- git diff --check

I also re-based this branch onto current main after the other slices landed and re-ran the validation on the rebased result.

### DIFF
--- a/config.py
+++ b/config.py
@@ -65,6 +65,12 @@ class LCMConfig:
     # Minimum number of same-depth fanin groups before one follow-on
     # condensation pass is allowed in cache-friendly mode
     cache_friendly_min_debt_groups: int = 2
+    # When enabled, turns can persist raw-backlog maintenance debt and use
+    # later bounded catch-up passes to reduce it.
+    deferred_maintenance_enabled: bool = False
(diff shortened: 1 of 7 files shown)


stephenschoettler (approved):
Clean extension of the lifecycle substrate. Migration is safe (ALTER TABLE with column existence checks), debt tracking is conservative (off by default, bounded passes), and status surfacing gives operators visibility.

3 well-structured tests covering the full debt lifecycle: persist → bounded catch-up → clear. Ship it.

PR #26, opened 2026-04-16
feat: broaden operator remediation flows

Tosko4:
## Summary
- broaden `/lcm doctor` so it separates observations from recommended remediation
- add a narrow backup-first `doctor clean apply` flow for safe pattern-matched cleanup candidates
- keep remediation limited to well-understood state transitions instead of a broad destructive sweep

## Details
- doctor now surfaces maintenance debt and cleanup candidates as operator-facing observations
- recommended actions are called out explicitly instead of being mixed into raw status output
- backup logic is shared between `/lcm backup` and cleanup apply
- cleanup apply only removes safe candidate session state and skips protected live-session references

## Validation
- python3 -m pytest tests/test_lcm_command.py -q
- python3 -m pytest tests/ -q
- python3 -m compileall -q .
- git diff --check

## Stacking note
This PR is intentionally stacked on `feat/deferred-maintenance-debt` while #25 is still open.

### DIFF
--- a/command.py
+++ b/command.py
@@ -37,6 +37,7 @@ def _help_text(error: str | None = None) -> str:
         "- /lcm or /lcm status: show current LCM runtime/session status",
         "- /lcm doctor: run read-only LCM health checks",
         "- /lcm doctor clean: best-effort scan of obvious junk/noise session candidates without deleting anything",
+        "- /lcm doctor clean apply: backup-first cleanup for safe pattern-matched candidates only",
         "- /lcm backup: create a timestamped SQLite backup before any future cleanup workflow",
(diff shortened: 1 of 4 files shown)


stephenschoettler (approved):
Good refactor — shared helpers for scan/backup, clean separation of observations vs recommended actions in doctor output, and the backup-first apply flow is the right safety model.

`delete_safe_rows_for_sessions` is appropriately conservative (only deletes when ALL session refs are in the candidate set, skips protected live sessions). Tests cover both happy path and backup-failure abort.

Nit: missing newline at EOF in command.py — not worth blocking on.

PR #28, opened 2026-04-16
feat: sanitize inline media before compaction

Tosko4:
## Summary
- sanitize inline image/audio/video data URIs before pre-compaction serialization
- preserve user text while replacing inline media with attachment markers
- clean assistant tool-call arguments too, including JSON strings and parsed dict/list payloads
- keep safe metadata for structured file blocks instead of dropping them

## Why
Raw inline media/base64 blobs should not be fed straight into pre-compaction extraction and summarization. This keeps the first slice focused on safer transcript preparation without introducing storage/externalization yet.

## Validation
- python3 -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q
- python3 -m pytest tests/ -q
- python3 -m compileall -q .
- git diff --check

## Notes
- scoped as plugin-only work in hermes-lcm
- sits next to the existing pre-compaction extraction path rather than introducing a broader transform framework yet
- covers chat-completions style blocks, Responses-style blocks, tool-call arguments, and structured file metadata preservation

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -17,7 +17,11 @@ from agent.context_engine import ContextEngine
 from .config import LCMConfig
 from .dag import SummaryDAG, SummaryNode
 from .escalation import summarize_with_escalation
-from .extraction import extract_before_compaction
+from .extraction import (
+    extract_before_compaction,
+    sanitize_pre_compaction_content,
+    sanitize_pre_compaction_tool_arguments,
+)
 from .schemas import LCM_DESCRIBE, LCM_DOCTOR, LCM_EXPAND, LCM_EXPAND_QUERY, LCM_GREP, LCM_STATUS
 from .session_patterns import (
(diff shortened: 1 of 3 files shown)


stephenschoettler (approved):
This is exactly the right first slice — clean media/base64 stripping before compaction, scoped to serialization only, no storage or retrieval changes.

Implementation is solid:
- Recursive content handler covers all multimodal formats (chat-completions blocks, responses-style blocks, plain strings)
- Regex correctly scopes to image/audio/video — leaves application/* data URIs alone
- Tool call argument sanitization handles both JSON strings and pre-parsed dicts
- Structured file block metadata preservation is a nice touch (keeps file_id, filename, mime_type)
- 11 well-targeted tests covering every content shape

Clean and focused. Ship it.

PR #33, opened 2026-04-17
fix: isolate LCM storage per profile and add source-aware retrieval

stephenschoettler:
## Summary
- fix profile-local DB resolution by preferring `HERMES_HOME/lcm.db` when available
- add `source` metadata to LCM message storage and support source-filtered search
- add schema migration for legacy DBs and persist `lcm_schema_version=2`
- extend `lcm_grep` with `session_scope` (current|all) and optional `source` filter
- update README with recall guidance and new grep/path behavior

## Tests
- added regression tests for DB path resolution, source filtering, schema params, and migration/version metadata
- ran full suite: `pytest -q` (40 passed)

## Notes
- left `manim-lcm-explainer/` untracked and excluded from this PR

### DIFF
--- a/dag.py
+++ b/dag.py
@@ -323,14 +323,25 @@ class SummaryDAG:
     # -- Search -------------------------------------------------------------
 
     def search(self, query: str, session_id: str | None = None,
-               limit: int = 20, sort: str | None = None) -> List[SummaryNode]:
+               limit: int = 20, sort: str | None = None,
+               source: str | None = None) -> List[SummaryNode]:
         """FTS5 search across all summary nodes."""
         terms = extract_search_terms(query)
         phrases = extract_quoted_phrases(query)
         if requires_like_fallback(query):
-            return self._search_like(query, session_id=session_id, limit=limit, sort=sort)
+            return self._search_like(query, session_id=session_id, limit=limit, sort=sort, source=source)
 
         order_by = _build_search_order_by(sort, "COALESCE(n.latest_at, n.created_at)")
+        source_clause = ""
+        source_args: list[Any] = []
+        if source:
+            source_clause = (
+                " AND EXISTS ("
+                "SELECT 1 FROM messages m "
+                "WHERE m.session_id = n.session_id AND m.source = ?"
+                ")"
+            )
+            source_args.append(source)
         fetch_limit = compute_search_fetch_limit(limit, terms, phrases)
         apply_directness_adjustment = should_apply_directness_rank_adjustment(terms, phrases)
         max_rank_bonus = compute_directness_rank_bonus_upper_bound(terms, phrases) * 2e-7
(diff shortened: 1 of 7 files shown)


Tosko4 (changes requested):
Thanks for taking this on — the profile-scoped DB path change looks useful, but I don't think the retrieval side is merge-ready yet.

I pulled the branch locally against current `origin/main` and hit the same stale-branch problem GitHub is showing now: the PR is behind and conflicts with current main in multiple files. The focused test slice on the branch still passes (`pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q`), but there's still a correctness gap in the advertised source-aware retrieval behavior.

Right now the `source` filter is only applied to raw-message search. `lcm_grep(...)` passes `source` into `_store.search(...)`, but summary retrieval still goes through `_dag.search(query, session_id=session_id, limit=limit)` with no equivalent source constraint, and `SummaryDAG.search(...)` doesn't have a source-aware filtering path.

I verified that with a small local repro on this branch:
- one raw `discord` hit
- one raw `telegram` hit
- one `discord` summary node
- one `telegram` summary node

Then calling `lcm_grep(query='docker', session_scope='all', source='discord')` returned the expected raw `discord` hit, but it also returned the unrelated `telegram` summary.

So at the moment the PR only filters one backend of merged retrieval. I'd rebase first, then make sure the source constraint is applied consistently across:
- raw message search
- summary/DAG search
- the merged `lcm_grep` result set

Once that's fixed, I think the overall direction makes sense.

stephenschoettler:
Thanks — I addressed the gap you called out and rebased the branch onto current `main`.

What changed:
- added per-message `source` persistence in the message store
- added lazy migration for legacy DBs missing the `source` column
- added `session_scope` + `source` to the `lcm_grep` schema/handler
- applied `source` filtering to **both** raw-message search and summary/DAG search
- kept the merged `lcm_grep` result set source-aware end-to-end
- added regression coverage for cross-session source-filtered summary retrieval

Local validation on the updated branch:
- `pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q` → `207 passed`
- `pytest -q` → `223 passed`
- `python -m compileall -q . && git diff --check` → clean

CI is green now as well. Would appreciate another pass when you have a minute.

Tosko4:
Thanks for the quick turnaround on this. The rebase + the cross-session fix definitely improved things, and I rechecked the branch locally.

I think there’s still one narrower leak left though.

I can reproduce it with a **single mixed-source session**:

1. add one raw `discord` message and one raw `telegram` message to the same session
2. create two D0 summaries in that same session, each pointing only at one of those source messages
3. run `lcm_grep(query='docker', session_scope='all', source='discord')`

Expected:
- the raw `discord` hit
- the `discord` summary only

Actual:
- the raw `discord` hit
- **and the unrelated `telegram` summary**

So it looks like the current summary filter is still operating at the **session** level ("does this session contain any message with source X?") rather than at the **summary-source** level ("do this node's `source_ids` resolve to messages with source X?").

So from what I can see, the original cross-session case is fixed now, but mixed-source sessions are still leaking summaries across sources.

Would you mind taking one more pass at that narrower case? I think adding a regression test for a mixed-source single-session setup would pin it down pretty cleanly.

@stephenschoettler

stephenschoettler:
Good catch. You were right — the previous fix stopped the cross-session leak, but mixed-source sessions could still leak summaries because the summary-side source check was effectively session-level instead of source-lineage-level.

I pushed another update that fixes that narrower case.

What changed:
- added a regression test for a mixed-source single-session setup
- changed summary filtering so it resolves a node's `source_ids` recursively down to message descendants before deciding whether the node matches the requested source
- this prevents unrelated summaries in the same session from leaking into `lcm_grep(..., source=...)`

Validation on the updated branch:
- `pytest --basetemp=/tmp/hermes-lcm-pr33 tests/test_lcm_core.py tests/test_lcm_engine.py -q` → `208 passed`
- `pytest -q` → `224 passed`
- `python -m compileall -q . && git diff --check` → clean

Would appreciate one more look.

Tosko4:
Rechecked this after the latest fix and after rebasing the branch onto current `main`.

What I re-verified locally on the rebased shape:
- the original cross-session source-filtered summary case
- the narrower mixed-source single-session repro that was still leaking before
- a nested summary-lineage case where a parent summary resolves through child summaries down to source messages

All three now behave correctly from what I can see, and the refreshed CI run on the updated head is green as well.

Thanks for tightening this up — this looks good to me now.

Tosko4 (approved):
Rechecked on the rebased head. The source-aware retrieval behavior we asked for now looks right, including the mixed-source single-session case. LGTM.

PR #43, opened 2026-04-18
feat: add read-only retention analysis command

Tosko4:
## Summary
- add a read-only `/lcm doctor retention` subcommand
- report per-session footprint and age without deleting anything
- include summary-node footprint in the reported token totals
- keep stale sessions visible in the candidate list instead of letting them disappear behind fresh heavy sessions

## Why
This gives operators a safe first retention-analysis step before any future prune/apply workflow. It makes it easier to see which stored sessions are old, large, or both, without jumping straight to destructive cleanup.

## Validation
- `source /home/nabu/.hermes/hermes-agent/venv/bin/activate && pytest tests/test_lcm_command.py -q`
- `source /home/nabu/.hermes/hermes-agent/venv/bin/activate && pytest -q`
- `source /home/nabu/.hermes/hermes-agent/venv/bin/activate && python -m compileall -q . && git diff --check`

### DIFF
--- a/command.py
+++ b/command.py
@@ -38,6 +38,7 @@ def _help_text(error: str | None = None) -> str:
         "- /lcm doctor: run read-only LCM health checks",
         "- /lcm doctor clean: best-effort scan of obvious junk/noise session candidates without deleting anything",
         "- /lcm doctor clean apply: backup-first cleanup for safe pattern-matched candidates only",
+        "- /lcm doctor retention: read-only retention analysis for stored session footprint and age",
         "- /lcm backup: create a timestamped SQLite backup before any future cleanup workflow",
         "- /lcm help: show this help",
     ])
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
## Code Review Summary

**Verdict: Approved**

Looks good.

### What works
- Safe, read-only scope for `/lcm doctor retention`.
- Includes summary-node footprint, so summary-only sessions are visible.
- Candidate ordering matches the stated operator goal: stale first, then larger footprint, with protected current-session entries deprioritized.
- Output is explicit about being non-destructive and points operators to `/lcm backup` before any future prune flow.
- Test coverage is good for old/heavy sessions, summary-only sessions, truncation behavior, and help/error text updates.

### Validation
- `pytest tests/test_lcm_command.py -q` → passing
- `pytest -q` → passing

Minor follow-up only: if retention analysis grows beyond this first pass, it may eventually want a cached/helper path instead of aggregating everything on demand. Not blocking.

PR #45, opened 2026-04-18
feat: add opt-in transcript GC for externalized tool output

Tosko4:
## Summary
- add explicit opt-in transcript GC for already-externalized summarized tool-result rows
- rewrite eligible tool rows to compact GC placeholders while preserving `store_id` lineage
- keep externalized payloads inspectable through existing `externalized_ref` retrieval paths
- update GC token accounting so retention/footprint reporting reflects the rewritten transcript rows
- clarify the README around externalization, inspectability, and transcript-GC behavior/tradeoffs

## Why
Once oversized tool output can be externalized and inspected separately, keeping the full raw blob inline in the transcript forever is wasteful. This adds a narrow, opt-in cleanup step after successful compaction without breaking summary lineage or payload retrieval.

## Validation
- `pytest -q`
- `python -m compileall -q .`
- `git diff --check`

### DIFF
--- a/store.py
+++ b/store.py
@@ -1,10 +1,11 @@
-"""Immutable message store — the source of truth.
+"""Immutable-first message store — the source of truth.
 
-Every message is persisted verbatim and never modified. The store is
-append-only with optional pruning of very old messages (configurable).
-
-Each message gets a monotonic store_id used as a stable reference.
+Every message is persisted durably in SQLite. The normal model is append-only,
+with one narrow opt-in exception: already-externalized summarized tool-result
+rows may be rewritten to compact GC tombstones while preserving the original
+row identity (`store_id`) for DAG/source lookup.
 """
+
 
 import json
 import logging
@@ -193,6 +194,15 @@ class MessageStore:
                             VALUES('delete', old.store_id, old.content);
                     END;
                     """,
+                    """
+                    CREATE TRIGGER IF NOT EXISTS msg_fts_update
+                        AFTER UPDATE OF content ON messages BEGIN
+                        INSERT INTO messages_fts(messages_fts, rowid, content)
+                            VALUES('delete', old.store_id, old.content);
+                        INSERT INTO messages_fts(rowid, content)
+                            VALUES (new.store_id, new.content);
+                    END;
+                    """,
                 ),
             ),
         )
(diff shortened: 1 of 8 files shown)


stephenschoettler (changes requested):
## Code Review Summary

**Verdict: Changes Requested**

### Critical
- `store.py` — transcript GC rewrites the tool row content to a compact placeholder, but leaves the original `token_estimate` unchanged. That means operator-facing footprint accounting can remain inflated after GC, which undercuts the feature's stated retention/footprint value.

### Warnings
- The PR body appears to describe the retention-analysis command rather than this transcript-GC feature. Please update the description so the review context matches the actual branch.

### Validation
- `pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q` → passing locally
- `pytest -q` → passing locally

Please fix the token accounting issue before merge.

store.py
```diff
@@ -261,12 +271,30 @@ def delete_session_messages(self, session_id: str) -> int:
...
+            "SELECT role, pinned, content FROM messages WHERE store_id = ?",
+            (store_id,),
+        ).fetchone()
+        if row is None:
+            return False
+        role, pinned, current_content = row
+        if role != "tool" or bool(pinned) or current_content == placeholder:
+            return False
+        self._conn.execute(
+            "UPDATE messages SET content = ? WHERE store_id = ?",
```
stephenschoettler: This rewrite updates only `content`. The original `token_estimate` is left intact, so any retention/footprint reporting based on stored token totals will still count the row as if it contained the full giant blob. That undercuts the operator-facing value of transcript GC. Please either recompute `token_estimate` from the placeholder here, or introduce a separate retained-footprint field and make the reporting path use that instead.

Tosko4:
Thanks — both points are addressed now.

Follow-up on the updated branch:
- transcript GC now recomputes `token_estimate` when a tool row is rewritten to the compact GC placeholder, so footprint/retention reporting reflects the rewritten transcript row instead of the old raw blob size
- added regression coverage at both the store level and the end-to-end compaction/GC path
- rebased the branch onto current `main`
- fixed the PR body so it now describes the transcript-GC slice instead of the retention-analysis command

Validation on the updated head:
- `pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q` → `215 passed`
- `pytest -q` → `234 passed`
- `python -m compileall -q . && git diff --check` → clean

CI is green again on the refreshed head as well.

stephenschoettler (approved):
Looks good now.

The blocking issue from my earlier review is fixed: GC now recomputes `token_estimate` when rewriting transcript rows, and the new tests cover both the store-level rewrite and the end-to-end footprint behavior.

I re-ran the branch locally and it is clean:
- `pytest --basetemp=/tmp/hermes-lcm-pr45 -q`
- `python -m compileall -q .`
- `git diff --check`

Approve.

Tosko4:
Rechecked this after the latest follow-up.

What I re-verified on the updated branch:
- transcript GC now updates `token_estimate` when a tool row is rewritten to the compact GC placeholder
- the store-level and end-to-end 9D regression coverage for that accounting path is in place
- the branch is rebased onto current `main`
- the PR body now matches the actual transcript-GC slice

The refreshed CI run on the new head is green as well.

From what I can see, the blocker you called out is addressed now. Looks good to me.

PR #47, opened 2026-04-18
docs: clarify lcm tool choice vs session_search

### LINKED ISSUE #38: UX/Prompting: add core guidance for when to use LCM tools vs session_search
## Problem
LCM tool availability is currently schema-driven, but there is no explicit core guidance telling the agent when to use:
- LCM tools for intra-session recall
- `session_search` for cross-session recall

## Why this matters
Without guidance, tool choice can be inconsistent and costlier than needed.

## Decision questions
1. Should guidance live in Hermes core (`prompt_builder`/`run_agent`) or plugin docs only?
2. Should this be generic for all context engines or LCM-specific?
3. Should guidance be injected conditionally when context-engine tools are present?

## Proposed approach
Conditionally inject guidance when context-engine tools are loaded:
- prefer context-engine tools for current-session recall
- prefer `session_search` for cross-session recall

## Acceptance criteria
- [ ] Agreed placement (core vs docs-only)
- [ ] If core: implementation merged with tests
- [ ] If docs-only: rationale documented
- [ ] Behavior validated on at least one long-session scenario

Tosko4:
## Summary
- clarify the LCM tool schemas so they explicitly describe current-session vs cross-session recall
- document the same guidance in the README, including the deliberate `lcm_grep(session_scope='all')` escape hatch
- add a regression test that verifies the tool-facing schema text keeps this distinction visible

## Why
- issue #38 is about fuzzy tool choice when both LCM tools and `session_search` are visible to the model
- schema wording was doing too much implicit routing work, especially in ACP/API/editor flows where both surfaces can appear together
- this keeps the fix plugin-local for now while still making the routing guidance explicit in the actual tool surface the model sees

## Validation
- `pytest tests/test_lcm_engine.py::TestEngineABC::test_tool_schemas -q`
- `pytest -q`
- `python -m compileall -q .`
- `git diff --check`
- runtime probe with a live Hermes agent confirming `session_search` and `lcm_grep` are both exposed and now advertise distinct recall roles

## Notes
- this does not change Hermes core behavior
- it intentionally keeps `lcm_grep(session_scope='all')` available, but makes that cross-session LCM-store path explicit instead of leaving the overlap implicit
- if stronger host-wide guidance is still wanted later, that can be a separate Hermes-core follow-up

Closes #38

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,12 +172,26 @@ That means patterns like `cron:*` can catch Hermes cron sessions today, while pl
 
 | Tool | Description |
 |------|-------------|
-| `lcm_grep` | Search raw messages AND summaries across all depths. FTS5 syntax. |
-| `lcm_describe` | Inspect DAG structure or an `externalized_ref` payload preview without loading full payload content. No node_id/externalized_ref = session overview. |
-| `lcm_expand` | Recover original content from a summary node, or open a stored `externalized_ref` payload directly. |
-| `lcm_expand_query` | Answer a question from expanded LCM context using either a query or explicit node_ids. Uses the expansion path/model instead of the summarization path. |
+| `lcm_grep` | Search raw messages AND summaries for the active/current session. Use this for intra-session recall after compaction; if you intentionally want cross-session LCM store hits, use `session_scope='all'`; otherwise prefer `session_search` for earlier separate sessions. |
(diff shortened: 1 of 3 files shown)


stephenschoettler (approved):
Looks good. Clearer tool-routing guidance, scoped correctly, and schema coverage is in place.

PR #50, opened 2026-04-19
fix: salvage verified search and FTS hardening from #41

stephenschoettler:
## Summary
- salvage the verified query-sanitization changes from #41 into a fresh branch from `main`
- sanitize FTS `MATCH` queries and LIKE-fallback term extraction while preserving balanced phrase quotes
- degrade external-content FTS rebuilds to LIKE search when disk space is too low instead of rebuilding
- add focused regression tests for the salvaged paths

## Why
- #41 mixed good changes with unresolved blockers in other files and was not merge-ready
- this PR carries forward only the subset that was validated locally and leaves the broken areas out
- important constraint: this PR does not attempt to fix the remaining `engine.py`, `extraction.py`, or `command.py` concerns from #41

## Validation
- `ulimit -n 4096 && pytest -q`
- `python -m compileall -q .`
- `git diff --check`
- `pytest -q tests/test_lcm_core.py::TestMessageStore::test_search_relevance_still_surfaces_direct_phrase_hit_when_phrase_matches_many_spammy_candidates tests/test_lcm_core.py::TestMessageStore::test_search_like_fallback_sanitizes_fts_syntax_chars tests/test_lcm_core.py::TestDbBootstrapGuards::test_sanitize_fts5_query_preserves_balanced_phrase_quotes tests/test_lcm_core.py::TestDbBootstrapGuards::test_ensure_external_content_fts_skips_rebuild_when_disk_is_low tests/test_lcm_engine.py::TestPostCompactionIngestion::test_ingest_after_compaction tests/test_lcm_engine.py::TestPostCompactionIngestion::test_multiple_compactions`

## Notes
- supersedes the verified subset of #41 rather than extending that branch
- intentionally excludes the unresolved `engine.py`, `extraction.py`, and `command.py` work from #41
- credit to @cedricwyh for the original hardening attempt and for surfacing the target areas that this narrower salvage PR carries forward
- after this replacement PR is open, #41 should be closed as superseded rather than merged

### DIFF
--- a/dag.py
+++ b/dag.py
@@ -37,6 +37,7 @@ from .search_query import (
     extract_search_terms,
     normalize_search_sort,
     requires_like_fallback,
+    sanitize_fts5_query,
     should_apply_directness_rank_adjustment,
 )
 from .store import _normalize_source_value, _UNKNOWN_SOURCE
@@ -335,8 +336,9 @@ class SummaryDAG:
           session-level source presence
         - mixed-source nodes may match more than one ``source`` filter
         """
-        terms = extract_search_terms(query)
-        phrases = extract_quoted_phrases(query)
+        safe_query = sanitize_fts5_query(query)
+        terms = extract_search_terms(safe_query)
+        phrases = extract_quoted_phrases(safe_query)
         if requires_like_fallback(query):
             return self._search_like(query, session_id=session_id, limit=limit, sort=sort, source=source)
 
@@ -355,7 +357,7 @@ class SummaryDAG:
                            JOIN summary_nodes n ON n.node_id = fts.rowid
                            WHERE nodes_fts MATCH ? AND n.session_id = ?
                            ORDER BY {order_by} LIMIT ? OFFSET ?""",
-                        (query, session_id, fetch_limit, offset),
+                        (safe_query, session_id, fetch_limit, offset),
                     ).fetchall()
                 else:
                     rows = self._conn.execute(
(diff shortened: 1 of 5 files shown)


Tosko4:
Update: I pushed a follow-up hardening commit onto this PR branch.

What changed in the update:
- cleanly degrade FTS rebuilds under low-disk conditions by removing stale FTS artifacts/triggers before falling back to LIKE-only search
- align `MessageStore` and `SummaryDAG` search handling so sanitization is applied consistently across MATCH queries, LIKE fallback, and directness scoring
- tighten malformed quote / wildcard query sanitization and add regression coverage around those cases

Why:
- the previous PR shape still had a real failure mode where degraded startup could leave broken FTS triggers behind and cause later writes to fail with missing-table errors
- search behavior was also still inconsistent across raw-message vs DAG retrieval paths for operator-bearing queries

Validation:
- `pytest -q`
- `git diff --check`

Current PR head:
- `83b5991` — `fix: harden FTS degrade paths and query sanitization`

Tosko4 (approved):
Reviewed the updated PR head after the hardening follow-up. The low-disk FTS degrade path now cleans up stale FTS artifacts/triggers before falling back to LIKE search, query sanitization is applied consistently across MessageStore and SummaryDAG, and the added regression coverage closes the main edge cases around malformed quotes/wildcards. Re-ran local validation before pushing and the refreshed GitHub CI is green on 3.11 / 3.12 / 3.13. LGTM.

PR #52, opened 2026-04-19
fix: restrict lcm_grep to active session only

Tosko4:
### Motivation
- Prevent a cross-session data exposure introduced by allowing `session_scope='all'` in `lcm_grep`, restoring per-session confidentiality by enforcing active-session-only searches.

### Description
- Remove the `all` option and cross-session guidance from the `lcm_grep` schema in `schemas.py`, leaving `session_scope` limited to `"current"` only.
- Update the `lcm_grep` handler in `tools.py` to always enforce the active session (`engine._session_id`) and to ignore/log any non-`current` `session_scope` values instead of dropping the session filter.
- Adjust tests in `tests/test_lcm_engine.py` to remove expectations for cross-session behavior and to validate source filtering inside the active session.
- Simplify the `lcm_grep` docstring to reflect active-session-only behavior.

### Testing
- Attempted to run the targeted pytest subset with `pytest -q tests/test_lcm_engine.py -k "tool_schemas or handle_grep_source_filter_in_current_session_includes_only_matching_summaries or handle_grep_source_filter_excludes_unrelated_summaries_in_mixed_source_session or handle_grep_unknown_source_filter_matches_unknown_messages_and_summaries"`, but test collection failed due to `ModuleNotFoundError: No module named 'agent'` so the suite could not be executed in this environment.
- Compiled modified files with `python -m compileall schemas.py tools.py tests/test_lcm_engine.py`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e4869a2854832bb83e347c441c2388)

### DIFF
--- a/schemas.py
+++ b/schemas.py
@@ -6,9 +6,8 @@ LCM_GREP = {
         "Search the current session's conversation history — raw messages AND summaries across all depths — "
         "to recover details from earlier in the active conversation, even if those turns were compacted. "
         "Prefer this for intra-session recall. If the user is asking about an earlier separate conversation "
-        "or broad cross-session history, prefer session_search instead. When you intentionally want the LCM "
-        "store across sessions, use session_scope='all' explicitly. Returns matches with depth labels showing "
-        "where each result lives."
+        "or broad cross-session history, prefer session_search instead. Returns matches with depth labels "
+        "showing where each result lives."
     ),
     "parameters": {
         "type": "object",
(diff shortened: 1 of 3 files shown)


stephenschoettler (changes requested):
Thanks for tightening lcm_grep to the active session only. The implementation direction looks right, but this branch is not ready to merge because the test migration is incomplete.

Blocking issue:
- tests/test_lcm_engine.py:1968
- tests/test_lcm_engine.py:2008

Both tests were changed from session_scope="all" to session_scope="current", but the seeded rows are still written into non-active sessions:
- mixed-session
- unknown-session

The fixture keeps engine._session_id = "test-session", so lcm_grep correctly searches the current session and returns no matches. That makes the assertions fail for the right reason.

I reproduced this locally with:
- python -m pytest tests/test_lcm_engine.py -k 'source_filter_excludes_unrelated_summaries_in_mixed_source_session or unknown_source_filter' -q

Result:
- 2 failed

Suggested fix:
- set engine._session_id to the seeded session before invoking lcm_grep, or
- seed the test data into the active session instead

Once those tests are updated, this should be in much better shape.

stephenschoettler (approved):
Reviewed locally. LGTM.

PR #54, opened 2026-04-20
fix: bound LIKE fallback search queries to prevent unbounded fetch

Tosko4:
### Motivation
- The LIKE fallback used for CJK/emoji queries executed unbounded `SELECT *` queries and materialized all matching rows in memory, which can be abused to cause high CPU/memory usage and denial-of-service. 
- Add a minimal, low-risk mitigation that keeps the fallback behavior but prevents full-table materialization by bounding the SQL fetch size.

### Description
- Compute a fetch bound with `compute_search_fetch_limit(limit, terms, phrases)` and use it as a SQL `LIMIT ?` parameter in `MessageStore._search_like` so the fallback `SELECT` is no longer unbounded (`store.py`).
- Apply the same change to `SummaryDAG._search_like`, adding the computed fetch limit and `LIMIT ?` to the summary-node LIKE fallback query (`dag.py`).
- Add regression tests in `tests/test_lcm_core.py` that trigger the LIKE fallback (emoji queries) and assert the executed SQL includes a `LIMIT` clause, while preserving existing scoring/sorting behavior.

### Testing
- Ran `pytest -q tests/test_lcm_core.py -k "search_like_fallback_applies_sql_limit"` and the new checks passed (`2 passed`).
- Ran `pytest -q tests/test_lcm_core.py -k "cjk_queries_fall_back_with_aligned_sort_modes or emoji_queries_fall_back_with_aligned_sort_modes"` to confirm existing fallback semantics remain correct and those tests passed (`4 passed`).
- Verified the changed code still performs the in-Python scoring/sorting and then returns the top `limit` results so external behavior is preserved while preventing unbounded result sets.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e531f2ef48832bbe8b9f46ee153758)

### DIFF
--- a/dag.py
+++ b/dag.py
@@ -401,6 +401,7 @@ class SummaryDAG:
         phrases = extract_quoted_phrases(query)
         if not terms:
             return []
+        fetch_limit = compute_search_fetch_limit(limit, terms, phrases)
 
         where: list[str] = ["summary IS NOT NULL"]
         args: list[Any] = []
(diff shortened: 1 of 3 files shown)


stephenschoettler (approved):
Reviewed locally. LGTM.

PR #55, opened 2026-04-20
fix: prevent cross-session rollover carry-over

Tosko4:
### Motivation
- Close a security gap where `rollover_session` could move retained DAG summary nodes from an arbitrary `old_session_id` into a new session when the caller supplied attacker-controlled IDs, enabling cross-session data exposure.
- Ensure carry-over only happens when the engine was actually bound to the supplied `old_session_id` prior to rollover to preserve session isolation.

### Description
- Hardened `LCMEngine.rollover_session` to compute the currently bound session (`self._session_id`) and only proceed with retained-summary carry-over when `old_session_id` equals that bound session; otherwise the carry-over is skipped and a warning is logged (changes in `engine.py`).
- Kept existing behaviour for ignored/stateless sessions and the `carry_over_context` flag, so no-op paths remain unchanged.
- Added a regression test `test_rollover_session_skips_carry_over_when_old_session_is_not_bound` in `tests/test_lcm_engine.py` that verifies a mismatched `old_session_id` does not move victim-session nodes into an attacker-controlled new session.

### Testing
- Ran `python -m py_compile engine.py tests/test_lcm_engine.py`, which succeeded.
- Ran `pytest -q tests/test_lcm_engine.py -k 'rollover_session or carry_over'`, which failed during collection due to a missing test dependency error (`ModuleNotFoundError: No module named 'agent'`) in this environment, so tests could not be executed here.
- The change is minimal and unit-test additions exercise the new guard to prevent cross-session DAG reassignment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e5333b57fc832bb7929717f7fbcedd)

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -626,6 +626,10 @@ class LCMEngine(ContextEngine):
         """
         previous_messages = previous_messages or []
         conversation_id = self._conversation_id or old_session_id or new_session_id
+        bound_session_id = self._session_id
+        can_carry_over = bool(
+            old_session_id and bound_session_id and old_session_id == bound_session_id
+        )
 
         if old_session_id:
             self.on_session_end(old_session_id, previous_messages)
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Reviewed locally. LGTM.

PR #56, opened 2026-04-20
Harden /lcm slash command registration with explicit opt-in

Tosko4:
### Motivation
- The `/lcm` slash surface was being registered unconditionally when the host exposed `register_command`, allowing any caller with slash access to reach handlers that expose DB/session metadata and create on-demand SQLite backups.
- This presents confidentiality and availability risks (metadata leakage and unbounded backup creation) in multi-tenant or untrusted gateway contexts.

### Description
- Gate slash-command registration behind an explicit environment opt-in: `LCM_ENABLE_SLASH_COMMAND` (default `false`) with a small boolean parser helper `_env_flag_enabled` in `__init__.py` and a log message when the surface is disabled.
- Only register `/lcm` when `register_command` is callable and `LCM_ENABLE_SLASH_COMMAND` is enabled; otherwise skip registration and log that it is disabled or unavailable.
- Add two tests in `tests/test_lcm_command.py` to assert that the slash command is not registered by default and that it is registered when `LCM_ENABLE_SLASH_COMMAND=1`.
- Document the new environment variable and default-off behavior in `README.md`.

### Testing
- Ran `python -m py_compile __init__.py tests/test_lcm_command.py` which succeeded to validate syntax for the changed files.
- Ran `pytest -q tests/test_lcm_command.py` which failed during collection with an `ImportError` (`cannot import name 'LCMEngine' from 'hermes_lcm.engine'`) due to the test environment lacking the full runtime package dependencies, not due to the registration change itself.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e5348d8a4c832bbefb0ac4356879ce)

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ Environment variables (all optional):
 | `LCM_EXPANSION_TIMEOUT_MS` | `120000` | Timeout for `lcm_expand_query` answer synthesis |
 | `LCM_DATABASE_PATH` | `~/.hermes/lcm.db` | SQLite database path (auto profile-scoped) |
 | `LCM_NEW_SESSION_RETAIN_DEPTH` | `2` | DAG depth retained after `/new` (`-1` = all, `0` = none, `2` = keep d2+) |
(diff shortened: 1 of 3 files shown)


stephenschoettler (approved):
Reviewed locally. LGTM.

PR #59, opened 2026-04-20
fix: scope `/lcm doctor retention` to active session

Tosko4:
### Motivation
- The `doctor retention` command previously enumerated every stored `session_id` and printed per-session metadata, enabling cross-session metadata disclosure in multi-tenant setups. 
- The goal is to prevent metadata enumeration while preserving the read-only retention analysis for the currently active session. 
- Provide a minimal, defensive change that avoids broad behavioral changes to other LCM commands.

### Description
- Early-return the retention scan when no active session is bound by checking `getattr(engine, "_session_id", "")` and returning an empty analysis. 
- Restrict the retention SQL query to the active session by adding `WHERE s.session_id = ?` and passing the `session_id` parameter. 
- Add a user-facing note `note: retention analysis is scoped to the active session only` to `_doctor_retention_text` output. 
- Update assertions in `tests/test_lcm_command.py` to reflect that `doctor retention` no longer exposes other sessions and to assert the new scoped behavior.

### Testing
- Ran `python -m compileall -q command.py tests/test_lcm_command.py`, which completed successfully to validate syntax. 
- Ran `pytest -q tests/test_lcm_command.py` in this environment but collection failed due to an import error (`ImportError: cannot import name 'LCMEngine' from 'hermes_lcm.engine'`), so the updated tests were not executed end-to-end here. 
- The change is small and covered by updated unit tests in `tests/test_lcm_command.py`, which should pass in a normal test environment where `LCMEngine` imports correctly.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e5333a8a68832bb571cf94fa415c3c)

### DIFF
--- a/command.py
+++ b/command.py
@@ -179,6 +179,18 @@ def _scan_clean_candidates(engine) -> dict[str, Any]:
 def _scan_retention_candidates(engine) -> dict[str, Any]:
     conn = engine._store._conn
     now = datetime.now().timestamp()
+    session_id = getattr(engine, "_session_id", "")
+    if not session_id:
+        return {
+            "error": None,
+            "sessions": [],
+            "sessions_analyzed": 0,
+            "stale_sessions_30d": 0,
+            "stale_sessions_90d": 0,
+            "retained_tokens_30d": 0,
+            "retained_tokens_90d": 0,
+            "protected_count": 0,
+        }
     try:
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Reviewed locally. LGTM.

PR #64, opened 2026-04-20
fix: validate numeric lcm_expand_query arguments

Tosko4:
### Motivation
- `lcm_expand_query` directly cast untrusted inputs (`max_tokens`, `max_results`, and `node_ids`) with `int(...)`, which can raise `ValueError` on malformed input and crash the tool call, causing an availability/DoS condition.
- The change adds defensive parsing so malformed numeric inputs return structured JSON errors instead of raising exceptions.

### Description
- Add a local helper `_parse_int_arg` in `lcm_expand_query` to parse integer arguments and produce an error message for invalid values.
- Validate `max_tokens` and `max_results` using the helper and return `{"error": "... must be an integer"}` when invalid.
- Validate each entry in `node_ids` before lookup and return `{"error": "node_ids must contain only integers"}` if any entry is non-numeric.
- Add regression tests in `tests/test_lcm_engine.py` that exercise invalid `max_tokens`, `max_results`, and `node_ids` inputs and assert the new error responses.

### Testing
- Static sanity check: ran `python -m py_compile tools.py tests/test_lcm_engine.py` which succeeded.
- Attempted to run the new test selection with `PYTHONPATH=. pytest -q tests/test_lcm_engine.py -k 'expand_query_rejects_non_numeric'`, but test collection failed with `ModuleNotFoundError: No module named 'agent'` in this environment, so `pytest` could not be completed here.
- The change is minimal and constrained to input validation and unit tests; the new tests should pass in a normal test environment where test dependencies are available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e5348aaa48832bb10d40109a7c788a)

### DIFF
--- a/tests/test_lcm_engine.py
+++ b/tests/test_lcm_engine.py
@@ -3077,6 +3077,35 @@ class TestEngineTools:
         assert result["node_ids"] == [node_id]
         assert result["matches"]
 
+    def test_handle_expand_query_rejects_non_numeric_limits(self, engine):
+        result = json.loads(
+            engine.handle_tool_call(
+                "lcm_expand_query",
+                {"query": "docker", "prompt": "What was the plan?", "max_tokens": "invalid"},
+            )
+        )
+
+        assert result["error"] == "max_tokens must be an integer"
+
+        result = json.loads(
+            engine.handle_tool_call(
+                "lcm_expand_query",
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Reviewed locally. LGTM.

Issue #68, opened 2026-04-21
Hermes itself's Compression threshold


xlionjuan:
It looks like Hermes' default Compression threshold 0.5 is affecting the usability, I found the context window being compressed and `dag_nodes` is 0, and I getting it multiple times, agent become stupid when compression happened

So I increased Hermes itself's Compression threshold to 0.8, which is a little larger then `LCM_CONTEXT_THRESHOLD`'s 0.75, and it is working

After lcm triggered, my agent is working better compared the previous, means lcm is working, but I found out that the system behavior is weird, sometime the lcm still not working, maybe 0.8 is too close to 0.75 so the system forcing to compression when doing long run? Can you clarify the content window threshold related settings at the README?

Sorry for my bad expression here since I'm not native speaker and this issue is too hard to explain.

---

Another question, is this normal? `hermes plugin`'s Context Engine page shows lcm not found, but my agent can see the lcm tool, I can see `/lcm` tool when I enabled `LCM_ENABLE_SLASH_COMMAND`.

<img width="1938" height="1410" alt="Image" src="https://github.com/user-attachments/assets/992e89ea-299b-4540-af08-12b526959304" />

<img width="1938" height="1410" alt="Image" src="https://github.com/user-attachments/assets/95d18e3e-d502-4571-9434-e53f720882f5" />

Thank you.

Tosko4:
I dug into this and I think there are two separate things mixed together here.

1. **Threshold ownership / tuning clarity**
   - This part *does* belong in `hermes-lcm`.
   - When Hermes is running with `context.engine: lcm`, the active compaction threshold should be `LCM_CONTEXT_THRESHOLD`.
   - Hermes core `compression.threshold` belongs to the built-in compressor, not to LCM.
   - Hermes core `compression.enabled` is still the global on/off gate for compaction in general, so that still needs to stay enabled when using LCM.

I opened a small docs PR here to make that boundary more explicit and to document the confusing signals from this report:
- PR: https://github.com/stephenschoettler/hermes-lcm/pull/69

2. **`hermes plugins` / context-engine UI saying `lcm (not found)` while the runtime still exposes LCM tools**
   - That looks like a **Hermes Agent host-side discovery / UI problem**, not an `hermes-lcm` storage or compaction bug.
[...]

PR #69, opened 2026-04-21
docs: clarify threshold ownership and host troubleshooting

Tosko4:
## Summary
- clarify that `LCM_CONTEXT_THRESHOLD` owns LCM compaction behavior when `context.engine: lcm` is active
- document that Hermes core `compression.threshold` belongs to the built-in compressor, while `compression.enabled` remains the global compaction gate
- add troubleshooting notes for the two operator-confusing signals reported in #68: `Context Engine: lcm (not found)` and unbound `/lcm status` output right after restart

## Why
- issue #68 mixes together an LCM threshold-tuning question and a Hermes host discovery/UI symptom
- the plugin README did not state the threshold ownership boundary sharply enough
- making that explicit should reduce false debugging paths and make it clearer which part belongs in Hermes host code vs the plugin repo

## Validation
- `git diff --check`
- reviewed the current Hermes host integration paths and verified the docs match the live contract:
  - `run_agent.py` context-engine selection and LCM fallback path
  - `hermes_cli/plugins_cmd.py` context-engine discovery UI path
  - `hermes-lcm` registration path in `__init__.py`

## Notes
- docs-only change, no plugin runtime behavior changed
- the `Context Engine: lcm (not found)` symptom described in #68 still appears to be a Hermes host-side discovery/UI problem when the runtime is otherwise working
- closes the documentation part of #68, but not the Hermes host-side UI/discovery part

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,34 @@ At runtime the tool list should include:
 - `lcm_status`
 - `lcm_doctor`
 
+## Troubleshooting signals
+
+### `hermes plugins` shows `lcm (not found)` but LCM tools still exist
+
+If all of these are true:
+
+- `plugins.enabled` contains `hermes-lcm`
+- `context.engine: lcm` is set
+- the runtime exposes LCM tools (or `/lcm` if enabled)
+
+then LCM itself is loaded.
+
+In that situation, a `Context Engine: lcm (not found)` line in `hermes plugins` is a **Hermes host discovery/UI issue**, not proof that `hermes-lcm` failed to load.
+
+Treat it as a host-side mismatch between the picker/status surface and the live runtime. The fix belongs in Hermes host discovery/UI, not in LCM storage or compaction logic.
+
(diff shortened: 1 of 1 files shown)


stephenschoettler (approved):
Approved.

This is a good docs-only clarification.

It cleanly separates:
- `LCM_CONTEXT_THRESHOLD` as the active compaction threshold when `context.engine: lcm` is loaded
- Hermes core `compression.threshold` as the built-in compressor setting
- Hermes core `compression.enabled` as the global compaction gate

The troubleshooting notes are also useful, especially around:
- `Context Engine: lcm (not found)` being a host discovery/UI signal rather than proof the runtime failed
- `/lcm status` showing unbound-looking fields immediately after restart before a live session is initialized

Important nuance: this clarifies the documentation side of #68, but it does not resolve the underlying runtime symptom reported there. The issue should stay open for technical triage.

LGTM for merge as a docs improvement.

PR #70, opened 2026-04-23
fix: reset session-scoped runtime on rebind

stephenschoettler:
## Summary
- reset session-scoped runtime counters when the same `LCMEngine` instance is rebound to a different session
- add regression coverage for the stale `compression_count` leak seen in #68
- verify `/lcm status` reports a clean new session after rebind instead of inheriting prior-session compaction state

## Why
- #68 exposed a real operator-facing failure mode: a reused engine instance could report `compression_count > 0` for a fresh session with `store_messages: 0` and `dag_nodes: 0`
- the host-side singleton reuse is still the larger integration problem, but `LCMEngine` was also leaking session-scoped runtime state across `on_session_start()` rebinds
- resetting those counters on session change makes the engine more robust and keeps status output honest without claiming to solve host lifecycle ownership by itself

## Validation
- `pytest tests/test_lcm_engine.py::TestEngineABC::test_on_session_start_resets_session_scoped_runtime_when_binding_new_session tests/test_lcm_command.py::test_lcm_status_does_not_leak_prior_session_compaction_count_after_rebind -q`
- `pytest tests/test_lcm_engine.py tests/test_lcm_command.py -q`
- `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q`
- `ulimit -n 4096 && pytest -q`
- `python -m compileall -q .`
- `git diff --check`

## Notes
- Refs #68
- This is intentionally scoped to the LCM-side leak: `compression_count`, token usage, probe state, and compaction cursors now reset when the engine binds a different session id
- It does not claim to fix Hermes host singleton lifecycle reuse by itself; it hardens `hermes-lcm` so the stale-status symptom is no longer inherited across rebinds

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -536,13 +536,29 @@ class LCMEngine(ContextEngine):
         if self._has_raw_backlog_debt():
             self._lifecycle.clear_debt(self._conversation_id)
 
-    def on_session_start(self, session_id: str, **kwargs) -> None:
-        self._session_id = session_id
-        self._session_platform = str(kwargs.get("platform") or "")
-        self._ingest_cursor = 0
+    def _reset_session_scoped_runtime_state(self) -> None:
+        self.compression_count = 0
+        self.last_prompt_tokens = 0
+        self.last_completion_tokens = 0
+        self.last_total_tokens = 0
         self._last_compacted_store_id = 0
+        self._ingest_cursor = 0
+        self._context_probed = False
+        self._context_probe_persistable = False
         self._last_overflow_recovery_failed = False
         self._last_condensation_suppressed_reason = ""
+
+    def on_session_start(self, session_id: str, **kwargs) -> None:
+        previous_session_id = self._session_id
+        if previous_session_id and previous_session_id != session_id:
+            self._reset_session_scoped_runtime_state()
+        else:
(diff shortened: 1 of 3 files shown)


Tosko4:
Post-merge review note: this was still sitting in my review queue, so I did a local pass on the merged change for the record.

I reviewed the landed commit locally and did not find any blockers. The session-scoped runtime reset on cross-session rebind looks correct, and the added regression coverage matches the stale `compression_count` symptom described in #68.

Local validation I reran:
- targeted regressions: `2 passed`
- `tests/test_lcm_engine.py tests/test_lcm_command.py -q`: `146 passed`
- full suite: `275 passed`
- `python -m compileall -q .`: clean
- `git diff --check`: clean

One non-blocking follow-up suggestion: there are now two partially overlapping reset paths, `_reset_session_scoped_runtime_state()` and `on_session_reset()` (plus `super().on_session_reset()`). The current change looks correct, but it may be worth aligning those paths later to reduce drift risk if more session-scoped runtime fields get added.

Given that this is already merged, I'd treat this as a post-merge LGTM rather than a gating review.

Tosko4 (approved):

PR #72, opened 2026-04-23
fix: route provider-prefixed LCM models

### LINKED ISSUE #71: Support provider-prefixed LCM summary/expansion models
## Summary
`hermes-lcm` lets users set `LCM_SUMMARY_MODEL` / `LCM_EXPANSION_MODEL`, but provider-prefixed values like `cerebras/gpt-oss-120b` are treated as plain model strings instead of explicit provider+model routing.

That becomes a real problem when Hermes chat compression is intentionally configured to follow the main model while LCM summarization needs a separate provider.

## Reproduction
Set:

```bash
LCM_SUMMARY_MODEL=cerebras/gpt-oss-120b
LCM_EXPANSION_MODEL=cerebras/gpt-oss-120b
```

and use Hermes auxiliary compression with:

```yaml
auxiliary:
  compression:
    provider: main
    model: ''
```

## Actual behavior
LCM passes `model="cerebras/gpt-oss-120b"` into Hermes auxiliary calls, but not an explicit `provider="cerebras"`.

Depending on the caller's auxiliary routing, the wrong provider can be selected or the shared auxiliary lane can be forced into an unintended config shape.

## Expected behavior
When an `LCM_*MODEL` value starts with a known provider prefix, `hermes-lcm` should split it into:
- `provider=<prefix>`
- `model=<rest>`

while leaving normal model slugs with a slash (for example `meta-llama/Llama-3.3-70B-Instruct`) untouched.

## Proposed fix
Add a small routing helper in `hermes-lcm` and use it in:
- `escalation.py`
- `extraction.py`
- `tools.py`

Also add regression tests covering:
- provider-prefixed values
- plain slugs that contain `/` but are not provider prefixes

## Why this matters
It cleanly separates:
- Hermes chat compression routing
from
- LCM summary / expansion routing

which avoids configuration bleed-through.

stephenschoettler:
## Summary
- Add a conservative LCM model routing helper for provider-prefixed overrides
- Route summary, extraction, and expansion calls with explicit provider/model kwargs when applicable
- Preserve aggregator-style model slugs like meta-llama/... and google/... as model-only overrides
- Document provider-prefixed LCM model overrides

## Test Plan
- pytest -q

Closes #71

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ Environment variables (all optional):
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_THRESHOLD_CHARS` | `12000` | Character threshold above which tool results are externalized |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_PATH` | `~/.hermes/lcm-large-outputs` | Override storage directory for externalized payloads |
 | `LCM_LARGE_OUTPUT_TRANSCRIPT_GC_ENABLED` | `false` | Opt-in rewrite of already-externalized summarized tool-result transcript rows to compact GC placeholders |
-| `LCM_SUMMARY_MODEL` | *(auxiliary)* | Override model for summarization. Direct provider prefixes such as `cerebras/gpt-oss-120b` route as `provider=cerebras`, `model=gpt-oss-120b`; aggregator model slugs such as `meta-llama/...` stay model-only. |
+| `LCM_SUMMARY_MODEL` | *(auxiliary)* | Override model for summarization. The explicit direct prefix `cerebras/gpt-oss-120b` routes as `provider=cerebras`, `model=gpt-oss-120b`; aggregator model slugs such as `meta-llama/...` and `anthropic/...` stay model-only. |
 | `LCM_EXPANSION_MODEL` | *(summary model / auxiliary)* | Override model for `lcm_expand_query` synthesis. Supports the same provider-prefixed routing as `LCM_SUMMARY_MODEL`. |
 | `LCM_SUMMARY_TIMEOUT_MS` | `60000` | Timeout for a single model-backed summarization call |
 | `LCM_EXPANSION_TIMEOUT_MS` | `120000` | Timeout for `lcm_expand_query` answer synthesis |
--- a/README.md
+++ b/README.md
@@ -252,8 +252,8 @@ Environment variables (all optional):
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_THRESHOLD_CHARS` | `12000` | Character threshold above which tool results are externalized |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_PATH` | `~/.hermes/lcm-large-outputs` | Override storage directory for externalized payloads |
 | `LCM_LARGE_OUTPUT_TRANSCRIPT_GC_ENABLED` | `false` | Opt-in rewrite of already-externalized summarized tool-result transcript rows to compact GC placeholders |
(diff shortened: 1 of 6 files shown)


Tosko4:
I think this still needs one adjustment before merge.

The parser change is intentionally conservative, which is good for not breaking aggregator model slugs. I checked it against the current Hermes model catalog and representative slash-bearing model IDs like:

- `anthropic/...`
- `google/...`
- `openai/...`
- `deepseek/...`
- `nvidia/...`
- `meta-llama/...`
- `qwen/...`
- `moonshotai/...`

Those remain model-only, so the current implementation does not appear to break existing aggregator-style model IDs.

The problem is the direct `cerebras/...` route itself. In the current Hermes host, `cerebras` is not a built-in provider in the provider registry. So this PR turns:

```text
LCM_SUMMARY_MODEL=cerebras/gpt-oss-120b
```

into:

```python
provider="cerebras"
model="gpt-oss-120b"
```

but that only works if the user has also configured a named custom provider called `cerebras` in Hermes config, for example via `providers:` / `custom_providers`. Without that, Hermes resolves `provider="cerebras"` as an unknown provider.

So I do not think the README should imply that setting only `LCM_SUMMARY_MODEL=cerebras/gpt-oss-120b` is enough. Either:

1. document that `cerebras/...` requires a named custom provider called `cerebras`, or
2. add host-side support for Cerebras as an actual provider before treating this as a working direct route.

I would also suggest adding a regression test or fixture for this contract: `cerebras/...` should only be considered a direct-provider route when that provider can actually be resolved, otherwise this creates a config that looks valid but fails at runtime.

Tosko4 (changes requested):
See previous comment

stephenschoettler:
Update pushed in `eb2fcf7`.

Summary:
- Provider-prefixed values now split only when the provider is resolvable through Hermes built-ins or named custom providers.
- Ambiguous OpenRouter/org slugs like `anthropic/...`, `google/...`, `meta-llama/...`, `qwen/...`, and `deepseek/...` stay model-only.
- Added focused tests for unresolved vs resolvable `cerebras/gpt-oss-120b` routing and summary/extraction/expansion call kwargs.

Validation:
- `python -m pytest tests/test_lcm_core.py::TestModelRouting -q`
- `python -m pytest tests/test_lcm_core.py::TestProviderPrefixedAuxiliaryCalls -q`
- `python -m pytest tests/test_lcm_core.py -q` (`134 passed`)
- `python -m compileall -q .`
- `git diff --check`
- GitHub Actions passed on 3.11, 3.12, and 3.13

Tosko4 (approved):
Re-reviewed after `eb2fcf7`.

This resolves my earlier concern. The routing now only splits `cerebras/gpt-oss-120b` into `provider=cerebras`, `model=gpt-oss-120b` when the Hermes host can actually resolve `cerebras` as a built-in or named custom provider. On an unresolved host it remains model-only, which avoids creating an unknown-provider runtime failure. The OpenRouter/org-style slugs I was worried about (`meta-llama/...`, `anthropic/...`, `google/...`, etc.) also remain model-only.

I also checked the updated README language, the summary/extraction/expansion call sites, and the focused tests. Locally I validated the PR rebased cleanly onto current `origin/main` and ran:

- `python -m pytest tests/test_lcm_core.py::TestModelRouting tests/test_lcm_core.py::TestProviderPrefixedAuxiliaryCalls -q` → 10 passed
- `python -m pytest tests/test_lcm_core.py -q` → 134 passed
- `python -m pytest -q` → 286 passed
- `python -m compileall -q .`
- `git diff --check`

No blocking issues from me now. The PR branch is currently behind `main`, so it still needs to be updated before merge, but the fixed shape looks good.

PR #73, opened 2026-04-23
fix: treat legacy blank-source rows as healthy

stephenschoettler:
## Summary
- Stop treating legacy blank-source rows as an action-required LCM doctor warning
- Keep source lineage counts visible as observations/backcompat metadata
- Mark `source_lineage_hygiene` healthy when backcompat normalization is working

## Test Plan
- `pytest tests/test_lcm_command.py::test_lcm_doctor_reports_legacy_blank_source_as_observation_without_warning tests/test_lcm_engine.py::TestEngineTools::test_handle_doctor_treats_legacy_blank_source_rows_as_healthy -q`
- `ulimit -n 8192; pytest -q`

### DIFF
--- a/command.py
+++ b/command.py
@@ -488,8 +488,9 @@ def _doctor_text(engine) -> str:
     if source_stats.get("error"):
         observations.append(f"source_lineage_error: {source_stats['error']}")
     if source_stats["legacy_blank_source_messages"]:
-        recommended_actions.append("review legacy blank-source rows before any destructive cleanup or migration step")
-        recommended_actions.append("treat `source=unknown` as the back-compat filter until legacy blank-source rows are normalized")
+        observations.append(
+            "legacy blank-source rows are normalized as `source=unknown` for back-compat filters"
+        )
 
     if clean_scan.get("protected_count"):
         observations.append(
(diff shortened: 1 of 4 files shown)


Tosko4:
I reviewed this both as a code change and as a product/architecture fit for hermes-lcm on top of Hermes Agent.

I think this direction makes sense.

The important distinction here is that legacy blank-source rows are not the same thing as broken source-lineage behavior if the runtime consistently normalizes them into the `unknown` bucket for filtering/backcompat. In that case, surfacing the count as an observation is useful, but treating it as action-required creates operator noise and makes existing migrated installs look less healthy than they actually are.

This PR keeps the right signals visible:

- `legacy_blank_source_messages` is still reported
- `effective_unknown_messages` still includes legacy blank rows
- doctor output still explains the backcompat normalization
- actual source-lineage read failures still fail/warn normally

It also avoids adding extra migration or cleanup machinery just to silence the warning, which is the right shape for this repo. For hermes-lcm, I would rather keep doctor output honest and low-noise than push users toward unnecessary cleanup work when the compatibility layer already handles the legacy state.

I validated locally:

```text
pytest tests/test_lcm_command.py::test_lcm_doctor_reports_legacy_blank_source_as_observation_without_warning tests/test_lcm_engine.py::TestEngineTools::test_handle_doctor_treats_legacy_blank_source_rows_as_healthy -q
2 passed

251 passed

pytest -q
276 passed

python -m compileall -q .
```

So from my side: this looks like the right product behavior, not just a passing code change.

Tosko4 (approved):

PR #74, opened 2026-04-25
fix: accept valid integer node_ids in lcm_expand_query

catgodtw:
PR #64 introduced a regression: every call with `node_ids` returns `{"error": "node_ids must contain only integers"}`, including valid integer inputs like `[1]`.

Root cause: the `_parse_int_arg` helper added in PR #64 reads from the args dict via `args.get(name, default)`, which always returns the original list when called inside the per-element loop with the same name. `int([1])` raises TypeError, so the loop reports failure for every element regardless of value.

The existing regression test only covered the rejection path (non- numeric strings), which still works because `int(["not-a-number"])` also fails — it just fails for the wrong reason.

Replace the inner-loop call with a direct `int(node_id)` wrapped in a try/except. Same error message is preserved for non-numeric inputs.

Add a regression test that constructs a real DAG node and asserts that calling `lcm_expand_query` with `node_ids: [<that node_id>]` returns the synthesized answer instead of the validation error.

Reproduction (before this fix):

    engine.handle_tool_call(
        "lcm_expand_query",
        {"node_ids": [1], "prompt": "What was discussed?"},
    )
    # -> {"error": "node_ids must contain only integers"}

After:

    # -> {"prompt": ..., "answer": "...", "node_ids": [1], "matches": [...]}

### DIFF
--- a/tests/test_lcm_engine.py
+++ b/tests/test_lcm_engine.py
@@ -3371,6 +3371,37 @@ class TestEngineTools:
 
         assert result["error"] == "node_ids must contain only integers"
 
+    def test_handle_expand_query_accepts_valid_integer_node_ids(self, engine, monkeypatch):
+        engine._store.append("test-session", {"role": "user", "content": "Discussed docker rollout plan"})
+        node_id = engine._dag.add_node(
+            SummaryNode(
+                session_id="test-session",
+                depth=0,
+                summary="Docker rollout summary",
+                token_count=10,
+                source_token_count=20,
+                source_ids=[1],
+                source_type="messages",
+                created_at=0,
+            )
+        )
+
+        def fake_synthesize(*, prompt, context_blocks, model, max_tokens, timeout):
+            return "Expansion answer"
+
+        monkeypatch.setattr(lcm_tools, "_synthesize_expansion_answer", fake_synthesize)
+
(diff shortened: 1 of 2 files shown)


Tosko4 (approved):
Reviewed after rebasing onto current `main`.

This is a narrow regression fix for `lcm_expand_query` explicit `node_ids`: valid integer node IDs were rejected because the per-element loop reused `_parse_int_arg("node_ids", ...)`, which reads the full original list from `args` instead of the current element.

The change is the right shape for hermes-lcm: plugin-local bug fix, no host changes, no new config surface, no dependencies, no architecture drift.

Validation run locally on the rebased branch:
- `python3 -m pytest tests/test_lcm_engine.py::TestEngineTools::test_handle_expand_query_rejects_non_numeric_node_ids tests/test_lcm_engine.py::TestEngineTools::test_handle_expand_query_accepts_valid_integer_node_ids -q` → 2 passed
- `python3 -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` → 259 passed
- `python3 -m pytest -q` → 284 passed
- local merge trial against current `origin/main`

CI is green on Python 3.11, 3.12, and 3.13.

LGTM.

PR #75, opened 2026-04-25
fix: strip thinking-model reasoning tags before persisting summaries

catgodtw:
Some thinking models (MiniMax-M2.7, GLM-5.1, Qwen QwQ, DeepSeek R1) inline their reasoning inside <think>...</think> blocks within message.content. _call_llm_for_summary returns this content directly to the engine, which persists it as the summary node text.

This causes two downstream problems:

1. The stored summary contains the model's internal reasoning, not just the requested summary, so it consumes tokens that should have been the actual summary.
2. The reasoning text frequently quotes the summarizer system prompt verbatim. When lcm_expand_query later feeds that summary back to the model as context for synthesis, the model treats it as a summarization task and produces a reply that talks about the summarization process rather than answering the user's actual question.

Reproduction with MiniMax-M2.7-highspeed: the stored DAG summary contained a verbatim <think>...</think> block quoting the summarizer's bullet-point instructions, and a subsequent lcm_expand_query call produced an answer that re-stated the summarization instructions instead of answering the asked question.

The host hermes-agent already strips these tags from primary responses in run_agent.py around line 3098-3100; the auxiliary client used by LCM does not, so the same pattern needs to apply locally before the summary text is returned.

Add _strip_reasoning_blocks() with a regex matching <think>, <thinking>, <reasoning>, <thought>, and <REASONING_SCRATCHPAD> tags (case-insensitive, DOTALL). Call it on the response content before strip() and return.

Tests:
- Unit tests for each supported tag, multiple blocks, idempotency, and preservation of content with unrelated angle brackets.
- Integration test that monkeypatches agent.auxiliary_client.call_llm to return contaminated content and asserts the returned summary is clean.

All 5 tests pass:
  pytest tests/test_lcm_engine.py::TestEscalationStripReasoning -q 5 passed in 0.25s

### DIFF
--- a/escalation.py
+++ b/escalation.py
@@ -9,11 +9,34 @@ Each level checks if Tokens(summary) < Tokens(source). If not, escalates.
 
 import inspect
 import logging
+import re
 from typing import List, Optional
 
 from .tokens import count_tokens
 
 logger = logging.getLogger(__name__)
+
+
+# Strip inline reasoning blocks emitted by thinking models (MiniMax-M2.7,
+# GLM-5.1, Qwen QwQ, DeepSeek R1, etc.) before persisting summary text.
+# Without this, the reasoning content — which often quotes the summarizer
+# system prompt verbatim — gets stored as the summary and later confuses
+# lcm_expand_query, which feeds the summary back to the model as context.
+# Tags mirror the set handled in hermes-agent run_agent.py.
+_THINK_BLOCK_RE = re.compile(
+    r"<(?P<tag>think|thinking|reasoning|thought|REASONING_SCRATCHPAD)\s*>"
+    r".*?"
+    r"</(?P=tag)\s*>",
+    re.IGNORECASE | re.DOTALL,
+)
+
+
+def _strip_reasoning_blocks(text: str) -> str:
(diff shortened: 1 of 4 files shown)


Tosko4 (approved):
Reviewed the scope and local validation.

This looks like the right shape for hermes-lcm: narrow plugin-side sanitization of derived auxiliary LLM outputs, no host changes, no new config surface, no dependencies, and no mutation of raw conversation data. The fix covers the three current call-then-return/persist paths: summary generation, expand-query synthesis, and pre-compaction extraction.

Validation run locally:
- `python3 -m pytest tests/test_lcm_engine.py::TestEscalationStripReasoning -q` → 7 passed
- `python3 -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` → 258 passed
- `python3 -m pytest -q` → 283 passed
- local merge trial against current `origin/main`

The only tradeoff I see is that literal reasoning-tag examples inside a generated summary/extraction/answer would be stripped too, but this is limited to derived auxiliary output and is a reasonable safety tradeoff for preventing thinking-model scratchpad leakage.

LGTM.

PR #81, opened 2026-04-29
fix: make cleanup apply rollback atomically

Tosko4:
## Summary
- make `/lcm doctor clean apply` delete cleanup candidates through one SQLite transaction
- keep the existing safety gates: disabled by default, explicit `LCM_DOCTOR_CLEAN_APPLY_ENABLED=true`, backup-first, and pattern-limited candidates only
- return a rollback error with the already-created backup path if a post-backup delete fails

## Why
The existing cleanup apply path was backup-first, but it coordinated deletes through helper methods that committed independently. If a later delete step failed after an earlier table was already committed, operators could be left with half-cleaned LCM state.

This keeps the existing command shape and opt-in semantics, but makes the destructive part atomic across messages, summary nodes, and lifecycle rows.

## Validation
- TDD: added a failing rollback regression first using a `summary_nodes` delete trigger that raises after backup creation, then implemented the atomic delete path
- `python -m pytest tests/test_lcm_command.py -q` → `26 passed`
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_lcm_command.py tests/test_packaging_install.py -q` → `304 passed`
- `python -m pytest -q` → `304 passed`
- `python -m compileall -q .`
- `git diff --check`
- runtime smoke through the active plugin path with a temp DB and explicit cleanup opt-in confirmed a safe `cron*` candidate is backed up and removed from store, DAG, and lifecycle state

### DIFF
--- a/command.py
+++ b/command.py
@@ -618,6 +618,76 @@ def _doctor_retention_text(engine) -> str:
     return "\n".join(lines)
 
 
+def _delete_clean_candidates_atomically(engine, session_ids: set[str]) -> dict[str, int]:
+    """Delete cleanup candidates in one SQLite transaction.
+
+    All LCM tables live in the same SQLite database, but the store, DAG, and
+    lifecycle helpers use separate connections and commit internally. Cleanup
+    apply is destructive, so do the coordinated deletes on one connection to
+    avoid half-cleaned state if a later table delete fails.
+    """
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Approved after local review and green CI.

PR #83, opened 2026-04-29
fix: upgrade lifecycle rotation columns

Tosko4:
## Summary
- Add a regression test for existing lifecycle tables that are missing rotation timestamp columns.
- Extend `ensure_lifecycle_state_columns(...)` to add `last_rollover_at` and `last_reset_at` in place.

## Why
- Older lifecycle/checkpoint DBs can already have `schema_version = 4` and an existing `lcm_lifecycle_state` table, while still missing the later rotation timestamp columns.
- `LifecycleStateStore._row_to_state()` reads those fields unconditionally, so startup/session lookup can fail with `IndexError: No item with that key` before the engine can bind lifecycle state.
- This keeps the fix narrow: it repairs the upgrade path for existing lifecycle tables without adding new transcript import/bootstrap machinery.

## Validation
- `python -m pytest tests/test_lcm_core.py::TestLifecycleStateStore::test_init_upgrades_existing_lifecycle_table_with_rotation_columns -q` → failed before the fix with `IndexError: No item with that key`, then passed after the fix
- `python -m pytest tests/test_lcm_core.py::TestLifecycleStateStore -q` → `5 passed`
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` → `279 passed`
- `python -m pytest -q` → `304 passed`
- `python -m compileall -q .`
- `git diff --check`
- Runtime smoke via active plugin discovery with a legacy DB confirmed:
  - the active plugin loads from `/home/nabu/.hermes/dev/hermes-lcm`
  - missing `last_rollover_at` / `last_reset_at` columns are added
  - lifecycle state loads successfully
  - the existing frontier marker is preserved

## Notes
- No source rows are rewritten by this migration; it only adds nullable timestamp columns when they are missing.

### DIFF
--- a/db_bootstrap.py
+++ b/db_bootstrap.py
@@ -122,6 +122,10 @@ def ensure_lifecycle_state_columns(conn: sqlite3.Connection) -> None:
         conn.execute(
             "ALTER TABLE lcm_lifecycle_state ADD COLUMN last_maintenance_attempt_at REAL"
         )
+    if "last_rollover_at" not in columns:
+        conn.execute("ALTER TABLE lcm_lifecycle_state ADD COLUMN last_rollover_at REAL")
+    if "last_reset_at" not in columns:
+        conn.execute("ALTER TABLE lcm_lifecycle_state ADD COLUMN last_reset_at REAL")
 
 
 def mark_migration_step_complete(conn: sqlite3.Connection, step_name: str) -> None:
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Approved after local review and green CI.

PR #85, opened 2026-04-29
fix: preserve lifecycle reset boundaries

Tosko4:
## Summary

This PR hardens the LCM session lifecycle boundary handling around reset, manual new sessions, and compression-created session splits.

Changes included:

* Records reset boundaries without immediately finalizing same-session resets.
* Finalizes a pending reset only when a later bind proves the session changed.
* Preserves the newest known frontier when finalizing after reset continuation.
* Keeps compression continuations distinct from manual new-session pruning.
* Treats `rollover_session(..., boundary_reason="compression")` as a compression continuation for host compatibility.
* Documents the manual new-session versus compression boundary contract.

## Why

Hermes hosts do not all notify context engines through the same lifecycle path yet. Some older or fallback paths can call `on_session_reset()` and later `on_session_start(new_session_id)` without using `rollover_session(...)`.

LCM needs to remain safe in both cases:

* A plain same-session reset should stay current and keep accepting frontier updates.
* A later fresh session bind should finalize the previous lifecycle row correctly.
* Compression-created session splits should preserve the logical conversation and carry depth-zero DAG state forward, not behave like manual `/new` pruning.

## Validation

Focused lifecycle coverage:

```bash
python -m pytest tests/test_lcm_engine.py::TestSessionRollover -q --tb=short
```

Result: 18 passed.

Default contributing validation:

```bash
pytest -q
python -m compileall -q .
git diff --check
```

Results: targeted suite 290 passed. Full suite 321 passed. Compile, shell syntax, and diff checks passed.

Install-path smoke:

Used `scripts/install.sh` with a temporary `HERMES_HOME` and profile-scoped plugin symlink pointing at this branch. Registered the plugin through its entrypoint, exercised `rollover_session(..., boundary_reason="compression")`, verified depth-zero DAG carry-over, and expanded the moved node successfully.

## Notes

This PR covers the LCM-owned lifecycle compatibility surface. Host CLI and gateway lifecycle notification work remains a separate Hermes Agent integration concern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,7 @@ That means patterns like `cron:*` can catch Hermes cron sessions today, while pl
 
 - **Manual `/new`** should call `rollover_session(old_session_id, new_session_id, ...)`. This finalizes the old physical session, records reset/rollover lifecycle state, binds the new physical session, and applies `LCM_NEW_SESSION_RETAIN_DEPTH` before carrying retained summary nodes forward.
 - **Compression-created session splits** should call `on_session_start(new_session_id, boundary_reason="compression", old_session_id=old_session_id, ...)`. Compression is a continuation of the same logical conversation, so LCM preserves the conversation id and carries all current DAG/message state forward, including depth-0 nodes.
-- For host compatibility, `rollover_session(..., boundary_reason="compression")` is also treated as a compression continuation, not as manual `/new` pruning.
+- For host compatibility, `rollover_session(..., boundary_reason="compression")` is also treated as a compression continuation when `carry_over_context` is enabled, not as manual `/new` pruning. If an integration explicitly passes `carry_over_context=False`, no prior-session messages or DAG nodes are moved into the new session.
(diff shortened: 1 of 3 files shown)


Tosko4:
Good catch, thanks. I pushed bff957d to keep the compression continuation path gated by carry_over_context. There is now a regression test for rollover_session(..., boundary_reason="compression", carry_over_context=False) that verifies no old messages or DAG nodes move into the new session, plus README wording for that flag behavior.

Validation is green locally and on CI.

stephenschoettler (approved):
Approved after local review.

I specifically checked the lifecycle boundary changes around reset, manual new-session rollover, compression-created session splits, and the follow-up fix for `carry_over_context=False`. The new pending reset boundary state is cleared/finalized in the right places, the compression rollover path no longer leaks prior context when carry-over is disabled, and the regressions cover the important compatibility paths.

Validation:
- `tests/test_lcm_engine.py::TestSessionRollover`: 19 passed
- `tests/test_lcm_engine.py`: 147 passed
- `tests/test_lcm_command.py`: 31 passed
- full suite with isolated `HERMES_HOME` and raised `ulimit`: 322 passed
- compileall, shell syntax, and diff whitespace checks passed

No blockers found.

PR #86, opened 2026-04-29
fix: make expand-query synthesis failures visible

Tosko4:
## Summary

`lcm_expand_query` now returns an explicit degraded payload when answer synthesis times out or comes back blank. That avoids two bad outcomes: an exception leaking out of the tool call, or a successful response with an empty answer.

The degraded response keeps the selected `node_ids` and `matches`, so callers can still see what context was found. Unexpected non-timeout synthesis errors still fail hard instead of being hidden as degraded output.

## Validation

• `pytest tests/test_lcm_engine.py::TestEngineTools -q --tb=short`
• `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q`
• `pytest -q`
• `python -m compileall -q .`
• `bash -n scripts/install.sh scripts/update.sh`
• `git diff --check`
• repo-shipped Hermes Agent host-load smoke for timeout and blank synthesis degradation

### DIFF
--- a/tests/test_lcm_engine.py
+++ b/tests/test_lcm_engine.py
@@ -3992,6 +3992,130 @@ class TestEngineTools:
         assert seen["prompt"] == "What was the plan?"
         assert seen["context_blocks"]
 
+    def test_handle_expand_query_timeout_returns_explicit_degraded_error(self, engine, monkeypatch):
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):

PR #87, opened 2026-04-29
fix: clarify current-session retrieval scope

Tosko4:
## Summary

This tightens the retrieval contract around `lcm_grep` and the other LCM tools.

`lcm_grep` is already current-session only, but the README still mentioned `session_scope='all'`. This removes that stale guidance, keeps `session_search` visible as the cross-session recall path, and makes unsupported `session_scope` values show up in the tool result instead of being silently ignored.

## Validation

• `pytest tests/test_lcm_engine.py::TestEngineABC tests/test_lcm_engine.py::TestEngineTools -q --tb=short`
• `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q`
• `pytest -q`
• `python -m compileall -q .`
• `bash -n scripts/install.sh scripts/update.sh`
• `git diff --check`
• repo-shipped Hermes Agent host-load smoke for the current-only schema and ignored unsupported scope

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,10 +334,10 @@ That means patterns like `cron:*` can catch Hermes cron sessions today, while pl
 
 | Tool | Description |
 |------|-------------|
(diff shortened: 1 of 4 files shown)


stephenschoettler (approved):

PR #88, opened 2026-04-29
feat: surface LCM runtime identity

Tosko4:
## Summary

This makes the loaded LCM runtime easier to identify from the normal status surfaces.

The main problem this addresses is install-path drift: when dogfooding or switching between a dev checkout and a live plugin install, it should be obvious which code tree is actually loaded and which database it is using.

This adds runtime identity to `engine.get_status()`, `lcm_status`, and `/lcm status`, including the plugin version, plugin path, module path, Hermes home, database path/source, session id, conversation id, and lifecycle session binding. The unbound `lcm_status` response now also includes runtime identity, so the loaded tree is visible even before a session has been started.

## Validation

* `pytest tests/test_lcm_command.py tests/test_packaging_install.py -q`
* `pytest tests/test_lcm_engine.py::TestEngineABC tests/test_lcm_engine.py::TestEngineTools -q --tb=short`
* `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q`
* `pytest -q`
* `python -m compileall -q .`
* `bash -n scripts/install.sh scripts/update.sh`
* `git diff --check`
* repo-shipped Hermes Agent host-load smoke for bound and unbound runtime identity

### DIFF
--- a/command.py
+++ b/command.py
@@ -57,6 +57,7 @@ def _status_text(engine) -> str:
     db_size = db_path.stat().st_size if db_exists else 0
     session_bound = bool(engine._session_id)
     source_stats = status.get("source_lineage") or {}
+    runtime_identity = status.get("runtime_identity") or {}
     source_stats = {
         "messages_total": int(source_stats.get("messages_total", 0) or 0),
         "attributed_messages": int(source_stats.get("attributed_messages", 0) or 0),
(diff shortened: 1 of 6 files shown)


stephenschoettler (approved):

PR #89, opened 2026-04-29
fix: normalize legacy blank source lineage

Tosko4:
## Summary

This makes legacy blank source rows safe and explicit without rewriting them by default.

The change does four things:

- treats legacy `NULL`, blank, and whitespace-only message sources as `unknown` in raw message filtering
- applies the same `unknown` back-compat semantics to summary/DAG source lineage filtering
- adds `/lcm doctor source` for a read-only normalization scan
- adds `/lcm doctor source apply` for backup-first, idempotent normalization to explicit `unknown`

The apply path only updates legacy blank-source message rows. It does not delete messages, rewrite content, or touch summaries.

## Why

Older databases can contain many persisted rows with an empty source. New writes already normalize missing sources to `unknown`, but old rows still need to be visible and explainable so source filtering does not behave differently depending on whether content is still raw or has been summarized.

This keeps the runtime behavior backward-compatible while giving operators a clear dry-run and a backup-first cleanup path if they want to physically normalize the old rows.

## Validation

```bash
pytest tests/test_lcm_core.py::TestMessageStore::test_source_unknown_filter_matches_null_and_whitespace_legacy_source_rows tests/test_lcm_core.py::TestMessageStore::test_get_source_stats_treats_null_and_whitespace_as_legacy_blank tests/test_lcm_core.py::TestMessageStore::test_source_normalization_plan_and_apply_are_idempotent tests/test_lcm_command.py::test_lcm_doctor_source_reports_dry_run_without_mutating tests/test_lcm_command.py::test_lcm_doctor_source_apply_is_backup_first_and_idempotent tests/test_lcm_engine.py::TestEngineTools::test_handle_grep_unknown_source_filter_matches_whitespace_legacy_summary_lineage tests/test_lcm_engine.py::TestEngineTools::test_handle_doctor_treats_legacy_blank_source_rows_as_healthy -q --tb=short
```

Result: `7 passed`

```bash
pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_lcm_command.py -q
```

Result: `323 passed`

```bash
pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q
```

Result: `295 passed`

```bash
pytest -q
```

Result: `328 passed`

```bash
python -m compileall -q .
bash -n scripts/install.sh scripts/update.sh
git diff --check
```

All passed.

[...]

PR #90, opened 2026-04-29
feat: report lifecycle fragmentation diagnostics

Tosko4:
## Summary

This adds a read-only lifecycle fragmentation diagnostic so operators can see when LCM lifecycle rows no longer line up cleanly with raw message coverage, summary DAG coverage, or the Hermes host session store.

The goal is visibility only. This PR does not repair, delete, rewrite, or normalize lifecycle rows.

## What changed

- Adds `LifecycleStateStore.get_fragmentation_stats(...)`.
- Compares lifecycle `current_session_id` and `last_finalized_session_id` references against:
  - LCM `messages` sessions
  - LCM `summary_nodes` sessions
  - Hermes `state.db.sessions`, when available
- Exposes the diagnostic through `lcm_status`, `lcm_doctor`, and `/lcm doctor`.
- Keeps a missing `state.db` non-warning, which matters for temp/test contexts and fresh host setups.
- Warns when an existing `state.db` is unreadable or schema-incompatible.
- Opens `state.db` using SQLite read-only URI mode.
- Documents the new status and doctor signals.

## Validation

- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_lcm_command.py -q`
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q`
- `python -m pytest -q`
- `python -m compileall -q .`
- `git diff --check`
- Temp DB smoke for `lcm_status`, `lcm_doctor`, and `/lcm doctor`
- Static diff scan for newly added destructive SQL
- Independent review pass, with follow-up fix for malformed existing `state.db` handling

## Notes

The warning should be read as diagnostic evidence, not proof that every mismatch is harmful. Cleanup and repair behavior should stay explicit and separate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,8 +338,8 @@ That means patterns like `cron:*` can catch Hermes cron sessions today, while pl
 | `lcm_describe` | Inspect current-session DAG structure or an `externalized_ref` payload preview without loading full payload content. No node_id/externalized_ref = session overview. |
 | `lcm_expand` | Recover original content from a current-session summary node, or open a stored `externalized_ref` payload directly. |
 | `lcm_expand_query` | Answer a question from expanded LCM context for the active/current session using either a query or explicit node_ids. For cross-session recall, use `session_search` first. |
-| `lcm_status` | Quick health overview — compression count, store size, DAG depth distribution, context usage, and active config. |
-| `lcm_doctor` | Run diagnostics — database integrity, FTS index sync, orphaned nodes, config validation, context pressure. |
+| `lcm_status` | Quick health overview — compression count, store size, DAG depth distribution, context usage, active config, and read-only lifecycle/session fragmentation stats. |
+| `lcm_doctor` | Run diagnostics — database integrity, FTS index sync, orphaned nodes, lifecycle/session fragmentation, config validation, context pressure. |
 
 ### Operator slash commands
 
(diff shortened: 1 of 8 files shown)


stephenschoettler (changes requested):
Requesting changes for one blocker:

- /lcm status docs claim lifecycle fragmentation counters, but command.py::_status_text() does not render lifecycle_fragmentation. lcm_status and /lcm doctor do, so either add the summary to _status_text() or narrow the README wording.

Also rebase after #88 and #89 landed before merge.

Tosko4:
Rebased on current main and narrowed the slash-command README line so it no longer claims /lcm status renders lifecycle fragmentation counters. The JSON lcm_status output still includes the stats, and /lcm doctor remains the text diagnostic surface. CI is green again on the new head.

stephenschoettler (changes requested):
Still requesting changes.

One prior blocker remains after the rebase: `_has_lifecycle_fragmentation()` still ignores `message_sessions_without_lifecycle_current` and `node_sessions_without_lifecycle_reference` in both `tools.py` and `command.py`.

I reproduced the false negative on the rebased head:

- active lifecycle session has normal LCM content
- a second message-only session has no lifecycle current reference
- `message_sessions_without_lifecycle_current=1`
- `_has_lifecycle_fragmentation(stats)` returns `False`

Same for a node-only orphan:

- `node_sessions_without_lifecycle_reference=1`
- `_has_lifecycle_fragmentation(stats)` returns `False`

The README/status wording issue looks narrowed now. The remaining fix is to include those two counters in the warn decision and add regressions for both false-negative cases.

Tosko4:
Good catch. I pushed a follow-up for the two false negatives: both helper predicates now include message-only and node-only sessions missing lifecycle references once lifecycle rows exist, and I added regressions for the JSON doctor path plus the `/lcm doctor` text output. I kept the no-lifecycle legacy case non-warning so blank-source backcompat rows do not become noisy in unbound/test contexts.

CI is green again on the new head.

stephenschoettler (approved):
Looks good now. The latest follow-up warns on message sessions missing any lifecycle reference, while last-finalized rollover history stays healthy. Local targeted validation passed and CI is green.

Issue #78, opened 2026-04-29
No summary nodes after compaction


asheriif:
Hi,

I set this up today to try it and ran into a breaking issue. I attempted to stress test this by dragging a long conversation with an agent with different topics, then after a few compactions, I asked the agent to list brief bullet points of the main topics we discussed, and it missed some main things.

When explicitly told, the agent started calling lcm_grep with vague keywords, and still didn't land upon the missing topics.

After a little bit of debugging, it turned out there have been 8 compactions yet 0 summary nodes, and after a bit more digging, it turned out the nodes exist in the local db, but they're not exposed in the current session. Hermes is rotating the session id, but lcm then refuses the carry-over because it was called by hermes with an old_session_id that doesn't match what it has currently bound.

This can be seen in the logs when it happens after every compaction:

`LCM compression boundary skipped carry-over: old_session_id=<hermes-old> does not match bound session=<lcm-bound>`

Now here's a dump of the LLM (GPT 5.5)'s analysis:

---

# LCM DAG continuity split-brain after compression

## Summary

Hermes/LCM can enter a split-brain state after context compression:

- The prompt contains LCM-injected summaries such as `[Recent Summary (d0, node 7)]`.
- The active LCM tools report `compression_count: 0` and zero current DAG nodes.
- `lcm_expand(node_id=7)` fails with `Node 7 not found in current session`.
- The LCM SQLite database does contain the summary nodes globally, but they are attached to an older/different LCM session ID.

In short: compaction happened, but DAG continuity was not carried into the active session. The summary text survives in the prompt as raw text; the expandable DAG linkage does not.

## User-facing symptom

The model sees prompt text like:

```text
[Recent Summary (d0, node 7)]
...
[Expand for details: ...]
```

But the LCM tools say there is no DAG in the current session:

```json
{
  "session_id": "20260429_002611_0a02c4",
  "compression_count": 0,
  "dag": {
    "total_nodes": 0,
    "total_tokens": 0,
    "depths": {}
  }
}
```

And expanding the visible node ID fails:

```json
{"error": "Node 7 not found in current session"}
```

This is especially confusing because the prompt explicitly instructs the agent to use `lcm_expand`, but the tool refuses the node reference visible in that same prompt.

## Expected behavior

After compression/session rollover:

1. LCM should preserve the logical conversation lineage.
2. Summary nodes referenced in the active prompt should remain attached to the active LCM session, or otherwise be expandable through the active context-engine tools.
[...]

PR #79, opened 2026-04-29
fix: preserve LCM DAG across split compression boundaries

stephenschoettler:
## Summary
- Preserves LCM DAG continuity when a Hermes compression rollover passes an `old_session_id` that differs from the currently bound LCM session.
- Uses the bound LCM session as the carry-over source only when it is the current lifecycle session and owns summary nodes.
- Leaves stale host `old_session_id` data untouched so unrelated sessions are not reassigned.

## Why
- Fixes #78.
- Users could see `[Recent Summary (d0, node N)]` blocks in the prompt while `lcm_status()` and `lcm_describe()` reported zero current DAG nodes.
- `lcm_expand(node_id=N)` then failed because the summary node still belonged to the LCM-bound source session instead of the active post-compression session.

## Validation
- `python -m pytest tests/test_lcm_engine.py::TestSessionRollover::test_compression_boundary_uses_bound_lcm_source_when_host_old_session_differs tests/test_lcm_engine.py::TestSessionRollover::test_compression_boundary_mismatch_resets_session_scoped_state tests/test_lcm_engine.py::TestSessionRollover::test_compression_boundary_continues_logical_session_without_resetting_state tests/test_lcm_engine.py::TestSessionRollover::test_compression_boundary_reassigns_externalized_payload_session_metadata -q` - 4 passed
- `python -m pytest tests/test_lcm_engine.py::TestSessionRollover -q` - 11 passed
- `python -m pytest tests/test_lcm_engine.py -q` - 135 passed
- `HERMES_HOME="$tmpdir/.hermes" python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` - 278 passed
- `HERMES_HOME="$tmpdir/.hermes" python -m pytest -q` - 303 passed
- `python -m compileall -q .`
- `bash -n scripts/install.sh scripts/update.sh`
- `git diff --check`

## Notes
- The existing safety behavior remains: if the mismatch does not point to a valid bound LCM source with current lifecycle state and summary nodes, the boundary is treated as a real new-session reset.
- Existing externalized payload reassignment coverage remains intact.

PR #91, opened 2026-04-29
fix: carry over finalized compression source

stephenschoettler:
## Summary
- Allows compression-boundary recovery when the LCM-bound source was just finalized before the host rollover signal arrives
- Preserves the finalized frontier marker during that recovery path
- Keeps unrelated stale host-session rows and nodes untouched
- Rejects bound-source carry-over when an explicit target conversation id differs from the bound lifecycle conversation

## Why
#79 handled the mismatch case where the LCM-bound source was still current. The reopened #78 report showed a second valid shape: the bound source owns the summary DAG but lifecycle has already finalized it, so the previous guard rejected the carry-over and recreated the split-brain state.

This keeps that safe recovery path, while avoiding cross-conversation carry-over when callers pass an explicit `conversation_id` that does not match the bound lifecycle state.

## Validation
- `python -m pytest tests/test_lcm_engine.py::TestSessionRollover -q`
- `python -m pytest tests/test_lcm_engine.py -q`
- `python -m pytest -q`
- `python -m compileall -q .`
- `git diff --check`

Refs #78

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -709,6 +709,7 @@ class LCMEngine(ContextEngine):
         kwargs: Dict[str, Any],
     ) -> None:
         previous_session_id = self._session_id
+        requested_conversation_id = kwargs.get("conversation_id")
         old_state = self._lifecycle.get_by_session(old_session_id)
         source_session_id = old_session_id
         source_state = old_state
@@ -718,6 +719,10 @@ class LCMEngine(ContextEngine):
             bound_conversation_matches = bool(
                 bound_state
                 and (not self._conversation_id or bound_state.conversation_id == self._conversation_id)
+                and (
+                    not requested_conversation_id
+                    or bound_state.conversation_id == requested_conversation_id
+                )
             )
             bound_is_active_source = bool(
                 bound_state and bound_state.current_session_id == previous_session_id
--- a/engine.py
+++ b/engine.py
@@ -715,13 +715,24 @@ class LCMEngine(ContextEngine):
 
         if previous_session_id and previous_session_id != old_session_id:
             bound_state = self._lifecycle.get_by_session(previous_session_id)
-            bound_is_current_conversation = bool(
+            bound_conversation_matches = bool(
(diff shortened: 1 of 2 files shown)


Tosko4:
I took a closer look at this against the #78 repro shape. This looks like the right fix for the finalized-bound-source case: current main leaves the node on the old LCM source and `lcm_expand` fails, while this branch carries over only the bound LCM source, leaves stale host `old_session_id` data alone, and preserves the finalized frontier.

Two small things before merging, though. Since we want asheriif to confirm the fix in their setup, I’d avoid auto-closing #78 from the PR body and use `Refs #78` instead of `Fixes #78`. Also, there is one narrow safety edge: if an explicit `conversation_id` is ever passed and it differs from the bound lifecycle conversation, the guard can still carry data across conversations. Hermes’ current compression path does not pass that kwarg, so this does not look like the reported runtime path, but a tiny guard/test would make this airtight.

Tosko4 (changes requested):
This does fix the #78 repro shape I tested: current main leaves the node on the old LCM source and `lcm_expand` fails, while this branch carries over only the bound LCM source and preserves the finalized frontier.

I’m still requesting changes for two small blockers. The PR body still says `Fixes #78`, which will auto-close the issue before asheriif can confirm it in their setup. I’d change that to `Refs #78`. Also, the explicit `conversation_id` edge is still open: if a caller passes a different `conversation_id` during the compression boundary, the current guard can move data from the bound lifecycle conversation into the requested conversation. Hermes’ current compression path does not pass that kwarg, so this is narrow, but since the API accepts it, a tiny guard plus regression would make the carry-over rule airtight.

Tosko4 (approved):
Looks good now. I pushed the small guard for explicit `conversation_id` mismatch, added the regression, and changed the PR body to `Refs #78` so the issue stays open for asheriif to confirm.

Local validation and CI are green on the new head.

PR #96, opened 2026-04-30
fix: fail open on session-end sqlite locks

### LINKED ISSUE #95: SQLite locks in on_session_end can block Hermes Discord gateway heartbeat / shard reconnects
## Bug Description

We hit a production issue where `hermes-lcm` caused the Hermes messaging gateway, especially Discord, to become slow/unresponsive and trigger shard reconnects.

Visible symptoms:

- Discord messages were received by the gateway but responses became very slow or absent.
- Discord logs showed heartbeat stalls / shard reconnects:
  ```text
  Shard ID None has stopped responding to the gateway. Closing and restarting.
  ```
- Hermes gateway shutdown/restart paths could stall while LCM was writing/finalizing session state.
- Logs showed SQLite lock errors:
  ```text
  sqlite3.OperationalError: database is locked
  ```
- A separate but related issue also appeared with multimodal/list message content:
  ```text
  sqlite3.ProgrammingError: Error binding parameter 4: type 'list' is not supported
  ```

## Environment

- Hermes Agent gateway running as a systemd service.
- Main gateway handles Discord + Telegram.
- `hermes-lcm` enabled as the context engine.
- LCM database uses SQLite/WAL.
- Discord slash command sync was also producing 429s; disabling slash sync helped reduce startup pressure but did not address the LCM lock issue itself.

## Observed Behavior

When LCM is active, SQLite write contention can occur during message ingestion / compression / session finalization.

In particular, `LCMEngine.on_session_end()` calls `_ingest_messages()` and lifecycle finalization directly. If SQLite is locked there, the exception can propagate or the call can block in a gateway lifecycle path. In a Discord gateway, this can starve the event loop enough to cause heartbeat failures and shard reconnects.

## Expected Behavior

LCM should be best-effort in gateway lifecycle/shutdown paths.

A SQLite lock during `on_session_end()` should not block or crash the messaging gateway. It should log a warning and fail open, preserving gateway availability.

## Local Mitigation Tested

We patched `engine.py` locally so `on_session_end()` catches SQLite lock errors and returns instead of propagating/blocking:

```python
from .db_bootstrap import is_sqlite_locked_error

def on_session_end(self, session_id: str, messages: List[Dict[str, Any]]) -> None:
    try:
        self._ingest_messages(messages)
        self._lifecycle.finalize_session(
            self._conversation_id,
            session_id,
            frontier_store_id=self._last_compacted_store_id,
        )
    except Exception as e:
        if is_sqlite_locked_error(e):
            logger.warning(
                "LCM session-end ingest/finalize skipped due to SQLite lock: %s",
                e,
            )
            return
        raise
```

Regression test added locally:

```python
def test_on_session_end_fails_open_when_ingest_store_is_locked(engine, monkeypatch, caplog):
    engine.on_session_start("test-session", platform="discord")

    def locked_ingest(messages):
        raise sqlite3.OperationalError("database is locked")

    monkeypatch.setattr(engine, "_ingest_messages", locked_ingest)

    with caplog.at_level(logging.WARNING):
        engine.on_session_end("test-session", [{"role": "user", "content": "hello"}])

    assert "LCM session-end ingest/finalize skipped" in caplog.text
```

## Verification

Local test suite after patch:

```text
269 passed in 19.11s
```

## Related Existing Work

I noticed PR #77:

- `fix: harden SQLite message persistence`

That seems related to SQLite hardening and multimodal/list content serialization, but this issue is specifically about gateway availability and fail-open behavior during `on_session_end()` / lifecycle paths.

## Suggested Fix

- Treat SQLite lock failures in gateway lifecycle paths as non-fatal.
- Add bounded write timeouts for gateway use cases.
- Ensure all message content that can be list/dict/multimodal is normalized before SQLite binding.
- Add regression coverage for:
  - `on_session_end()` fail-open on `sqlite3.OperationalError: database is locked`
  - multimodal/list `content` ingestion
  - concurrent gateway session writes if feasible

## Impact

Without this, enabling `hermes-lcm` on a live Discord gateway can degrade the bot’s responsiveness and trigger reconnect loops. Gateway availability should take precedence over perfect LCM persistence in lifecycle/shutdown paths.

stephenschoettler:
## Summary
- Bound `on_session_end()` SQLite writes to a short busy timeout so real writer-lock contention returns quickly instead of waiting through the normal 30 second timeout.
- Split raw-message ingest and lifecycle-finalization lock handling so logs make the persistence tradeoff explicit.
- Added a real held-writer-lock regression that proves `on_session_end()` returns within a small bound.

## Why
- #95 is about gateway availability, not just swallowing the final SQLite exception.
- A normal SQLite locked write waits through `PRAGMA busy_timeout` before raising, so catching the exception after the wait still blocks heartbeat-critical lifecycle paths.
- The final raw-message flush is best effort. If it cannot acquire the DB quickly, availability wins and the warning states that final messages may be absent from the plugin-local store.

## Validation
- `python -m pytest tests/test_lcm_engine.py::TestSessionRollover::test_on_session_end_returns_quickly_under_real_sqlite_writer_lock tests/test_lcm_engine.py::TestSessionRollover::test_on_session_end_fails_open_when_ingest_store_is_locked tests/test_lcm_engine.py::TestSessionRollover::test_on_session_end_fails_open_when_finalize_store_is_locked tests/test_lcm_engine.py::TestSessionRollover::test_on_session_end_reraises_non_lock_errors -q` -> 4 passed
- `python -m pytest tests/test_lcm_engine.py::TestSessionRollover -q` -> 57 passed
- `ulimit -n 4096 && python -m pytest -q` -> 406 passed
- `bash -n scripts/install.sh scripts/update.sh` -> passed
- `git diff --check` -> passed

## Notes
- Full local pytest needed `ulimit -n 4096` on this Arch/Python 3.14 machine to avoid pytest tempdir cleanup hitting the default 1024 file-descriptor ceiling.

Fixes #95

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -117,6 +117,19 @@ def _git_runtime_identity(root: Path) -> dict[str, Any]:
     }
 
 
+def _is_sqlite_locked_error(exc: BaseException) -> bool:
+    """Return True when an exception chain represents SQLite lock contention."""
+    seen: set[int] = set()
+    current: BaseException | None = exc
+    while current is not None and id(current) not in seen:
+        seen.add(id(current))
+        message = str(current).lower()
+        if isinstance(current, sqlite3.Error) and "locked" in message:
+            return True
+        current = current.__cause__ or current.__context__
+    return False
+
+
 _SYNTHETIC_ASSISTANT_NOISE = {
     "ack",
     "acknowledged",
@@ -1241,13 +1254,22 @@ class LCMEngine(ContextEngine):
             if current_thread_session_id == session_id:
                 self._clear_thread_context_stateless(session_id)
             return
-        # Ensure all messages are persisted
-        self._ingest_messages(messages)
-        self._lifecycle.finalize_session(
-            self._conversation_id,
-            session_id,
-            frontier_store_id=self._last_compacted_store_id,
-        )
+        try:
+            # Ensure all messages are persisted
+            self._ingest_messages(messages)
+            self._lifecycle.finalize_session(
+                self._conversation_id,
+                session_id,
+                frontier_store_id=self._last_compacted_store_id,
+            )
+        except Exception as exc:
+            if _is_sqlite_locked_error(exc):
+                logger.warning(
+                    "LCM session-end ingest/finalize skipped due to SQLite lock: %s",
+                    exc,
+                )
+                return
+            raise
 
     def on_session_reset(self) -> None:
         self._pending_reset_session_id = self._session_id
--- a/engine.py
+++ b/engine.py
@@ -13,8 +13,9 @@ import sqlite3
 import subprocess
 import threading
 import time
+from contextlib import contextmanager
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Iterator, List, Optional
 
 from agent.context_engine import ContextEngine
 
@@ -48,6 +49,7 @@ logger = logging.getLogger(__name__)
 
 _PLUGIN_ROOT = Path(__file__).resolve().parent
 _PLUGIN_METADATA: dict[str, str] | None = None
+_SESSION_END_BUSY_TIMEOUT_MS = 50
 
 
 def _strip_metadata_scalar(value: str) -> str:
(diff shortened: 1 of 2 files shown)


Tosko4 (changes requested):
I do not think this fixes #95 yet.

The direction is right: `on_session_end()` should not be allowed to take the gateway down. The problem is that the lock is only caught after SQLite has already waited through the configured busy timeout. I tested this with a real writer lock, not a monkeypatched exception, and `on_session_end()` returned after about 30 seconds:

```text
LCM session-end ingest/finalize skipped due to SQLite lock: database is locked
elapsed_seconds: 30.05
```

That still leaves the Discord gateway thread blocked long enough to hit the heartbeat/shard reconnect problem the issue is about. The current tests pass because they raise `database is locked` immediately from a monkeypatch, so they do not cover the actual production failure mode.

There is also an LCM/DAG contract wrinkle here. Skipping lifecycle finalization is probably acceptable as best-effort debt, but skipping `_ingest_messages()` means the final raw messages may never enter the plugin-local store. Those are different failures and should probably be handled or logged differently. If we choose availability over perfect persistence in this path, the code should be explicit about the data-loss/deferred-flush tradeoff, not make it look like a harmless lifecycle skip.

I would want this to either use a genuinely short/non-blocking session-end write path, move the flush off the gateway-critical path, or add some deferred retry/debt mechanism. And it needs a real-lock regression that asserts `on_session_end()` returns within a small bound under a held writer lock.

stephenschoettler:
Updated #96 with the requested fix.

Changes:
- Session-end SQLite writes now run under a short busy timeout instead of the normal 30 second timeout.
- Real writer-lock contention returns quickly and fails open.
- Raw-message ingest lock handling and lifecycle-finalization lock handling are logged separately.
- The raw-message persistence tradeoff is explicit in the warning.
- Added a held-writer-lock regression that failed before the fix and now passes.

Validation:
- Focused session-end lock tests: 4 passed
- `TestSessionRollover`: 57 passed
- Full suite with `ulimit -n 4096`: 406 passed
- `git diff --check`: passed

Tosko4 (approved):
This addresses the blocker I raised earlier. I rechecked the real writer-lock path and the session-end hook now returns quickly instead of waiting through SQLite's normal busy timeout. The lock handling is scoped to SQLite lock contention, restores the original busy timeouts, and leaves non-lock failures visible.

Validated locally: focused lock tests, the full `TestSessionRollover` class, the full test suite, `git diff --check`, and `compileall`.

PR #99, opened 2026-05-02
fix: route configured custom provider model overrides

zicochaos:
## Summary
- Route LCM model overrides through configured non-canonical Hermes custom providers when the prefix resolves from Hermes config
- Keep canonical/built-in provider names and OpenRouter-style slugs model-only unless they are conservative registry allowlist entries
- Add regression coverage for a non-canonical custom provider prefix such as `my-provider/model-a`
- Add negative coverage for `openai-codex/...` and `custom:openai-codex/...` until the Hermes auxiliary resolver supports those routes end-to-end

## Why
Some Hermes installations may define auxiliary/model routes as named custom providers, for example:

```env
LCM_SUMMARY_MODEL=my-provider/model-a
LCM_EXPANSION_MODEL=my-provider/model-b
```

Before this change, LCM only split allowlisted prefixes like `cerebras/...`, so non-canonical configured custom provider values stayed model-only with no explicit provider.

This keeps the OpenRouter namespace safety concern intact: registry/canonical provider names are not split just because they contain a `/`, and canonical built-ins such as `openai-codex/...` remain model-only. `custom:`-prefixed canonical names are also not treated as safe explicit custom routes here, because the current Hermes auxiliary resolver does not support that path end-to-end.

## Test plan
- `PYTHONPATH="$PWD:$HOME/.sm/hermes-agent" /Users/sbochna/.local/bin/uv run --python 3.11 --with pytest python -m pytest -q tests/test_lcm_core.py::TestModelRouting tests/test_lcm_core.py::TestProviderPrefixedAuxiliaryCalls` → `14 passed`
- `PYTHONPATH="$PWD:$HOME/.sm/hermes-agent" /Users/sbochna/.local/bin/uv run --python 3.11 --with pytest python -m pytest -q` → `411 passed`

### DIFF
--- a/model_routing.py
+++ b/model_routing.py
@@ -15,10 +15,11 @@ class ModelRoute:
 ProviderResolver = Callable[[str], bool]
 
 
-# Conservative allowlist: only split prefixes explicitly supported by LCM
-# overrides. Do not infer from arbitrary provider names, because many Hermes
-# provider IDs are also valid OpenRouter model namespaces, for example
-# ``anthropic/...``, ``deepseek/...``, ``google/...``, and ``x-ai/...``.
+# Conservative allowlist for built-in provider registry fallbacks. Named custom
+# providers from the Hermes config are always safe to split; registry-only
+# providers still need an allowlist because many provider IDs are also valid
+# OpenRouter model namespaces, for example ``anthropic/...``, ``google/...``
+# and ``x-ai/...``.
 _PROVIDER_PREFIXES = frozenset({"cerebras"})
 
 
@@ -35,7 +36,7 @@ def _provider_route_is_resolvable(provider: str) -> bool:
     try:
         from hermes_cli.auth import PROVIDER_REGISTRY
 
-        return provider in PROVIDER_REGISTRY
+        return provider in _PROVIDER_PREFIXES and provider in PROVIDER_REGISTRY
     except Exception:
         return False
 
@@ -47,9 +48,10 @@ def parse_lcm_model_override(
 ) -> ModelRoute:
     """Parse an LCM model override into explicit provider/model routing.
 
-    Values whose first path segment is both allowlisted and resolvable by the
-    Hermes host are split into ``provider=<prefix>`` and ``model=<rest>``.
-    Other values, even when they contain ``/``, remain model-only overrides.
+    Values whose first path segment is resolvable by the Hermes host are split
+    into ``provider=<prefix>`` and ``model=<rest>``. The default resolver only
+    treats named custom providers (plus conservative registry allowlist entries)
+    as resolvable so OpenRouter-style model slugs remain model-only overrides.
     """
     model = (value or "").strip()
     if not model:
@@ -59,7 +61,7 @@ def parse_lcm_model_override(
     provider = provider.strip().lower()
     rest = rest.strip()
     can_resolve_provider = provider_resolver or _provider_route_is_resolvable
-    if sep and provider in _PROVIDER_PREFIXES and rest and can_resolve_provider(provider):
+    if sep and rest and can_resolve_provider(provider):
         return ModelRoute(provider=provider, model=rest)
 
     return ModelRoute(provider=None, model=model)
--- a/model_routing.py
+++ b/model_routing.py
@@ -15,16 +15,28 @@ class ModelRoute:
 ProviderResolver = Callable[[str], bool]
 
 
-# Conservative allowlist for built-in provider registry fallbacks. Named custom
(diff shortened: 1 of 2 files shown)


Tosko4 (changes requested):
I think the general direction is right, but the current example and test are proving a path that the real Hermes auxiliary resolver does not actually take.

I checked this against the current Hermes host code, not just the parser. A non built-in named custom provider works as expected:

```text
my-provider/model-a => provider=my-provider, model=model-a
```

But `openai-codex/gpt-5.4-mini` does not become a named custom provider route in Hermes. `openai-codex` is a canonical built-in provider, and Hermes intentionally prevents bare custom providers from shadowing canonical built-ins. I also checked `custom:openai-codex/gpt-5.4-mini`: the LCM parser can split it, but `agent.auxiliary_client.resolve_provider_client()` normalizes `custom:openai-codex` back to `openai-codex` before the named-custom branch, so it still tries the built-in Codex path rather than the custom provider entry.

So I would keep the feature, but narrow the claim and tests to the behavior that is actually supported today:

1. Use a non built-in custom provider example, for example `my-provider/model-a`.
2. Add a config-backed regression for a real `providers.my-provider` or `custom_providers` entry, not only a lambda resolver.
3. Add a negative regression that `openai-codex/gpt-5.4-mini` stays model-only when `openai-codex` is a canonical built-in provider.
4. If support for `custom:openai-codex/model` is desired, I think that needs a matching Hermes Agent auxiliary-router change first. I would not document that form from this plugin PR until the host actually routes it correctly.

With that adjustment, this looks generally useful rather than environment-specific. As written, the code shape is close, but the PR text and the `openai-codex` regression make it look like it fixes a route that still does not work through the real host resolver.

stephenschoettler (changes requested):
Confirming Tosko4's existing blocker with local validation.

The non built-in custom-provider direction is useful, but this PR should stay scoped to non-canonical custom provider names. Against current Hermes host code, my temp config check produced:

```text
my-provider/model-a => provider=my-provider, model=model-a
openai-codex/gpt-5.4-mini => provider=None, model=openai-codex/gpt-5.4-mini
custom:openai-codex/gpt-5.4-mini => provider=custom:openai-codex, model=gpt-5.4-mini
```

That last parse still routes through built-in Codex in `agent.auxiliary_client.resolve_provider_client()`, not the named custom provider. So the requested change is the same as Tosko4's: use a non-canonical provider example, add config-backed positive coverage for that path, and add negative coverage for `openai-codex/...` plus `custom:openai-codex/...` until Hermes Agent supports that host route.

Local validation: focused model-routing tests pass, full suite passes with 408 passed, compileall passes.

zicochaos:
Thanks — you were right. I narrowed the PR to non-canonical named custom providers only.

Updated in b49cc24:
- replaced the positive openai-codex example with my-provider/model-a
- added config-backed positive coverage for a non-canonical custom provider
- added negative coverage for openai-codex/... staying model-only even if a custom config entry exists
- added negative coverage for custom:openai-codex/... staying model-only until the Hermes auxiliary resolver supports that route end-to-end
- updated routing comments/PR body to avoid claiming support for canonical built-in shadowing

Validation:
- focused routing/aux tests: 14 passed
- full suite: 411 passed

Tosko4 (approved):
This addresses my blocker. I rechecked the updated head against the real Hermes host path: non-canonical custom providers split correctly, while `openai-codex/...` and `custom:openai-codex/...` stay model-only for now.

Local validation on `b49cc24` passed for the focused routing/aux tests, the full suite, compileall, and diff check. CI still needs the usual fork workflow approval, but the routing shape looks right to me now.

Tosko4:
@stephenschoettler @zicochaos I cleared my change request with the approval on b49cc24. The original blocker looks addressed to me now.

Only thing left from my side: GitHub still shows the PR as blocked because Stephen’s earlier change request is still active, so that may get in the way of a squash merge until it’s cleared.

stephenschoettler (approved):
Verified the updated head b49cc24 locally. The previous blocker is resolved: the PR is now scoped to non-canonical named custom providers, keeps canonical built-ins model-only, and keeps custom: canonical routes model-only until Hermes Agent supports that path end to end.

Validation:
- PYTHONPATH="$PWD:/home/w0lf/hermes-agent" python -m pytest -q tests/test_lcm_core.py::TestModelRouting tests/test_lcm_core.py::TestProviderPrefixedAuxiliaryCalls, 14 passed
- PYTHONPATH="$PWD:/home/w0lf/hermes-agent" python -m pytest -q, 411 passed
- python -m compileall -q ., passed

Issue #101, opened 2026-05-04
Feature: add bounded cross-session search for plugin-local LCM history


GodsBoy:
## Summary

`hermes-lcm` is a current-session tool today. The plugin-local LCM SQLite database (`lcm.db`) accumulates messages across many sessions, including rows that may have been externally backfilled from other sources, but the `lcm_grep` tool surface cannot reach beyond the active session. This issue proposes adding bounded archive search to `lcm_grep` over rows already present in `lcm.db`, with `current` remaining the safe default.

> Note: this issue was originally framed as "imported OpenClaw / lossless-claw history". Per maintainer guidance on PR #102, the reframing is "bounded search over rows already present in `lcm.db`, including externally backfilled rows that may carry source strings such as `openclaw-lcm:*`". The feature stands on rows-already-present, not on an implied or supported import path.

## Current limitation

In `tools.py`, `lcm_grep` reads `requested_session_scope` from args, then hard-codes the search to the current session and logs a warning for any other value:

```python
session_scope = "current"
...
if requested_session_scope != "current":
    logger.warning("Ignoring unsupported session_scope=%s for lcm_grep", requested_session_scope)
```

The schema's `session_scope` enum is locked to `["current"]`. Underneath, both `MessageStore.search` and `SummaryDAG.search` are now bound on `session_id=None` as the explicit all-sessions sentinel (#103, #104), so the data layer is ready. The gap is at the tool layer.

## Why `session_search` is not sufficient

The README correctly directs cross-session queries to Hermes' built-in `session_search`. That tool covers Hermes-tracked session history, but it does not reach plugin-local LCM rows that are only present in `lcm.db`.

Operators may have rows in `~/.hermes/lcm.db` that carry source strings such as:

- `openclaw-lcm:agent:sammy:telegram:direct:503782402:conversation:88`
- `openclaw-lcm:agent:nehemiah:telegram:group:-1003814898545:topic:5:conversation:5`
- `openclaw-lcm:agent:david:telegram:group:-1003814898545:topic:9:conversation:6`

These rows exist in the local LCM database but are not reachable via `lcm_grep` today. The issue is a search-tool gap, not a claim about how those rows got there.

## Proposed scope

Extend `lcm_grep` with:

- `session_scope` values:
  - `current` (default): unchanged behavior
  - `all`: search every session in the local LCM database
  - `session`: search a single session id supplied via `session_id`
- Required filter when `session_scope='session'`:
[...]

PR #102, opened 2026-05-04
feat: add bounded cross-session search to lcm_grep

GodsBoy:
## Summary
- `lcm_grep` gains explicit opt-in `session_scope='all'` and `session_scope='session'` (with `session_id`) for bounded archive recovery over rows already present in `lcm.db`. Default `session_scope='current'` is preserved byte-for-byte.
- Cross-session scopes return raw-message hits only. Cross-session DAG/summary expansion is intentionally deferred.
- `lcm_expand` gains a `store_id` mode that fetches a single raw message regardless of session, paged via `content_offset`. `node_id` and `externalized_ref` modes remain current-session.
- `limit` is clamped to 200 with `limit_clamped_from` echoed when the input exceeds the cap.
- Each result carries `from_current_session` so callers can prefer in-session expansion paths when both are available.

## Why
- Closes the gap that `lcm_grep` is unreachable for rows already present in the plugin-local LCM database, including externally backfilled rows that may carry source strings such as `openclaw-lcm:*`.
- The narrowing positions this as **bounded archive recovery**, not a memory-system expansion. Tool descriptions continue to direct normal cross-session recall to `session_search`.
- No claim of an official OpenClaw/lossless-claw importer; the feature stands on rows-already-present.
- Refs #101.

## Validation
- `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> **402 passed**
- `pytest -q` -> **439 passed**
- `python3 -m compileall -q .` -> clean
- `bash -n scripts/install.sh scripts/update.sh` -> clean
- `git diff --check origin/main` -> clean

## Notes
- **Rebased** onto current `main` (past #104 and #106). Branch history was rewritten and force-pushed accordingly. GitHub Actions check runs on a fork PR from a non-collaborator require maintainer approval to run; the empty checks on the head are a permissions thing, not a CI failure.
- **Defers to follow-up issue #108**: pushing `role`/`time_from`/`time_to` filters into the search layer. The maintainer's Bug 2 reproduction is captured as the regression contract on that issue.
- **Cross-session DAG/summary expansion** stays deferred. Sessions remain bound on source-id provenance; returning non-expandable summary hits would push this tool toward a broader memory-system shape than the maintainer's scope direction allows.
- **`lcm_expand` store_id mode** is the only cross-session drilldown path in this PR.
- Empty-engine-session safety relies on the data-layer fix from #104 (empty `session_id` is now a literal scoped filter); the unknown-scope fallback routes through current-session and naturally returns zero results rather than leaking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,19 +248,20 @@ for earlier separate sessions or broad cross-session history.
 
 | Tool | Use |
 |------|-----|
-| `lcm_grep` | Search current-session raw messages and summaries. Use `session_search` for earlier separate sessions or broad cross-session recall. |
+| `lcm_grep` | Search current-session raw messages and summaries. Opt into `session_scope='all'` or `session_scope='session'` (with `session_id`) for bounded archive recovery over rows already present in `lcm.db`, including externally backfilled rows that may carry source strings such as `openclaw-lcm:*`; broader scopes return raw-message hits only. Use `session_search` for earlier separate sessions or broad cross-session recall. |
 | `lcm_describe` | Inspect the current-session DAG or preview an `externalized_ref` without loading full content. |
-| `lcm_expand` | Recover source messages, child summaries, or externalized payloads with pagination. |
+| `lcm_expand` | Recover source messages, child summaries, or externalized payloads with pagination. Use `store_id` to fetch a single raw message regardless of session, suitable for drilling into a cross-session `lcm_grep` result. |
 | `lcm_expand_query` | Answer a question using expanded current-session LCM context while returning a bounded answer. |
 | `lcm_status` | Show runtime health, context pressure, config, source lineage, and lifecycle stats. |
 | `lcm_doctor` | Run database, FTS, lifecycle, config, and context-pressure diagnostics. |
 
 ### Retrieval contract
 
-LCM retrieval tools are current-session tools. `session_scope` is kept for
-schema compatibility and currently accepts only `current`; unsupported values
-are ignored and reported in the tool result. Use Hermes `session_search` for
-earlier sessions.
+LCM retrieval tools default to current-session scope. `lcm_grep` accepts
+`session_scope='all'` or `session_scope='session'` as an explicit opt-in for
+bounded archive search over rows already present in `lcm.db` (raw-message hits
+only); use Hermes `session_search` for broad cross-session history outside the
+LCM database.
 
 Within the current session, `source` filters raw rows directly and filters
 summary nodes by descendant raw-message source lineage. `unknown` is a real
--- a/schemas.py
+++ b/schemas.py
@@ -3,11 +3,13 @@
 LCM_GREP = {
     "name": "lcm_grep",
     "description": (
-        "Search the current session's conversation history — raw messages AND summaries across all depths — "
-        "to recover details from earlier in the active conversation, even if those turns were compacted. "
-        "Prefer this for intra-session recall. If the user is asking about an earlier separate conversation "
-        "or broad cross-session history, prefer session_search instead. Returns matches with depth labels "
-        "showing where each result lives."
+        "Search the LCM database for past conversation content (raw messages AND summaries across all depths) "
+        "to recover details from the active session or, when explicitly scoped, from other sessions in the "
+        "plugin-local LCM database (including history imported from OpenClaw or lossless-claw). "
+        "Default scope is current-session only; broader scopes must be requested explicitly. "
+        "Cross-session summary hits are returned as snippets but cannot be expanded by node_id in this version "
+        "(they carry cross_session_expand_supported=false); use lcm_expand with the result's store_id for raw-message "
+        "expansion across sessions. For Hermes-tracked session history outside the LCM database, use session_search."
     ),
     "parameters": {
         "type": "object",
@@ -22,7 +24,10 @@ LCM_GREP = {
             },
             "limit": {
                 "type": "integer",
-                "description": "Max results to return (default 10)",
+                "description": (
+                    "Max results to return (default 10, hard upper bound 200). "
+                    "Values above the cap are clamped and reported via limit_clamped_from in the response."
+                ),
                 "default": 10,
             },
             "sort": {
@@ -36,12 +41,23 @@ LCM_GREP = {
             },
             "session_scope": {
                 "type": "string",
-                "enum": ["current"],
+                "enum": ["current", "all", "session"],
                 "description": (
-                    "Current-session only. Cross-session recall should use session_search instead; "
-                    "unsupported values are ignored and reported in the tool result."
+                    "Scope of the search across the plugin-local LCM database. "
+                    "'current' (default) restricts to the active session and preserves historical behavior. "
+                    "'all' searches every session in the local LCM database. "
+                    "'session' restricts to the session_id supplied via the session_id parameter. "
+                    "Cross-session search returns snippets and message store_ids; cross-session summary node expansion is deferred. "
+                    "For Hermes-tracked session history outside the LCM database, use session_search."
                 ),
                 "default": "current",
+            },
+            "session_id": {
+                "type": "string",
+                "description": (
+                    "When session_scope='session', the explicit session id to restrict the search to. "
+                    "Must not be supplied with session_scope='current' or session_scope='all'."
+                ),
             },
             "source": {
                 "type": "string",
@@ -49,6 +65,27 @@ LCM_GREP = {
                     "Optional source/platform filter (for example cli, discord, telegram). "
                     "Applies directly to raw messages and to summaries via descendant source lineage. "
                     "Use 'unknown' for explicit unknown-source content."
+                ),
+            },
+            "role": {
+                "type": "string",
+                "enum": ["user", "assistant", "tool"],
+                "description": (
+                    "Optional role filter. Applies to message hits only; summary hits are returned unfiltered "
+                    "(role does not exist on summary nodes). The response echoes role_filter_applies='messages_only' when this is set."
+                ),
+            },
+            "time_from": {
+                "type": "string",
+                "description": (
+                    "Optional inclusive ISO 8601 lower bound for message timestamps and summary latest_at. "
+                    "Examples: '2026-01-01T00:00:00Z' or '2026-01-01T00:00:00+00:00'."
+                ),
+            },
+            "time_to": {
+                "type": "string",
+                "description": (
+                    "Optional inclusive ISO 8601 upper bound for message timestamps and summary latest_at."
                 ),
             },
         },
--- a/schemas.py
+++ b/schemas.py
@@ -124,26 +124,37 @@ LCM_DESCRIBE = {
 LCM_EXPAND = {
     "name": "lcm_expand",
     "description": (
-        "Recover the original detail behind a current session summary node, or open an "
-        "externalized payload ref directly. Given a node_id, returns the "
-        "source messages or lower-depth summaries that were compacted into "
-        "that node. Output is bounded by default, but raw recovery is pageable: "
-        "use source_offset/source_limit to page immediate sources and content_offset "
-        "to continue an oversized message or externalized payload. Given externalized_ref, "
-        "returns the stored payload content plus cursor metadata. Use after lcm_describe "
-        "to drill into specific parts of the active conversation or large externalized "
-        "tool output. For cross-session recall, prefer session_search first."
+        "Recover the original detail behind a summary node, externalized payload, or raw message. "
+        "Mode selection (exactly one): node_id (current session only) returns the source messages "
+        "or lower-depth summaries that were compacted into a summary node; externalized_ref "
+        "(current session only) returns a stored externalized payload's content; store_id returns "
+        "a single raw message by store_id and works across sessions, suitable for drilling into "
+        "cross-session lcm_grep results. Output is bounded by max_tokens; raw recovery is pageable "
+        "via content_offset (and source_offset/source_limit for node_id mode). For Hermes-tracked "
+        "session history outside the LCM database, prefer session_search."
     ),
     "parameters": {
         "type": "object",
         "properties": {
             "node_id": {
                 "type": "integer",
-                "description": "Summary node ID to expand",
+                "description": (
+                    "Summary node ID to expand. Current-session only — cross-session DAG expansion "
+                    "is not supported in this version."
+                ),
             },
             "externalized_ref": {
                 "type": "string",
-                "description": "Optional externalized payload ref filename to expand instead of a summary node.",
+                "description": "Externalized payload ref filename to expand instead of a summary node. Current-session only.",
+            },
+            "store_id": {
+                "type": "integer",
+                "description": (
+                    "Raw message store_id to fetch. Works across sessions, so a store_id surfaced by "
+                    "a cross-session lcm_grep result can be expanded directly. Returns the message's "
+                    "content paged by content_offset; if the row references an externalized payload, "
+                    "the ref is surfaced but content is not hydrated."
+                ),
             },
             "max_tokens": {
                 "type": "integer",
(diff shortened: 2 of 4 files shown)


GodsBoy:
## Code-review pass: applied fixes pushed (commit \`19d05f8\`)

A multi-agent review surfaced 17 findings across correctness, contract, and idiom. Most are now addressed in the new commit; the remainder are advisory or filed separately.

### Why \`session_scope='all'\` is being re-introduced after #52

The reviewer noted that #52 (\`fix: restrict lcm_grep to active session only\`) explicitly removed this scope. This PR re-introduces it with the constraints that were absent then:

- **Hard 200-result clamp** with \`limit_clamped_from\` echoed when input exceeds the cap, so a single call cannot pollute the agent's context window with cross-session results
- **Compact snippets only**, not raw content
- **\`from_current_session\` flag** on every result so callers can prefer in-session expansion paths when both are available
- **Default stays \`current\`** byte-for-byte, plus a new defensive guard that refuses \`scope=current\` when the engine has no session bound (so the data-layer truthiness leak filed as #103 cannot reach the tool surface here)
- **Cross-session DAG expansion is intentionally deferred**; cross-session summary hits carry \`cross_session_expand_supported: false\` so callers cannot misinterpret them as drillable

If any of these mitigations are insufficient or if the original concern from #52 was different, happy to revisit.

### Applied (commit \`19d05f8\`)

- **Reject \`limit <= 0\`** explicitly instead of silent coercion to 1
- **Reject naive ISO 8601 timestamps** (no \`Z\` or offset) so cross-machine reproducibility holds
- **Reject inverted \`time_from > time_to\`** with explicit error
- **Defensive empty-session guard** in \`scope=current\` (filed underlying data-layer leak as #103)
- **Catch \`OverflowError\`** in \`store_id\` integer parse (covers \`float('inf')\`)
- **Echo \`time_from\`/\`time_to\`** as caller's input string (ISO 8601), not as parsed Unix float — schema declares string
- **\`from_current_session\` honest** when no engine session is bound (was \`True\` for foreign rows when both ids were empty)
- **Cross-session externalized_ref** surfaces \`externalized_note\` explaining payload metadata is session-scoped (foreign refs are traceability-only)
- **Hygiene:** module-level \`datetime\` import; \`frozenset\` for valid-scope/valid-role constants; module-level test imports

Validation: \`pytest -q\` -> **443 passed** (added 6 regression tests). \`compileall\`, \`bash -n\`, \`git diff --check\` clean.

### Filed separately

- **#103** — pre-existing data-layer leak: \`MessageStore.search\` and \`SummaryDAG.search\` use truthiness check on \`session_id\`, so empty string falls through to all-session search. Found during this review but not introduced here. Safe to land independently.

### Advisory / surfaced for maintainer judgment (not applied)

- Cross-session source-filtered DAG search runs per-row recursive CTEs (up to 5000) — pre-existing pattern, amplified by removing the per-session bound. Worth measuring on a production-size DB before deciding whether to add a bulk-prefetch variant or document the known-slow combination.
- Asymmetric mode-detection in \`lcm_expand\` (empty \`externalized_ref\` truthy-treated, others use \`is not None\`) — design call; left as-is since changing it would alter behavior for callers passing \`externalized_ref=\"\"\`.
- Validation section above lists \`git diff --check origin/main\` rather than CONTRIBUTING.md's plain \`git diff --check\` — same effect on a clean tree, scoped to the PR delta intentionally.

asheriif:
Hi, not to undermine your work here at all, but just wanted to share my opinion:

- This blurs the line between a compression engine and a memory system.
- Hermes has `session_search` already built in, and it does exactly this: search across sessions.
- This is a big change that adds maintenance overhead.

I personally would rather LCM remain focused as a compression engine, but this is of course up to the maintainers.

stephenschoettler (changes requested):
Requesting changes. The core shape is useful and the existing suite is green, but I found two correctness/security-boundary gaps that need tests before merge.

1. `tools.py:657-660`: unsupported `session_scope` falls back to `current` without the empty-session guard.

The recognized `current` path now refuses to search when `engine._session_id` is empty, but the unknown-scope fallback sets `search_session_id = engine._session_id` and continues. On this branch, with a fresh unbound engine (`_session_id == ""`), this returns rows from multiple sessions while the response reports `session_scope: "current"` and `ignored_session_scope: "bogus"`.

Repro from the PR head:

```python
engine = LCMEngine(config=LCMConfig(database_path=tempfile.mktemp(suffix='.db')))
assert engine._session_id == ""
engine._store.append("session-a", {"role": "user", "content": "docker from a"})
engine._store.append("session-b", {"role": "user", "content": "docker from b"})
json.loads(engine.handle_tool_call("lcm_grep", {"query": "docker", "session_scope": "bogus", "limit": 10}))
```

Observed: `total_results == 2`, with hits from both `session-a` and `session-b`.

Fix: route unsupported scopes through the same guarded current-session path, or reject unsupported scopes before searching. Add a regression for unbound engine plus invalid `session_scope`.

2. `tools.py:670-746`: `role`, `time_from`, and `time_to` filters are applied after bounded search candidates are fetched.

`lcm_grep` fetches only `source_limit = max(limit * 4, limit, 20)` from the store/DAG, then applies role/time filters in Python. That means valid filtered matches outside the first candidate window disappear.

Repro from the PR head:

```python
# one older matching assistant row
assistant_id = engine._store.append("test-session", {"role": "assistant", "content": "docker target assistant"})
engine._store._conn.execute("UPDATE messages SET timestamp=? WHERE store_id=?", (1000.0, assistant_id))

# 25 newer matching user rows fill the pre-filter candidate window
for i in range(25):
    sid = engine._store.append("test-session", {"role": "user", "content": f"docker noisy user {i}"})
    engine._store._conn.execute("UPDATE messages SET timestamp=? WHERE store_id=?", (2000.0 + i, sid))
engine._store._conn.commit()

json.loads(engine.handle_tool_call("lcm_grep", {"query": "docker", "role": "assistant", "limit": 1, "sort": "recency"}))
```

Observed: `total_results == 0`. Expected: the assistant row is returned. The same issue reproduces with `time_to` when newer excluded rows fill the first candidate window and an older valid row exists beyond it.

Fix: push role/time filtering into the data-layer queries, or keep paging candidates until enough post-filtered matches are collected, bounded by a hard candidate cap. Add tests where the only matching role/time row sits outside the initial top 20 candidates.

Validation I ran locally:

- `ulimit -n 8192; python -m pytest tests/test_lcm_engine.py::TestHandleGrepCrossSession tests/test_lcm_engine.py::TestHandleExpandStoreId -q -o addopts=`: 31 passed.
- `ulimit -n 8192; python -m pytest -q -o addopts=`: 443 passed.
- `python -m compileall -q .`: passed.
- `bash -n scripts/install.sh scripts/update.sh`: passed.
- Added-line security scan: no findings.

The green suite does not cover the two gaps above.

stephenschoettler:
## Maintainer scope direction

asheriif's concern is valid: LCM should not become the general cross-session memory system. Hermes already has `session_search` for broad Hermes-tracked history, and the default tool guidance should continue to send normal prior-session recall there.

I do still think #101 identifies a real gap: plugin-local LCM/imported rows, especially OpenClaw or lossless-claw history stored only in `lcm.db`, can be a data island that `session_search` does not cover. That use case is worth supporting, but the scope needs to stay narrow.

For this PR, please frame and implement the feature as plugin-local archive/import recovery, not general memory:

- Keep `session_scope=current` as the default and primary path.
- Keep docs/tool descriptions steering normal cross-session Hermes recall to `session_search`.
- Allow `session_scope=all` / `session_scope=session` only as an explicit opt-in for plugin-local LCM rows, including imported OpenClaw/lossless-claw rows not indexed elsewhere.
- Prefer returning bounded raw-message hits with `store_id`, compact snippets, `from_current_session`, and hard limit clamping.
- Keep `lcm_expand store_id` as the only cross-session drilldown path in this PR.
- Remove or defer cross-session summary/DAG hits unless this PR also provides a real expansion contract. Returning non-expandable summary hits makes this look too much like a broader memory layer.
- Remove or defer `role`, `time_from`, and `time_to` filters unless they are applied in the search layer or otherwise proven not to lose valid results after candidate bounding.
- Drop unrelated `.gitignore` churn from this PR if practical, since the scope is already sensitive.

So the direction is: acceptable if narrowed to bounded plugin-local archive/import search, not acceptable as a broad memory-system expansion.

stephenschoettler:
One clarification after re-reading #101/#102: we should not imply that hermes-lcm or Hermes Agent has an official OpenClaw/lossless-claw history importer unless one is linked.

The valid use case here is narrower: local LCM rows may already exist across sessions, and some operators may have externally backfilled rows from OpenClaw/lossless-claw databases with source strings like `openclaw-lcm:*`.

Please update the rationale/docs to say "externally backfilled/plugin-local rows" unless this PR can link the importer that creates those rows. The feature should stand on: "bounded search over rows already present in `lcm.db`", not on an implied supported import path.

GodsBoy:
## Narrowing per maintainer review (`bcd7c2a`)

Force-pushed the branch with a rewritten history and a narrowing commit that addresses the `CHANGES_REQUESTED` review and the two follow-up scope-direction comments. Reviewers will see a smaller, more focused diff; the prior commits are still reachable via the review thread anchored to commit SHAs.

### History rewrite

- Dropped the `chore: gitignore local agent-tooling artifacts` precursor commit per the "drop unrelated `.gitignore` churn" direction. Local hygiene moved to a per-clone `.git/info/exclude`.

### Bug fixes (Stephen's review)

- **Unknown `session_scope` fallback now inherits the empty-engine-session guard.** Previously an unbound engine plus an invalid scope returned multi-session rows while reporting `session_scope: current`. Added a regression test that mirrors the exact reproduction from the review.
- **`role`, `time_from`, `time_to` filters removed.** Pushing them into the search layer is the right shape but a meaningfully larger change; deferred to a follow-up issue with the maintainer's Bug 2 reproduction captured as the test contract.

### Scope narrowing (maintainer scope-direction comments)

- **Cross-session summary/DAG hits removed.** Broader scopes (`session_scope='all'` / `'session'`) return raw-message hits only. Dropped the `cross_session_expand_supported` marker since there are no longer non-expandable hits to flag. Cross-session DAG expansion stays intentionally deferred.
- **Reframed away from implied importer.** README, schema descriptions, and tests now position the cross-session opt-in as bounded archive recovery over rows already present in `lcm.db` (including externally backfilled rows that may carry source strings such as `openclaw-lcm:*`) rather than as "imported OpenClaw or lossless-claw history". No claim of an official importer.
- **Default and primary path preserved.** `session_scope='current'` remains the default. Tool descriptions continue to direct normal cross-session recall to `session_search`.
- **`lcm_expand store_id` retained as the sole cross-session drilldown** for raw-message hits.

### Validation

- `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> **398 passed**
- `pytest -q` -> **435 passed**
- `python3 -m compileall -q .` -> clean
- `bash -n scripts/install.sh scripts/update.sh` -> clean
- `git diff --check 0f60d720` -> clean

### Commits on the branch (after rewrite)

- \`a8dbcef\` feat(grep): add bounded cross-session search to lcm_grep
- \`a827b73\` feat(expand): add cross-session raw expansion by store_id
- \`01a15fb\` docs: document lcm_grep cross-session scopes and lcm_expand store_id
- \`ab2529d\` test: cover cross-session grep and store_id expand
- \`41de03e\` fix: tighten lcm_grep/lcm_expand validation per review (#102)
- \`bcd7c2a\` fix: narrow PR per maintainer scope direction (refs #102)

A follow-up issue has been filed for the role/time/source filters with a search-layer push, capturing Stephen's Bug 2 reproduction as the test contract.

stephenschoettler:
Thanks for narrowing this. The branch direction now looks much closer to the maintainer scope: bounded plugin-local archive recovery, current-session default, raw-message broader scopes only, `store_id` as the cross-session drilldown path, and role/time filters deferred to #108.

Still blocked before re-review:
- GitHub reports `mergeStateStatus=DIRTY` / conflicting.
- There are no check runs on the current head.
- The PR body is stale. It still mentions `role`, `time_from`, `time_to`, cross-session summary markers, `.gitignore` churn, and imported OpenClaw/lossless history.

Please rebase/update on current `main`, refresh the PR body to match the narrowed `bcd7c2a` diff, rerun the documented validation, and request review again. Keep #102 standing on bounded search over rows already present in `lcm.db`; importer/provenance work is separate in #105/#107.

GodsBoy:
## Rebased and ready for re-review (`c837e9b`)

Addressed the three blockers from the previous comment:

### Rebased onto current `main`

Branch history was rewritten on top of `6b0269b` (past #104 and #106). Force-pushed accordingly. Resolution choices:

- **README**: took the pruned shape from #106 wholesale, then added the bounded-archive opt-in language to the `lcm_grep` and `lcm_expand` rows of the tool table and a one-paragraph retrieval-contract note. No re-introduction of the long retrieval-contract section #106 removed.
- **Empty-engine-session safety**: dropped our defensive tool-layer error in favor of the data-layer fix from #104 (empty `session_id` is now a literal scoped filter, so an unbound engine on `scope=current` naturally returns zero results). The unknown-scope fallback routes through current-session and inherits the same protection. Stephen's `test_handle_grep_unbound_current_session_does_not_search_all_sessions` now passes against this branch.
- **Tests**: kept our overlapping `test_empty_engine_session_with_unknown_scope_does_not_leak` but updated it to assert zero-results behavior matching the data-layer contract instead of the now-removed tool-layer error.

### PR body refreshed

The body now describes the narrowed `c837e9b` shape with no mentions of `role`/`time_from`/`time_to`, no `cross_session_expand_supported`, no `.gitignore` churn, and no "imported OpenClaw/lossless-claw history" framing. References #108 (the deferred filter follow-up) and the deferred cross-session DAG expansion explicitly.

### Validation

- `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> **402 passed**
- `pytest -q` -> **439 passed**
- `python3 -m compileall -q .` -> clean
- `bash -n scripts/install.sh scripts/update.sh` -> clean
- `git diff --check origin/main` -> clean

### Note on CI check runs

GitHub Actions on a fork PR from a non-collaborator requires a maintainer to approve the workflow run for it to fire. The empty `check-runs` count on the head is the default permissions behavior, not a CI failure. Approving once on the next push should re-enable auto-runs on subsequent updates.

stephenschoettler (approved):
Approved after re-review of `c837e9b`.

The previous blockers are addressed:
- branch is rebased on current `main`; conflicts are gone
- PR body now matches the narrowed scope
- role/time filters are deferred to #108 and are absent from the schema
- no official OpenClaw/lossless-claw importer claim remains
- invalid `session_scope` fallback plus empty-engine session behavior is covered by the data-layer fix and regressions
- broader scopes return raw-message hits only, with `store_id` as the drilldown path

Local validation on this head:
- `python -m pytest tests/test_lcm_engine.py::TestHandleGrepCrossSession tests/test_lcm_engine.py::TestHandleExpandStoreId -q` -> 22 passed
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> 402 passed
- `python -m pytest -q` -> 439 passed
- `python -m compileall -q .` -> clean
- `bash -n scripts/install.sh scripts/update.sh` -> clean
- `git diff --check origin/main...HEAD` -> clean

I also found a carried-over summary expansion issue that reproduces on latest `main` and on this branch, tracked separately in #110. Not blocking this PR because it predates this branch.

Tosko4 (approved):
Reviewed `c837e9b`. This matches the narrowed scope now: `current` remains the default path, broader `lcm_grep` scopes are explicit and raw-message only, source filtering stays in the search layer, and `store_id` is the one cross-session drilldown path.

I rechecked the blocker shapes from the earlier review. Invalid scope on an unbound engine now stays scoped to current and returns zero results, role/time filters are gone from this PR, current-session summary search still works, broader scopes do not return summary hits, and cross-session externalized refs are surfaced as traceability only rather than expanded.

Validation I ran locally on this head:

`python -m pytest tests/test_lcm_engine.py::TestHandleGrepCrossSession tests/test_lcm_engine.py::TestHandleExpandStoreId -q -o addopts=`: 22 passed

`python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=`: 402 passed

`python -m pytest -q -o addopts=`: 439 passed

`python -m compileall -q .`, `bash -n scripts/install.sh scripts/update.sh`, and `git diff --check origin/main...HEAD` are clean.

Approving from my side. GitHub Actions still shows `action_required` on the fork head, so hosted CI needs maintainer workflow approval if you want that signal on this exact SHA.

PR #104, opened 2026-05-04
fix: treat empty session ids as scoped search filters

### LINKED ISSUE #103: Bug: empty session_id silently disables session filter in MessageStore.search and SummaryDAG.search (cross-session leak under default scope)
## Summary

\`MessageStore.search\` (\`store.py:568\`) and \`SummaryDAG.search\` (\`dag.py:359\`) gate the session WHERE clause on Python truthiness:

\`\`\`python
if session_id:
    where.append(\"m.session_id = ?\")
    args.append(session_id)
\`\`\`

An empty string \`\"\"\` is falsy, so the predicate is omitted entirely and the FTS query runs across every session in the database.

\`LCMEngine\` initializes \`_session_id = \"\"\` (\`engine.py:242\`). When an agent calls \`lcm_grep\` (default \`session_scope='current'\`) before a real session is bound, the response advertises \`session_scope: 'current'\` while actually returning rows from every session in the local LCM database.

PR #102 added a defensive guard in \`lcm_grep\` itself (refuse the call when scope=current and \`engine._session_id\` is empty) but did not change the data-layer behavior. The leak is still reachable through any future caller that bypasses the tool surface.

## Reproduction

1. Initialize a fresh \`LCMEngine\` without setting \`_session_id\`.
2. Append messages under multiple session ids via \`engine._store.append(\"a\", ...)\`, \`engine._store.append(\"b\", ...)\`.
3. Call \`engine._store.search(\"keyword\", session_id=engine._session_id)\` (i.e., \`session_id=\"\"\`).
4. Result: hits from both \"a\" and \"b\".

## Expected

When \`session_id\` is an empty string (i.e., the engine has no current session bound), the search should either:
- Return zero results (treating empty as a real, never-matching session id), or
- Raise an explicit error so the caller cannot accidentally rely on falsy-truthy coercion.

## Suggested fix

Replace the truthiness check with an \`is None\` check in both \`store.py\` and \`dag.py\`, and document that an empty string is treated as a literal (non-matching) session id rather than an \"all sessions\" sentinel. \`session_scope='all'\` already passes \`session_id=None\` explicitly, so the contract stays consistent.

\`\`\`python
if session_id is not None:
    where.append(\"m.session_id = ?\")
    args.append(session_id)
\`\`\`

## Severity

Pre-existing on \`main\`. Made more visible by PR #102's \`from_current_session\` field (which consistently reports False under this state) but not introduced by it. Worth fixing because:
- It silently violates the documented current-session contract
- Any tool surface that bypasses lcm_grep's defensive guard reaches it
- It is a one-line fix with regression coverage straightforward to add

## Notes

Found during multi-agent code review of PR #102 (cross-session search). Filed separately so the data-layer fix can land independently of the tool-surface work.

stephenschoettler:
## Summary
- Treat `session_id=""` as a literal scoped session filter in `MessageStore.search` and `SummaryDAG.search`.
- Keep `session_id=None` as the explicit all-sessions sentinel.
- Add regression coverage for FTS, LIKE fallback, and unbound `lcm_grep` current-session behavior.

## Why
- `LCMEngine` starts with `_session_id = ""` before a real session is bound.
- The old truthiness checks treated that empty value like no filter, so current-session search could return rows from other sessions.
- The fix makes the data-layer contract explicit: only `None` means all sessions.

Fixes #103

## Validation
- RED: `python -m pytest tests/test_lcm_core.py::TestMessageStore::test_search_empty_session_id_does_not_search_all_sessions tests/test_lcm_core.py::TestMessageStore::test_search_like_fallback_empty_session_id_does_not_search_all_sessions tests/test_lcm_core.py::TestSummaryDAG::test_search_empty_session_id_does_not_search_all_sessions tests/test_lcm_core.py::TestSummaryDAG::test_search_like_fallback_empty_session_id_does_not_search_all_sessions tests/test_lcm_engine.py::TestEngineTools::test_handle_grep_unbound_current_session_does_not_search_all_sessions -q` failed before the fix with 5 failures.
- GREEN: same focused command passed after the fix, 5 passed.
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` hit local fd exhaustion after 349 passed.
- `ulimit -n 8192; python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=` passed, 379 passed, 38 warnings.
- `ulimit -n 8192; python -m pytest -q -o addopts=` passed, 416 passed, 38 warnings.
- `python -m compileall -q .` passed.
- `bash -n scripts/install.sh scripts/update.sh` passed.
- `git diff --check` passed.

## Notes
- Independent delegate review approved the diff before commit.
- No user-facing command schema changes.

### DIFF
--- a/dag.py
+++ b/dag.py
@@ -335,6 +335,8 @@ class SummaryDAG:
 
         Retrieval contract:
         - ``session_id`` limits which sessions are eligible
+        - ``session_id=None`` means all sessions; an empty string is treated as
+          a literal session id
         - ``source`` filters summaries by descendant raw-message lineage, not by
           session-level source presence
         - mixed-source nodes may match more than one ``source`` filter
@@ -356,7 +358,7 @@ class SummaryDAG:
         source_match_cache: dict[int, bool] = {}
         while True:
             try:
-                if session_id:
+                if session_id is not None:
                     rows = self._conn.execute(
                         f"""SELECT n.*, rank as search_rank FROM nodes_fts fts
                            JOIN summary_nodes n ON n.node_id = fts.rowid
@@ -420,7 +422,7 @@ class SummaryDAG:
 
         where: list[str] = ["summary IS NOT NULL"]
         args: list[Any] = []
-        if session_id:
+        if session_id is not None:
             where.append("session_id = ?")
             args.append(session_id)
         like_clauses = []
--- a/store.py
+++ b/store.py
@@ -538,6 +538,8 @@ class MessageStore:
 
         Retrieval contract:
         - ``session_id`` limits which sessions are eligible
+        - ``session_id=None`` means all sessions; an empty string is treated as
+          a literal session id
         - ``source`` limits which raw rows inside those sessions are eligible
         - ``source='unknown'`` means the explicit unknown-source bucket, with
           legacy blank-source rows treated as equivalent for back-compat
@@ -565,7 +567,7 @@ class MessageStore:
             try:
                 where = ["messages_fts MATCH ?"]
                 args: list[Any] = [safe_query]
-                if session_id:
+                if session_id is not None:
                     where.append("m.session_id = ?")
                     args.append(session_id)
                 if source_clause:
@@ -632,7 +634,7 @@ class MessageStore:
 
         where: list[str] = ["content IS NOT NULL"]
         args: list[Any] = []
-        if session_id:
+        if session_id is not None:
             where.append("session_id = ?")
             args.append(session_id)
         source_clause, source_args = _source_filter_clause("source", source)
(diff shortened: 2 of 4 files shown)


Tosko4 (approved):
Looks good to me. I rechecked the empty-string session contract across raw message search, DAG summary search, LIKE fallback, and the unbound current-session `lcm_grep` path.

The important bit is preserved: `session_id=None` still means all sessions, while `session_id=""` now stays a scoped literal and no longer leaks into cross-session search. I also ran the targeted regression slice, the core/engine/packaging slice, and the full suite locally. Full suite: `416 passed`.

PR #111, opened 2026-05-06
fix: reconcile ingest cursor after session restart

### LINKED ISSUE #109: Bug: gateway restart mid-session resets _ingest_cursor to 0, silently dropping all subsequent tool result rows
## Summary

Gateway restart mid-session resets `_ingest_cursor` to 0 while the store already has thousands of persisted messages for that session. After restart, `_ingest_messages()` silently ingests nothing — all subsequent tool call results (`role='tool'` rows) are never written to `lcm.db`. User/assistant text rows continue to be stored correctly, but tool results are permanently lost.

## Relationship to #1

This is the same root cause as issue #1 (fixed in 4bc6b9c via `_ingest_cursor`), but triggered by a **gateway restart mid-session** rather than a post-compaction cursor reset.

Fix 4bc6b9c handles the compaction case by resetting `_ingest_cursor = len(compressed)` in `compress()`. It does **not** handle the gateway restart case, where `on_session_start()` is called with the **same session ID** on reconnect.

## Related Issues

- **#1** (closed): Post-compaction ingestion cursor bug — same root cause, different trigger path
- **#95** (closed): SQLite lock in `on_session_end()` blocking gateway heartbeat — same lifecycle path, different failure mode

## Reproduction

1. Start a long-running session (any platform)
2. Accumulate many messages (e.g., 10,000+ store rows)
3. Restart the Hermes gateway mid-session (`hermes restart` or systemd restart)
4. Continue the session — send messages, make tool calls
5. Query `lcm.db` directly:

```sql
SELECT role, COUNT(*), source
FROM messages
WHERE session_id = '<your_session_id>'
  AND store_id > <store_id_at_restart>
GROUP BY role, source;
```

**Observed:** `role='tool'` count = 0 for all messages after restart. `role='assistant'` and `role='user'` rows are present.

**Expected:** All roles persisted continuously, including tool results.

## Evidence (production observation)

Confirmed via direct SQL query on `lcm.db` from a 2,288-message CLI session:

- Session `20260504_093451_e6889bdf`, 2,288 total messages
- Gateway restart occurred at store 11626 (Telegram gateway config change + `hermes restart`)
- **Stores 11627–13529 (~1,900 messages): zero `role='tool'` rows stored**
- Stores before 11626 and after 13529: tool results stored normally
- `role='assistant'` and `role='user'` messages stored throughout — confirming ingestion didn't fully stop, only tool results dropped

```sql
-- Confirms gap
SELECT MIN(store_id), MAX(store_id), COUNT(*)
FROM messages
WHERE session_id = '20260504_093451_e6889bdf'
  AND role = 'tool'
  AND source = 'cli';
-- Result: 11605 | 13636 | 62  (gap of ~1,900 stores in the middle)
```

## Root Cause

In `engine.py` `on_session_start()`, lines 1264–1266:

```python
else:
    self._clear_pending_reset_boundary()
    self._ingest_cursor = 0          # ← resets on same-session-id restart
    self._last_compacted_store_id = 0
```

This `else` branch fires when `previous_session_id == session_id` — i.e., when the gateway restarts and re-binds the same session. `_ingest_cursor` resets to 0, but the store already has e.g. 11,626 persisted rows.

When `on_session_end()` subsequently calls `_ingest_messages(messages)`, the in-memory `messages` list is the short post-restart context (e.g., 50 messages). The cursor at 0 means `new_messages = messages[0:]\ = all 50`. But the store count check in `_ingest_messages()` sees existing rows and the deduplication logic skips rows it believes are already stored — or the `messages` list never grows large enough to include the tool results appended during the post-restart run.

The net effect: tool result rows appended to the conversation **after** the gateway restart are never persisted.

## Suggested Fix

`on_session_start()` should detect a same-session-id restart and set `_ingest_cursor` to the existing store count for that session rather than resetting to 0:

```python
else:
    self._clear_pending_reset_boundary()
    # On same-session restart, pick up from existing store count
    # to avoid re-ingesting or silently dropping post-restart messages
    existing_count = self._store.get_session_count(self._session_id) if self._session_id else 0
    self._ingest_cursor = existing_count
    self._last_compacted_store_id = 0
    self._last_overflow_recovery_failed = False
    self._last_condensation_suppressed_reason = ""
```

This mirrors the pattern from the #1 fix — use a durable cursor independent of in-memory list length.

## Impact

- Tool call results are **silently and permanently lost** after any gateway restart mid-session
- Post-mortem debugging is impossible: the DB shows the correct session ID and all user/assistant text but no tool results — making it look like the agent made no tool calls
- Forensic investigation of operational issues (e.g., which terminal command deleted a file) becomes impossible for the affected window

stephenschoettler:
## Summary
- Reconcile the process-local ingest cursor when an existing session is rebound after a gateway restart.
- Match replayed active-context rows against a contiguous durable-store suffix before advancing the cursor.
- Skip explicit LCM scaffold messages, but preserve legitimate new system rows.
- Add restart regressions for normal, compacted, large, repeated non-tail, and new-system cases.

## Why
- Fixes #109.
- `_ingest_cursor` is process-local, but gateway restarts can rebind the same session with an existing durable message store.
- Reconstructing the cursor from the durable store prevents duplicate replay of old tail rows while preserving post-restart tool result rows.
- Suffix-only matching avoids silently dropping a new turn whose content repeats older non-tail history.

## Validation
- `python -m pytest tests/test_lcm_engine.py::TestEngineABC::test_existing_session_restart_reconciles_cursor_before_ingest tests/test_lcm_engine.py::TestEngineABC::test_existing_compacted_session_restart_skips_synthetic_context_but_persists_new_tool tests/test_lcm_engine.py::TestEngineABC::test_existing_large_session_restart_reconciles_beyond_short_tail_window tests/test_lcm_engine.py::TestEngineABC::test_existing_session_restart_does_not_skip_repeated_non_tail_messages tests/test_lcm_engine.py::TestEngineABC::test_existing_session_restart_persists_new_system_message -q` -> 5 passed
- `ulimit -n 8192 && python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> 384 passed, 38 warnings
- `ulimit -n 8192 && python -m pytest -q` -> 421 passed, 38 warnings
- `python -m compileall -q .`
- `git diff --check`
- Static scan clean for hardcoded secrets, shell injection, eval/exec, pickle, and obvious SQL formatting.
- Independent review passed after addressing Codex feedback.

## Notes
- Addressed Codex P1 by replacing arbitrary subsequence matching with durable-tail suffix matching.
- Addressed Codex P2 by narrowing scaffold detection so non-LCM system messages still persist.
- The default validation slice hit local `OSError: [Errno 24] Too many open files` under the default fd limit before rerunning. The same suite passed with `ulimit -n 8192`.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -259,7 +259,11 @@ class LCMEngine(ContextEngine):
         # messages have been persisted.  After compress() shortens the
         # list, the cursor resets to len(compressed) so that only
         # genuinely new messages (appended after compaction) get ingested.
+        # The cursor is process-local; existing sessions rebound after a
+        # gateway restart reconcile it against the durable store on the
+        # next ingest.
         self._ingest_cursor: int = 0
+        self._ingest_cursor_needs_reconcile = False
 
         # State required by ContextEngine ABC and run_agent.py compatibility
         self.model = ""
@@ -650,6 +654,7 @@ class LCMEngine(ContextEngine):
         # Reset cursor to the length of the compressed context so that
         # only messages appended *after* this point get ingested next time.
         self._ingest_cursor = len(compressed)
+        self._ingest_cursor_needs_reconcile = False
 
         logger.info(
             "LCM compaction #%d: %d messages → %d (%d leaf pass%s, %d→%d tokens, %d DAG nodes%s)",
@@ -1086,6 +1091,7 @@ class LCMEngine(ContextEngine):
         self.cache_metrics_available = False
         self._last_compacted_store_id = 0
         self._ingest_cursor = 0
+        self._ingest_cursor_needs_reconcile = False
         self._context_probed = False
         self._context_probe_persistable = False
         self._last_overflow_recovery_failed = False
@@ -1272,6 +1278,7 @@ class LCMEngine(ContextEngine):
             session_id,
             conversation_id=kwargs.get("conversation_id"),
         )
+        self._schedule_ingest_cursor_reconciliation()
         self._log_session_filter_diagnostics()
 
     def on_session_end(self, session_id: str, messages: List[Dict[str, Any]]) -> None:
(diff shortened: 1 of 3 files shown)


stephenschoettler:
Addressed the Codex review in `7caea99`.

Changes:
- Replaced arbitrary subsequence reconciliation with durable-tail suffix matching.
- Narrowed scaffold detection so only explicit LCM scaffold system messages are skipped.
- Added regressions for repeated non-tail content and legitimate new system messages.

Validation:
- Focused restart tests: 5 passed
- Default slice with `ulimit -n 8192`: 384 passed
- Full suite with `ulimit -n 8192`: 421 passed
- `compileall`, install/update shell syntax, and `git diff --check`: passed
- Independent re-review: passed

Tosko4 (approved):
Re-reviewed the current head `61d7312` against latest `main`. The main fix is in the right shape now: same-session restart reconstructs the process-local ingest cursor from the durable store, matches only against the durable tail, and keeps post-restart tool rows from falling through the cracks.

I found one remaining edge during review and pushed a small fix for it: system messages were still treated as scaffold if they merely mentioned `Lossless Context Management (LCM)`. That was too broad. `61d7312` narrows the skip to the actual LCM scaffold note and adds coverage for both sides: a legitimate new system message mentioning LCM is persisted, while the exact scaffold note is still skipped.

I also checked the earlier Codex points. The arbitrary-subsequence issue is gone; the current logic is suffix-based. The broad system-message skip is fixed by the new commit. I also ran a temp-copy probe against the local live LCM DB shape: rebinding a 3,372-row existing session, replaying 5 durable tail rows, then adding a new assistant/tool pair appended exactly 2 rows and left the cursor at 7. That is the behavior this PR needs.

Validation on `61d7312`:

`python -m pytest` for the 7 restart-focused tests: 7 passed

`python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=`: 386 passed

`python -m pytest -q -o addopts=`: 423 passed

`python -m compileall -q .`, `bash -n scripts/install.sh scripts/update.sh`, and `git diff --check origin/main...HEAD` are clean.

GitHub Actions is green on Python 3.11, 3.12, and 3.13. Approving from my side.

PR #114, opened 2026-05-06
fix: fall back to Hermes compression threshold

Tosko4:
## Summary

- Falls back to Hermes `compression.threshold` from `config.yaml` when `LCM_CONTEXT_THRESHOLD` is not set in the process environment.
- Keeps `LCM_CONTEXT_THRESHOLD` as the explicit override when it is present.
- Adds tests for both the Hermes-config fallback and env override precedence.

## Why

Hermes gateway/runtime config can have the intended compression threshold without exporting a matching `LCM_CONTEXT_THRESHOLD` env var into the LCM process. In that case LCM silently falls back to its built-in `0.75`, which can drift from the active Hermes compression config and reintroduce context-window warnings.

## Tests

- `python3 -m pytest tests/test_lcm_core.py -q`
- `python3 -m pytest -q`
- `python3 -m py_compile config.py`

### DIFF
--- a/config.py
+++ b/config.py
@@ -1,6 +1,12 @@
 """LCM configuration with defaults and env var overrides."""
 import os
 from dataclasses import dataclass, field
+from pathlib import Path
+
+try:
+    import yaml
+except Exception:  # pragma: no cover - optional fallback for minimal installs
+    yaml = None
 
 
 def _parse_pattern_list(raw: str) -> list[str]:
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Approved.

I reviewed the threshold fallback against latest `main`, including #118. The fallback keeps `LCM_CONTEXT_THRESHOLD` as the explicit override, otherwise reads Hermes `compression.threshold`, and fails closed to the existing default if the config is missing or malformed. The PyYAML/no-PyYAML paths are both covered.

Validation I ran on a clean merge with current `origin/main`:
- `python -m pytest tests/test_lcm_core.py::TestConfig -q -o addopts=` -> 6 passed
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py tests/test_import_lossless_claw.py -q -o addopts=` -> 439 passed
- `python -m pytest -q -o addopts=` -> 476 passed
- `python -m compileall -q .`
- `python -m py_compile config.py`
- Static scan clean for secrets, shell injection, eval/exec, pickle, and obvious SQL formatting.

PR #115, opened 2026-05-06
fix: encode importer SQLite URI paths

Tosko4:
## Summary

The importer opened source DBs and backup targets with hand-built SQLite URIs like `file:{path}?mode=ro`. Paths containing URI-reserved characters such as `#` or `?` were then parsed as URI syntax instead of as filesystem paths.

This changes those read-only connections to build the URI from `Path.resolve().as_uri()` before appending `?mode=ro`, and adds regressions for both the source DB path and backup target path cases.

## Validation

`python -m pytest tests/test_import_lossless_claw.py -q -o addopts=`: 9 passed

`python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py tests/test_import_lossless_claw.py -q -o addopts=`: 418 passed

`python -m pytest -q -o addopts=`: 455 passed

`python -m compileall -q .`, `bash -n scripts/install.sh scripts/update.sh`, and `git diff --check origin/main...HEAD` are clean.

Refs https://github.com/stephenschoettler/hermes-lcm/pull/107#discussion_r3193809311

PR #116, opened 2026-05-06
feat: add explicit LCM session loader

### LINKED ISSUE #112: Feature: session_load tool for full ordered raw-message transcript by session_id
## Summary

Add a `session_load` (or equivalently shaped) tool to `hermes-lcm` that returns the **full ordered raw-message transcript for a given `session_id`** in a single, bounded, paged tool call.

PR #102 (merged 2026-05-06) shipped the building blocks for this — `lcm_grep` now accepts `session_scope='session'` with explicit `session_id`, and `lcm_expand` gained a `store_id` mode for bounded raw-message retrieval regardless of session. Those primitives make targeted cross-session lookup possible. They do not, however, give the agent a clean path to **load an entire prior session's transcript** without an N-round-trip walk: one `lcm_grep` to enumerate the rows, then one `lcm_expand` per row to materialise raw content. For multi-hundred-message sessions that pattern is impractical at the tool-call level.

This issue proposes a single-call counterpart for the "give me the verbatim transcript of session X" operator use case, layered on the same data-layer primitives PR #102 already validated.

## Why

We run `hermes-lcm` v0.7.0 (now bumping to a `main` SHA past v0.8.0 to pick up PR #102 / #111) inside a Hermes deployment that uses Litestream to replicate `lcm.db` to S3. After an ECS task replacement, the next task restores the same `lcm.db` byte-for-byte from S3 and the agent inherits a fully populated archive of every prior session it has held — but the *current* Hermes session is brand new and has no shared `session_id` with any of those rows.

Concrete operator scenarios where the building blocks alone are awkward:

1. **"What did we talk about in our last session about X?"** — `session_search` returns a session id and a ~500-char summary preview, but the answer the user wants sits below the preview cutoff. Today the agent can either (a) hallucinate the rest, (b) tell the user the gap exists, or (c) start fan-out walking via repeated `lcm_grep(session_scope='session') + lcm_expand(store_id=...)` calls until token budget runs out. None of those is what an analyst wants when continuing a multi-week investigation across restarts.
2. **Post-Litestream-restore continuity check.** After a forced task replacement (we just shipped this last week), confirming that the previous task's conversation actually survived the restore currently means probing it tool-call by tool-call. A single `session_load` tool would let the agent reattach to a prior session in one call.
3. **Audit / runbook reconstruction.** A reviewer asking "show me the transcript that produced this finding" maps cleanly to "load that `session_id`", not to "build it up by paging row-by-row".

`session_search` (Hermes built-in) is correctly positioned as the *discovery* tool — it tells you which sessions exist and what they roughly contain. The proposed `session_load` is the natural *retrieval* counterpart for the plugin-local LCM database, in the same way `lcm_grep` is the discovery tool for in-database FTS and `lcm_expand` is its retrieval counterpart.

## Proposed scope

- New tool: `session_load` (name negotiable — `lcm_session_load`, `lcm_load_session`, etc.)
- Required arg: `session_id`
- Bounded by design (mirroring PR #102's clamping):
  - `limit` (clamped to a sensible cap, e.g. 200) — number of raw messages returned per call
  - `cursor` / `after_store_id` for pagination across calls
  - Optional `time_from` / `time_to` to slice a long session
  - Optional `roles=[...]` filter (subject to the same data-layer-pushdown constraint #108 is tracking)
- Result shape:
  - Ordered list of raw messages (chronological), each carrying `store_id`, `role`, `timestamp`, `content`, `source`
  - `next_cursor` for follow-up pages, or `null` when exhausted
  - `total_messages` for the session if cheaply computable
  - `session_id` echoed back for safety
- Cross-session DAG / summary expansion stays explicitly out of scope (consistent with PR #102's deferral) — `session_load` returns raw rows only, same provenance contract as `lcm_grep(session_scope='session')`
- Default behaviour rejects an empty / missing `session_id` (no implicit "current" or "all" fallback)

## Why a separate tool, not an `lcm_grep` extension

`lcm_grep` is search-shaped — it requires a `query` and returns ranked matches. The "load this whole session" use case is enumeration, not search. Forcing it through `lcm_grep` with an empty query (a) muddies the tool's contract, (b) burns the FTS path on a query that doesn't need it, and (c) loses the natural "ordered, paginated, bounded" semantics that a dedicated load tool gets for free.

An equivalent shape worth considering: extending `lcm_expand` with a `session_id` mode that streams ordered rows for a given session. The functional outcome would be similar; this issue uses `session_load` as the working name without prejudice to the final tool surface.

## Out of scope

- Cross-session DAG / summary-node expansion (deferred in #102; same deferral applies here)
- Hermes-side `session_search` integration changes
- Loading sessions that are not present in plugin-local `lcm.db`
- Any privacy / redaction layer beyond the existing local-DB threat model
- Schema changes to `messages` / `summary_nodes`

## Notes

This is a feature request grounded in an operator use case, layered on top of the primitives PR #102 already shipped. Filing it now while the cross-session work is fresh.

Tosko4:
## Summary

This adds `lcm_load_session`, a small raw transcript loader for #112. It is meant for the case where a caller already has a concrete LCM `session_id` and wants to enumerate that session directly instead of using search-shaped tools.

What changed:

- new `lcm_load_session` tool schema, handler registration, and manifest entry
- explicit `session_id` requirement, with no implicit current-session fallback
- ordered raw rows by `store_id`
- row pagination through `after_store_id` and `next_cursor`
- optional role and timestamp filters
- hard caps for row count and per-message content size
- truncation metadata so large individual rows can be recovered with `lcm_expand(store_id=...)`

@castaples this targets the workflow from #112. @stephenschoettler could you take a look when you get a chance?

Closes #112

## Validation

- `python -m pytest tests/test_lcm_engine.py::TestEngineABC::test_tool_schemas tests/test_lcm_engine.py::TestEngineABC::test_readme_documents_session_scope_contract tests/test_lcm_engine.py::TestHandleLoadSession tests/test_packaging_install.py::test_plugin_manifest_lists_all_registered_tools tests/test_packaging_install.py::test_plugin_entrypoint_registers_lcm_context_engine -q -o addopts=`
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py tests/test_import_lossless_claw.py -q -o addopts=`
- `python -m pytest -q -o addopts=`
- `python -m compileall -q .`
- `python -m py_compile scripts/import_lossless_claw.py`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,13 +149,13 @@ Expected signals:
 
 - plugin list includes `hermes-lcm`
 - selected context engine is `lcm`
-- tool list includes `lcm_grep`, `lcm_describe`, `lcm_expand`, `lcm_expand_query`, `lcm_status`, and `lcm_doctor`
+- tool list includes `lcm_grep`, `lcm_load_session`, `lcm_describe`, `lcm_expand`, `lcm_expand_query`, `lcm_status`, and `lcm_doctor`
 
 Typical output:
 
 ```text
 Plugins (1):
-  ✓ hermes-lcm v0.8.0 (6 tools)
+  ✓ hermes-lcm v0.8.0 (7 tools)
 
 Provider Plugins:
   Context Engine: lcm
@@ -249,6 +249,7 @@ for earlier separate sessions or broad cross-session history.
 | Tool | Use |
 |------|-----|
 | `lcm_grep` | Search current-session raw messages and summaries. Opt into `session_scope='all'` or `session_scope='session'` (with `session_id`) for bounded archive recovery over rows already present in `lcm.db`, including externally backfilled rows that may carry source strings such as `openclaw-lcm:*`; broader scopes return raw-message hits only. Use `session_search` for earlier separate sessions or broad cross-session recall. |
+| `lcm_load_session` | Load one ordered raw-message transcript page for an explicit `session_id`. This is not search: it returns raw rows in `store_id` order, bounded by `limit`, with per-message content bounded by `max_content_chars`, and continues with `after_store_id` from `next_cursor`. |
 | `lcm_describe` | Inspect the current-session DAG or preview an `externalized_ref` without loading full content. |
 | `lcm_expand` | Recover source messages, child summaries, or externalized payloads with pagination. Use `store_id` to fetch a single raw message regardless of session, suitable for drilling into a cross-session `lcm_grep` result. |
 | `lcm_expand_query` | Answer a question using expanded current-session LCM context while returning a bounded answer. |
@@ -260,8 +261,9 @@ for earlier separate sessions or broad cross-session history.
 LCM retrieval tools default to current-session scope. `lcm_grep` accepts
 `session_scope='all'` or `session_scope='session'` as an explicit opt-in for
 bounded archive search over rows already present in `lcm.db` (raw-message hits
-only); use Hermes `session_search` for broad cross-session history outside the
-LCM database.
+only). Once a session id is known, `lcm_load_session` can enumerate that session's
+raw transcript in chronological `store_id` pages without a search query. Use
+Hermes `session_search` for broad cross-session history outside the LCM database.
 
 Within the current session, `source` filters raw rows directly and filters
 summary nodes by descendant raw-message source lineage. `unknown` is a real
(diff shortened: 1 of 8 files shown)


stephenschoettler (approved):
Approved.

I reviewed `lcm_load_session` against latest `main` after #115. The tool has the right shape for #112: explicit `session_id`, no implicit current/all fallback, store-order pagination with `after_store_id`/`next_cursor`, bounded rows, bounded per-message content, truncation metadata, and raw rows only. Filters are pushed into the SQL layer with placeholders.

Validation I ran on a clean merge with current `origin/main`:
- Focused schema/session-load/manifest slice -> 11 passed
- `ulimit -n 8192 && python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py tests/test_import_lossless_claw.py -q -o addopts=` -> 449 passed
- `ulimit -n 8192 && python -m pytest -q -o addopts=` -> 486 passed
- `python -m compileall -q .`
- `python -m py_compile scripts/import_lossless_claw.py`

GitHub CI is green on the updated branch.

PR #118, opened 2026-05-06
feat: add message-level ignore patterns for LCM ingest

### LINKED ISSUE #117: Add message-level ignore patterns for cron/noise suppression
## Problem

`LCM_IGNORE_SESSION_PATTERNS` and `LCM_STATELESS_SESSION_PATTERNS` already provide noise suppression at the session/source level (e.g. `cron:*`). They do not catch a real and recurring case: Hermes cron deliveries that arrive inside a normal Telegram or WhatsApp conversation as ordinary messages whose content starts with a clear marker like:

```
Cronjob Response: hermie-hourly-heartbeat
(job_id: eee06bdbb09b)
-------------
```

or sometimes:

```
>>>Cronjob Response<<<: hermie-hourly-heartbeat
(job_id: eee06bdbb09b)
-------------
```

From LCM's perspective the session/platform is `telegram`, not `cron`, so session-level globs do not catch them. These alerts pollute LCM history and cross-session search results.

## Proposed feature

A new env var `LCM_IGNORE_MESSAGE_PATTERNS`: a comma-separated list of Python regex patterns. If any pattern matches a message's normalized content, the message is skipped before `append_batch`/store ingest, so it never enters the messages table, FTS index, or DAG node lineage.

### Design constraints

- Disabled by default. Default value is empty.
- Patterns are compiled once at engine startup, not per message.
- Compilation errors are tolerated: invalid regex emits a `logger.warning` and is skipped, valid patterns from the same list still take effect.
- Filtering applies to all roles by default, since cron alerts can arrive under any role depending on the gateway.
- Existing `LCM_IGNORE_SESSION_PATTERNS` / `LCM_STATELESS_SESSION_PATTERNS` behavior is preserved unchanged.
- `lcm_status` surfaces the configured patterns, source string (`default` / `env`), and a process-lifetime `ignored_message_count` counter.
- Anchored patterns (e.g. `^Cronjob Response:`) are best-effort against plain-text content. Multimodal/structured content is normalized via `normalize_content_value`, which produces a JSON wrapper and breaks `^` anchoring; operators who need multimodal coverage should use unanchored patterns.

### Example operator config

```
LCM_IGNORE_MESSAGE_PATTERNS=^Cronjob Response:,^>>>Cronjob Response<<<:
```

### Out of scope (deferred to follow-up)

Filtering matching messages out of the compaction chunk fed to the summarizer. On the rare turn where a matching message sits inside the chunk being compacted, its text may briefly appear in the resulting summary node text (its `store_id` will not appear in the node's `source_ids`, because it was never written to the store). A separate issue will track the compaction-input filter once this PR ships and stabilizes.

## Use case

Hermes hourly heartbeat cron sends Telegram alerts starting with `Cronjob Response: hermie-hourly-heartbeat`. Those alerts are useful live in chat, but they pollute LCM history and cross-session search. Normal Telegram conversation should still be stored. Only matching cron announcement messages should be ignored.

GodsBoy:
Closes #117.

## Summary

Adds `LCM_IGNORE_MESSAGE_PATTERNS`, a comma-separated list of Python regex patterns that drops noise out of the immutable LCM message store at ingest time. Designed for the case where Hermes cron alerts (e.g. `Cronjob Response: hermie-hourly-heartbeat`) are delivered into a normal Telegram or WhatsApp conversation as ordinary user-visible messages, so session-level globs like `cron:*` cannot catch them.

The implementation mirrors the existing `LCM_IGNORE_SESSION_PATTERNS` shape: dataclass field + source-diagnostics field, env-parse block in `LCMConfig.from_env()`, dedicated helper module (`message_patterns.py`), one-shot diagnostic log gated by the existing `_logged_filter_config` flag, and three new keys in `lcm_status`. Default behavior is unchanged: when the env var is absent, no filtering happens and all existing tests still pass.

## What's in this PR

- `config.py`: `ignore_message_patterns: list[str]` + `ignore_message_patterns_source: str` fields, `from_env()` block reading `LCM_IGNORE_MESSAGE_PATTERNS`.
- `message_patterns.py` (new): `compile_message_patterns(patterns)` which catches `re.error` per-pattern with a `logger.warning`, and `matches_message_pattern(text, patterns)` using `re.search` so user-supplied anchors (`^`, `\b`) and inline flags (`(?is)`) work as written.
- `engine.py`: compile patterns at init, filter matching messages in `_ingest_messages` before `append_batch`, increment a process-lifetime `_ignored_message_count` counter, surface the new keys in `lcm_status`, extend `_log_session_filter_diagnostics`.
- `tests/test_lcm_core.py`: helper-level `TestMessagePatterns`, plus `LCM_IGNORE_MESSAGE_PATTERNS` coverage in `TestConfig`.
- `tests/test_lcm_engine.py`: new `TestMessageFiltering` class covering anchored matching, wrapper variants, inline-flag patterns, the must-not-regress baseline (active pattern + non-matching messages), multimodal content (both anchored-does-not-match and unanchored-substring-does-match cases), role-agnostic filtering, invalid-regex tolerance, coexistence with `LCM_IGNORE_SESSION_PATTERNS`, status surface, and once-per-engine-instance diagnostic logging.
- `README.md`: env var table row + a noise-suppression section explaining when to use session-level vs message-level filtering, the operator example config line, and two documented limitations.

## Operator example

```
LCM_IGNORE_MESSAGE_PATTERNS=^Cronjob Response:,^>>>Cronjob Response<<<:
```

`lcm_status` then reports `ignore_message_patterns`, `ignore_message_patterns_source` (`default` or `env`), and a process-lifetime `ignored_message_count` so operators can confirm their config was loaded and watch how often it fires.

## Documented limitations

- **Anchored patterns are best-effort against multimodal content.** Structured payloads (lists of content parts) are normalized via `normalize_content_value`'s JSON serialization before matching, so `^Cronjob Response:` binds to the JSON wrapper rather than to the inner text. Operators who need multimodal coverage should write unanchored patterns. Tests assert both the anchored-does-not-match and unanchored-substring-does-match cases explicitly so the contract is documented in code.
- **Compaction-window edge.** The filter runs at ingest. On a rare turn where a matching message is part of the chunk being summarized in the same turn it arrived, its text may briefly appear inside the resulting summary node text. The summary node's `source_ids` will not reference the filtered message (it was never written to the store), so DAG lineage stays clean. Closing this window cleanly requires also adjusting the `len(compacted_chunk)`-based slice arithmetic in `compress()`, which materially expands the diff; tracked as follow-up work.

## Test plan

- [x] `pytest tests/test_lcm_core.py::TestMessagePatterns` — 6 helper tests pass.
- [x] `pytest tests/test_lcm_core.py::TestConfig` — defaults + from_env coverage pass with the new field.
- [x] `pytest tests/test_lcm_engine.py::TestMessageFiltering` — 12 ingest/status/log scenarios pass.
- [x] `pytest tests/test_lcm_engine.py::TestSessionFiltering` — regression guard for session-level filter, all green.
- [x] `pytest tests/test_lcm_engine.py::TestEngineIngest` — regression guard for cursor invariant, all green.
- [x] Full suite: 471 passed (up from 453 before this PR).

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,7 @@ environment variables:
 | `LCM_NEW_SESSION_RETAIN_DEPTH` | `2` | DAG depth retained after manual `/new` (`-1` all, `0` none) |
 | `LCM_IGNORE_SESSION_PATTERNS` | empty | Comma-separated session globs excluded from LCM storage |
 | `LCM_STATELESS_SESSION_PATTERNS` | empty | Comma-separated session globs kept read-only |
+| `LCM_IGNORE_MESSAGE_PATTERNS` | empty | Comma-separated regex patterns; matching message content (plain text, or the normalized form for structured/multimodal content) is excluded from LCM storage |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_ENABLED` | `false` | Store oversized tool outputs in plugin-managed JSON files |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_THRESHOLD_CHARS` | `12000` | Externalization threshold for tool output text |
 | `LCM_LARGE_OUTPUT_TRANSCRIPT_GC_ENABLED` | `false` | Rewrite already-externalized summarized tool rows to compact placeholders |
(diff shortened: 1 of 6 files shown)


Tosko4 (approved):
Reviewed the current head after pushing one small maintainer follow-up commit.

The issue is valid: session-level filters do not catch cron deliveries that are injected into normal chat sessions, and this keeps the new behavior opt-in instead of hard-coding that setup. The follow-up commit fixes the restart cursor edge I found, so filtered active-context messages and historical filtered store rows are both ignored during replay matching instead of causing duplicate ingest after restart. I also tightened the README wording around the compaction-window caveat so it does not overpromise about DAG summary text.

Validation on 3264d8e: the focused filter/config/session/ingest slice passed locally, the broad core/engine/packaging/importer slice passed, the full suite passed, and compile/syntax/diff checks are clean. GitHub CI is green on 3.11, 3.12, and 3.13.

Approving from my side.

Issue #119, opened 2026-05-06
Follow-up review findings from PR #118 (LCM_IGNORE_MESSAGE_PATTERNS)


GodsBoy:
Multi-agent code review of PR #118 surfaced findings split across three categories: small uncontroversial fixes (already opened as a follow-up PR), a pre-existing tool-handler gap that affects all pattern types and warrants a separate scope discussion, and items that need maintainer input on fix shape before any code change lands.

## Reviewers

- correctness, testing, maintainability, project-standards (always-on)
- ce-agent-native-reviewer, ce-learnings-researcher (CE always-on)
- api-contract (lcm_status JSON shape gained 3 keys)
- adversarial (398 substantive changed lines, hot ingest-path regex matching)
- kieran-python (Python diff with new module + engine wiring)

### #1 (pre-existing) - `lcm_status` tool handler does not forward pattern-config keys to agents

**Status: not introduced by #118.** The `tools.py:lcm_status` handler at `tools.py:1160-1243` builds its own JSON dict and only forwards a curated subset of `engine.get_status()`. Specifically, the existing `session_filters` block at `tools.py:1235-1238` only forwards two booleans:

\`\`\`python
\"session_filters\": {
    \"ignored\": engine._session_ignored,
    \"stateless\": engine._session_stateless,
},
\`\`\`

It has never forwarded `ignore_session_patterns`, `ignore_session_patterns_source`, `stateless_session_patterns`, or `stateless_session_patterns_source`. PR #118 added three new keys to `engine.get_status()` (`ignore_message_patterns`, `ignore_message_patterns_source`, `ignored_message_count`) and they follow the same pattern - also not forwarded.

**Options for discussion:**
1. Surface all pattern config (session, stateless, message) via `lcm_status` together. Treats agent visibility as a uniform contract for all three pattern types.
2. Keep the curated subset and update README docs to clarify that pattern-config keys live on `engine.get_status()` but not on the agent-facing `lcm_status` tool.
3. Surface only the newly-added message-pattern keys (asymmetric, probably worse than option 1 or 2).

**Why this matters:** the README and #118 PR body both claim `lcm_status` surfaces the new keys. The claim is true at the engine layer (which is what `TestMessageFiltering::test_status_surfaces_message_pattern_keys` asserts) but the agent-facing tool surface is the one operators actually consume.

### #3 ReDoS guard for ingest path

[...]

PR #120, opened 2026-05-06
chore: address review follow-ups for message-ignore-patterns

GodsBoy:
Refs #119.

Small uncontroversial follow-ups from the multi-agent code review of PR #118 (`feat: add message-level ignore patterns for LCM ingest`). The full review report is on issue #119; this PR addresses only the items that don't need design discussion.

## What's in this PR

- **README framing fix.** The compaction-window edge case in the noise-suppression section was described as occurring "On a rare turn where..." Adversarial review's traffic math (hourly cron alerts + ~40-turn compaction cadence) suggested the framing oversold the guarantee. Updated to "When..." with honest language about long-running sessions where this can affect multiple summary nodes per day. Behavior unchanged.
- **Stateless + message-filter coexistence test.** Verifies the message filter is correctly bypassed when a session matches `stateless_session_patterns`. Only the ignored-session case was tested before; the stateless early-return path is identical but uncovered. (Three of the review reviewers independently flagged this gap.)
- **All-filtered cursor regression test.** When every message in a batch matches a configured pattern, `_ingest_cursor` advances to `len(messages)` and a subsequent call with the same list does not re-process. Asserts both that the cursor advances correctly and that the counter does not double-increment on the second call.
- **`LCM_STATUS` schema description update.** Mentions session-filter state explicitly so agents diagnosing missing messages know to look there.

## Intentionally excluded

The review surfaced three items that need maintainer input on fix shape before any code change lands. They are documented in #119 for discussion:

- **Pre-existing `lcm_status` tool-handler gap.** `tools.py:lcm_status` builds its own JSON dict and forwards a curated subset of `engine.get_status()`. The `session_filters` block currently only forwards the `ignored` and `stateless` booleans, not the raw pattern lists or source strings. This applies to all pattern types (session, stateless, message), not just the new ones from #118. The fix is straightforward but represents a contract expansion that affects existing behavior, so worth discussing first.
- **ReDoS guard for ingest path.** Operator-supplied regex without length cap or timeout can stall ingest synchronously. Multiple fix shapes possible (input length cap, `regex` library migration with timeout support, threading watchdog).
- **Multimodal anchored-pattern UX.** Documented as a known limitation; design call about whether to extract text parts before matching or keep current behavior.

## Test plan

- [x] `pytest tests/test_lcm_engine.py::TestMessageFiltering` -> 16 passed (was 14 in #118; +2 new tests)
- [x] `pytest -q` -> 478 passed (full suite, no regressions)

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,13 +270,15 @@ Two operator-facing limitations to know about:
   JSON wrapper rather than to the inner text. If you expect cron alerts to
   arrive as multimodal payloads from your gateway, use unanchored patterns
   (for example `Cronjob Response:` instead of `^Cronjob Response:`).
-- **Compaction-window edge.** The filter runs at ingest time. On a rare
-  turn where a matching message is part of the chunk being summarized in
-  the same turn it arrived, the message's text may briefly appear inside
-  the resulting summary node text. The summary node's `source_ids` will
-  not reference the filtered message (it was never written to the store),
-  so DAG lineage stays clean; only the serialized summary text can carry
-  it. Closing this window is tracked as follow-up work.
+- **Compaction-window edge.** The filter runs at ingest time. When a
+  matching message is part of the chunk being summarized in the same
+  turn it arrived, the message's text may appear inside the resulting
+  summary node text. In long-running sessions where compaction triggers
+  every several dozen turns, this can affect multiple summary nodes per
+  day rather than only happening rarely. The summary node's `source_ids`
+  will not reference the filtered message (it was never written to the
+  store), so DAG lineage stays clean; only the serialized summary text
+  can carry it. Closing this window is tracked as follow-up work.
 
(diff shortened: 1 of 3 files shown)


stephenschoettler (approved):
Approved.

This is the right narrow follow-up for #119: it lands the low-risk pieces only, while leaving the `lcm_status` contract expansion, regex ReDoS guard, and multimodal matching behavior for explicit maintainer decisions.

Validation against latest `main`:
- Clean merge with current `origin/main`.
- `python -m pytest tests/test_lcm_engine.py::TestMessageFiltering tests/test_lcm_engine.py::TestEngineABC::test_tool_schemas -q -o addopts=` -> 17 passed
- `ulimit -n 8192 && python -m pytest -q -o addopts=` -> 488 passed
- `python -m compileall -q .`
- GitHub CI is green on the updated branch.

PR #122, opened 2026-05-06
fix: guard message ignore regex matching

### LINKED ISSUE #119: Follow-up review findings from PR #118 (LCM_IGNORE_MESSAGE_PATTERNS)

stephenschoettler:
## Summary
- Add per-pattern timeout handling for message ignore regex matching when the optional `regex` engine is available.
- Treat regex timeouts as non-matches, log once per pattern, and continue evaluating later patterns so ingest remains available.
- Keep minimal installs importable with a capped stdlib `re` fallback when `regex` is unavailable.
- Cache stdlib fallback detection after the first `TypeError` so minimal installs do not throw an exception on every message.
- Document the timeout/fallback behavior.

## Why
- Refs #119.
- `LCM_IGNORE_MESSAGE_PATTERNS` runs in synchronous ingest. A pathological operator regex can otherwise stall the agent instead of merely failing to filter noise.
- Minimal installs should not pay per-message exception overhead when timeout support is absent.
- This keeps the plugin's no-required-dependencies posture while using `regex` timeouts in environments that have it installed.

## Validation
- RED: `python -m pytest tests/test_lcm_core.py::TestMessagePatterns::test_stdlib_timeout_fallback_is_cached_after_first_type_error -q` failed before the fix because timeout fallback raised `TypeError` once per message.
- `python -m pytest tests/test_lcm_core.py::TestMessagePatterns -q` -> 8 passed
- `python -m compileall -q message_patterns.py tests/test_lcm_core.py`
- `git diff --check`

## Notes
- This does not close #119. Multimodal/text-part matching remains separate.
- In minimal environments without `regex`, the fallback only caps the text sent to stdlib `re`; it cannot enforce a true per-match timeout.
- Branch was rebased onto current `origin/main` at `f5f3932`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ claim that Hermes core has no persisted record of pre-compression history.
 
 - Hermes Agent with the pluggable context engine slot ([PR #7464](https://github.com/NousResearch/hermes-agent/pull/7464))
 - Python 3.11+
-- No required third-party runtime dependencies. `tiktoken` is used if available; otherwise LCM falls back to character-based token estimates.
+- No required third-party runtime dependencies. `tiktoken` is used if available; otherwise LCM falls back to character-based token estimates. `regex` is used if available to apply timeouts to message ignore patterns; otherwise LCM keeps a capped stdlib `re` fallback.
 
 ## Install
 
@@ -263,7 +263,10 @@ LCM_IGNORE_MESSAGE_PATTERNS=^Cronjob Response:,^>>>Cronjob Response<<<:
 
 Invalid regex entries are logged at warning level and dropped; the
 surviving patterns in the same list still take effect, so a misconfigured
-entry never crashes ingest.
+entry never crashes ingest. Pattern matching uses a 50 ms per-pattern timeout
+when the optional `regex` package is installed. If only stdlib `re` is
+available, LCM keeps the plugin importable and caps fallback matching to the
+first 100,000 characters of the normalized content.
 
 Two operator-facing limitations to know about:
 
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ claim that Hermes core has no persisted record of pre-compression history.
 
 - Hermes Agent with the pluggable context engine slot ([PR #7464](https://github.com/NousResearch/hermes-agent/pull/7464))
 - Python 3.11+
-- No required third-party runtime dependencies. `tiktoken` is used if available; otherwise LCM falls back to character-based token estimates. `regex` is used if available to apply timeouts to message ignore patterns; otherwise LCM keeps a capped stdlib `re` fallback.
+- No required third-party runtime dependencies. `tiktoken` is used if available; otherwise LCM falls back to character-based token estimates. `regex` is used if available to apply timeouts to message ignore patterns; if it is not installed, message-level regex filtering is disabled with a warning rather than running unbounded stdlib `re` matches.
 
 ## Install
 
@@ -264,18 +264,12 @@ LCM_IGNORE_MESSAGE_PATTERNS=^Cronjob Response:,^>>>Cronjob Response<<<:
 Invalid regex entries are logged at warning level and dropped; the
 surviving patterns in the same list still take effect, so a misconfigured
 entry never crashes ingest. Pattern matching uses a 50 ms per-pattern timeout
-when the optional `regex` package is installed. If only stdlib `re` is
-available, LCM keeps the plugin importable and caps fallback matching to the
-first 100,000 characters of the normalized content.
+when the optional `regex` package is installed. If `regex` is not installed,
+LCM logs a warning and disables message-level regex filtering rather than
+running unbounded stdlib `re` matches in the ingest path.
 
-Two operator-facing limitations to know about:
+One operator-facing limitation to know about:
 
-- **Anchored patterns are best-effort against multimodal content.**
-  Structured payloads (lists of content parts) are normalized via JSON
-  serialization before matching, so a `^`-anchored pattern binds to the
-  JSON wrapper rather than to the inner text. If you expect cron alerts to
-  arrive as multimodal payloads from your gateway, use unanchored patterns
-  (for example `Cronjob Response:` instead of `^Cronjob Response:`).
 - **Compaction-window edge.** The filter runs at ingest time. When a
   matching message is part of the chunk being summarized in the same
   turn it arrived, the message's text may appear inside the resulting
--- a/message_patterns.py
+++ b/message_patterns.py
@@ -2,31 +2,46 @@
 
 Patterns are Python regex strings. Compilation is tolerant: an invalid
 pattern emits a warning and is skipped, leaving valid patterns in the
-same list still active. Matching uses ``re.search`` so user-supplied
-anchors (``^``, ``\\b``) and inline flags (``(?is)``) work as written.
+same list still active. Matching passes a per-pattern timeout when the
+runtime regex engine supports it so catastrophic backtracking cannot
+stall synchronous ingest indefinitely. User-supplied anchors (``^``,
+``\b``) and inline flags (``(?is)``) work as written.
 """
 
 from __future__ import annotations
 
 import logging
 import re
-from typing import Iterable, List
+from typing import Any, Iterable
+
+try:  # pragma: no cover - exercised when the optional dependency is absent
+    import regex as _regex_engine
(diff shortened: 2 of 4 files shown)


stephenschoettler:
Pushed `b19a3c3` after rebasing onto current `main`.

Addressed the Codex P2 finding by caching timeout-support detection after the first stdlib fallback `TypeError`, so minimal installs do not throw and catch on every message.

Validation:
- RED regression failed before the fix: `test_stdlib_timeout_fallback_is_cached_after_first_type_error`
- `python -m pytest tests/test_lcm_core.py::TestMessagePatterns -q` -> 8 passed
- `python -m compileall -q message_patterns.py tests/test_lcm_core.py`
- `git diff --check`

Tosko4 (changes requested):
I think this still needs one more pass before merge.

The optional `regex` path is fine, but the minimal install path is still exposed to catastrophic backtracking. In this repo’s default CI/runtime shape `regex` is not installed, so `_search_with_timeout()` falls back to stdlib `re` and runs `pattern.search(text[:100_000])` with no actual timeout. That 100k cap does not protect the hot ingest path from pathological patterns: locally, `(a+)+$` against only 22 `a`s plus `!` already took about 0.48s, and the curve gets ugly fast.

Since #119’s point here is ingest availability, I would either make the timeout-capable engine a real dependency for this feature path, or make the stdlib fallback fail much more safely instead of presenting the cap as a guard. A conditional test that exercises the real `regex` timeout path would also help, since the current tests mostly simulate timeout behavior with fake patterns.

Local checks I ran: `TestMessagePatterns` passed, compile/diff checks passed, and the stdlib fallback timing probe above reproduced the remaining risk.

Tosko4 (approved):
I pushed `96bfb32` to close the blocker I had requested changes on.

The previous stdlib `re` fallback still allowed catastrophic backtracking in minimal installs. A character cap was not a real availability guard, so this now takes the safer route: if the optional `regex` package is not installed, configured message-level regex filters are disabled with a warning rather than running unbounded stdlib matching in the ingest path.

That is fail-open for filtering, but fail-safe for the gateway. Operators who need `LCM_IGNORE_MESSAGE_PATTERNS` active should install `regex`; minimal installs remain importable and do not hang on a bad pattern.

Validation on `96bfb32`:

- message-pattern + engine filtering slice: 30 passed
- explicit no-hang smoke with missing `regex`: passed under `timeout 2s`
- broad core/engine/command/packaging/importer slice: 504 passed
- full suite: 504 passed

Approved.

PR #123, opened 2026-05-06
fix: match structured message text parts

### LINKED ISSUE #119: Follow-up review findings from PR #118 (LCM_IGNORE_MESSAGE_PATTERNS)

stephenschoettler:
## Summary
- Match `LCM_IGNORE_MESSAGE_PATTERNS` against structured-content text parts before normalized JSON fallback.
- Parse nested text payloads such as `text.value` and text stored under `content` for text blocks.
- Let anchored cron/noise patterns match the text an operator sees in multimodal payloads.
- Preserve normalized JSON fallback for structured payloads with no text parts.
- Update docs for the new text-first contract.

## Why
- Refs #119.
- Current matching serialized some structured text blocks to JSON first, so `^Cronjob Response:` did not match operator-visible text when the gateway wrapped it as nested structured content.
- Operators should not need weaker unanchored patterns only because content is represented as structured text parts.

## Validation
- RED: `python -m pytest tests/test_lcm_engine.py::TestMessageFiltering::test_anchored_pattern_matches_structured_text_value_parts -q` failed before the fix because both structured text-value messages were persisted.
- `python -m pytest tests/test_lcm_engine.py::TestMessageFiltering::test_anchored_pattern_matches_structured_text_value_parts tests/test_lcm_engine.py::TestMessageFiltering::test_anchored_pattern_matches_multimodal_text_part_content tests/test_lcm_engine.py::TestMessageFiltering::test_structured_content_without_text_parts_falls_back_to_normalized_json -q` -> 3 passed
- `python -m pytest tests/test_lcm_engine.py::TestMessageFiltering -q` -> 18 passed
- `python -m compileall -q message_content.py tests/test_lcm_engine.py`
- `git diff --check`

## Notes
- This does not close #119 by itself while #121 and #122 are still pending review/merge.
- The matching contract is text-first for structured content with text parts, then normalized JSON fallback when no text parts exist.
- Branch was rebased onto current `origin/main` at `f5f3932`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ environment variables:
 | `LCM_NEW_SESSION_RETAIN_DEPTH` | `2` | DAG depth retained after manual `/new` (`-1` all, `0` none) |
 | `LCM_IGNORE_SESSION_PATTERNS` | empty | Comma-separated session globs excluded from LCM storage |
 | `LCM_STATELESS_SESSION_PATTERNS` | empty | Comma-separated session globs kept read-only |
-| `LCM_IGNORE_MESSAGE_PATTERNS` | empty | Comma-separated regex patterns; matching message content (plain text, or the normalized form for structured/multimodal content) is excluded from LCM storage |
+| `LCM_IGNORE_MESSAGE_PATTERNS` | empty | Comma-separated regex patterns; matching message content (plain text, extracted text parts for structured/multimodal content, or normalized JSON fallback when no text parts exist) is excluded from LCM storage |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_ENABLED` | `false` | Store oversized tool outputs in plugin-managed JSON files |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_THRESHOLD_CHARS` | `12000` | Externalization threshold for tool output text |
 | `LCM_LARGE_OUTPUT_TRANSCRIPT_GC_ENABLED` | `false` | Rewrite already-externalized summarized tool rows to compact placeholders |
@@ -246,11 +246,14 @@ noise:
   so only the message content is distinctive.
 
 Message-level patterns are Python regex strings, comma-separated, compiled
-once at engine start. They run against the normalized message content (the
-same string LCM would have written to the store). Matching messages are
-skipped before storage, so new matching rows do not enter the messages table
-or FTS index. Filtering is role-agnostic by default, since cron alerts can
-be re-emitted under any role depending on the gateway.
+once at engine start. They run against plain message text. For structured
+multimodal payloads, LCM matches against concatenated text parts first, so
+anchored patterns bind to the text an operator sees. If a structured payload
+contains no text parts, matching falls back to the normalized JSON form that
+LCM would have written to the store. Matching messages are skipped before
+storage, so new matching rows do not enter the messages table or FTS index.
+Filtering is role-agnostic by default, since cron alerts can be re-emitted
+under any role depending on the gateway.
 
 Example operator config:
 
--- a/engine.py
+++ b/engine.py
@@ -49,7 +49,11 @@ from .session_patterns import (
 )
 from .message_patterns import compile_message_patterns, matches_message_pattern
 from .lifecycle_state import LifecycleStateStore
-from .message_content import normalize_content_value, text_content_for_pattern_matching
+from .message_content import (
+    normalize_content_value,
+    stored_text_content_for_pattern_matching,
+    text_content_for_pattern_matching,
+)
 from .store import MessageStore
 from .tokens import count_message_tokens, count_messages_tokens, count_tokens
 from . import tools as lcm_tools
@@ -1793,10 +1797,15 @@ class LCMEngine(ContextEngine):
             logger.debug("LCM ingest cursor reconciliation probe failed: %s", exc)
             self._ingest_cursor_needs_reconcile = False
 
-    def _matches_ignore_message_patterns(self, msg: Dict[str, Any]) -> bool:
+    def _matches_ignore_message_patterns(self, msg: Dict[str, Any], *, stored_row: bool = False) -> bool:
         if not self._compiled_ignore_message_patterns:
             return False
-        text = text_content_for_pattern_matching(msg.get("content")) or ""
+        content = msg.get("content")
+        text = (
+            stored_text_content_for_pattern_matching(content)
+            if stored_row
+            else text_content_for_pattern_matching(content)
+        ) or ""
         return matches_message_pattern(text, self._compiled_ignore_message_patterns)
 
     def _is_replayed_context_scaffold_message(self, msg: Dict[str, Any]) -> bool:
@@ -1891,7 +1900,7 @@ class LCMEngine(ContextEngine):
         stored_tail = [
             self._message_replay_identity(row)
             for row in stored_rows
-            if not self._matches_ignore_message_patterns(row)
+            if not self._matches_ignore_message_patterns(row, stored_row=True)
         ]
         cursor = self._find_reconciled_cursor_for_store_tail(
             messages,
--- a/engine.py
+++ b/engine.py
@@ -49,7 +49,7 @@ from .session_patterns import (
 )
 from .message_patterns import compile_message_patterns, matches_message_pattern
 from .lifecycle_state import LifecycleStateStore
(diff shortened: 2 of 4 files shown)


stephenschoettler:
Pushed `7edd9af` after rebasing onto current `main`.

Addressed the Codex P2 finding by parsing nested structured text values (`text.value`) and text stored under `content` for text blocks before falling back to normalized JSON.

Validation:
- RED regression failed before the fix: `test_anchored_pattern_matches_structured_text_value_parts`
- `python -m pytest tests/test_lcm_engine.py::TestMessageFiltering -q` -> 18 passed
- `python -m compileall -q message_content.py tests/test_lcm_engine.py`
- `git diff --check`

Tosko4 (changes requested):
This is going in the right direction, but I found a restart/cursor edge that should be fixed before merge.

The new matcher filters live structured content using the operator-visible text, but stored historical rows are already normalized JSON strings. That makes the ignore predicate asymmetric during restart reconciliation. If a DB already contains a structured cron message stored before this change, then after enabling `^Cronjob Response:` the active replay drops that structured message, while the stored tail still keeps the JSON row. Cursor matching fails and the real non-ignored row before it is appended again.

I reproduced it with a temp DB: store `normal A` plus a structured text part containing `Cronjob Response: heartbeat` before the filter exists, then restart with `ignore_message_patterns=['^Cronjob Response:']` and ingest the same active context. The resulting store contents were `normal A`, the old JSON cron row, then `normal A` again.

There is also a docs mismatch: the new README text says structured payloads match concatenated text parts first, but the old “anchored patterns are best-effort against multimodal content” limitation is still below it and now contradicts the intended contract.

`TestMessageFiltering` passes, compile/diff checks pass, but this legacy stored-row/restart case needs coverage and a symmetric matching fix.

Tosko4 (approved):
I pushed `5c472df` to close the restart/cursor edge I had requested changes on.

The issue was the active side of the matcher had moved to text-first structured content, while durable rows from before the filter existed were still stored as canonical JSON. That meant restart reconciliation could filter the live structured cron row but fail to filter the equivalent stored JSON row, then duplicate the non-ignored prefix.

The follow-up keeps replay identity storage-normalized, but decodes stored canonical JSON only for the ignore-pattern predicate during reconciliation. That makes active and stored filtering symmetric without changing storage, FTS, or source identity.

Validation on `5c472df`:

- targeted structured/restart slice: 7 passed
- broad core/engine/command/packaging/importer slice: 500 passed
- full suite: 500 passed

Approved.

PR #127, opened 2026-05-06
fix: prevent ignored sessions from rebinding current LCM session

### LINKED ISSUE #126: Ignored/stateless sessions can rebind current LCM session after cron runs
## Summary

When Hermes runs a cron job in the same gateway process as a conversational session (Telegram, CLI, etc.), the cron session's `on_session_start` rebinds `LCMEngine._session_id` to the cron session id even though the session matches `LCM_IGNORE_SESSION_PATTERNS` (or `LCM_STATELESS_SESSION_PATTERNS`). Every LCM tool that defaults to "the current session" then reports the cron session for the rest of the cron run, even from concurrent foreground tool calls.

Symptom observed by an operator: `lcm_status` returns `session_id=cron_<job-hash>_<timestamp>`, `platform=cron`, `session_filters.ignored=true`, `store.messages=0` while the foreground Telegram DM session has stored rows in `lcm.db` and is the conversation the operator is actively in.

## Why this happens

Hermes-LCM registers a single `LCMEngine` per gateway process via `ctx.register_context_engine(engine)`. Every `AIAgent` in that process resolves the same singleton. Cron schedules a `concurrent.futures.ThreadPoolExecutor` worker per job, constructs an `AIAgent(platform="cron", session_id="cron_...")`, and that constructor calls `context_compressor.on_session_start(cron_session_id, ...)`.

Inside `LCMEngine.on_session_start`, the cron session does not match either of the two existing carve-outs:

- `boundary_reason == "compression"` (compression rollover path) does not apply.
- `_is_live_auxiliary_child_session(...)` returns `False` because cron is constructed without `parent_session_id`, so its `state.db` row has `parent_session_id=NULL` and the in-process call-stack walk does not find an auxiliary agent frame.

Execution falls through to `_apply_session_start_metadata`, which unconditionally sets `self._session_id = session_id` before `_refresh_session_filters()` evaluates the ignore/stateless patterns. By the time `_session_ignored=True` is set, the rebind has already happened. Tool reads of `engine._session_id` (`tools.py:lcm_status`, `lcm_grep` default scope, `_get_session_node`, `_get_externalized_payload`, `_expand_message_sources`, `_serialize_loaded_message`, `lcm_describe`, `lcm_expand`, `lcm_expand_query`, `lcm_doctor`, `command.py:_status_text`) then surface the cron id as the "current session" until the next foreground `on_session_start` rebinds the engine back.

The cron's own lifecycle hooks (`compress`, `should_compress_preflight`, `_ingest_messages`, `handle_tool_call`) correctly short-circuit on `_session_ignored=True`, so cron transcripts do not pollute the foreground store. The bug is purely a "what session does the user-facing tool surface report" issue, not a data-corruption issue. But during a long-running tool-use loop on the foreground side, every concurrent `lcm_status` / `lcm_grep` call sees the cron id, which is confusing at best and breaks any agent tooling that trusts the reported session id.

## Repro

Empirical repro that reaches the bug from the engine surface alone:

```python
from hermes_lcm.config import LCMConfig
from hermes_lcm.engine import LCMEngine

cfg = LCMConfig(
    database_path="/tmp/lcm-repro.db",
    ignore_session_patterns=["cron:*"],
)
engine = LCMEngine(config=cfg)

engine.on_session_start("telegram-fg", platform="telegram", context_length=200_000)
engine._store.append("telegram-fg", {"role": "user", "content": "tg-1"}, token_estimate=1)

engine.on_session_start("cron_xxx", platform="cron", context_length=200_000)

# engine._session_id == "cron_xxx" (rebound)
# engine._session_ignored is True (correctly classified as ignored)
# but lcm_status / lcm_grep / lcm_describe / etc. all read _session_id and surface "cron_xxx"
```

In a real deployment this happens whenever a cron tick fires while a Telegram (or any other foreground) session is processing tool calls in the same gateway process. The LCM database already contains historical evidence: cron session rows from before `LCM_IGNORE_SESSION_PATTERNS=cron:*` was added still have ingested message rows, confirming the rebind has been happening for as long as cron and conversational sessions have shared a process.

## Expected behavior

The user-visible "current session" surface should remain the foreground session for as long as one is bound, even while a side-channel session (cron, debug, or any other session matching the configured ignore/stateless patterns) is being serviced by the engine.

The engine's bound state can keep tracking the side channel because that is what gates the side channel's own `compress` / `_ingest_messages` calls. Only the user-facing view needs to stay stable.

## Affected surfaces

- `lcm_status` (`session_id`, `store.messages`, `dag.*`, `session_filters`).
- `lcm_grep` default scope (`session_scope=current`) and unknown-scope fallback.
- `_get_session_node`, `_get_externalized_payload`, `_expand_message_sources`, `_expand_child_nodes`, `_serialize_loaded_message`.
- `lcm_describe` overview.
- `lcm_expand` (`store_id` mode `from_current_session` flag, `node_id` mode session-scope check).
- `lcm_expand_query` search and node lookup.
- `lcm_doctor` (the diagnostic checks themselves run against whatever session is bound).
- `command.py` `_status_text` and `_doctor_retention_text` (operator `/lcm status` and `/lcm doctor retention`).

## Severity

High for any deployment running cron jobs in the same gateway process as conversational sessions, which is the standard `hermes serve` topology. Not destructive; ignored sessions still skip ingest and compaction. The bug shows up as a binding leak, not data corruption.

## Proposed fix shape

Add a separate `_foreground_session_id` field that tracks the most recent non-ignored, non-stateless binding, plus a `current_session_id` property that returns `_foreground_session_id` (with fallback to `_session_id` when no foreground has ever bound). Keep `on_session_start` rebinding `_session_id` exactly as it does today so cron's own lifecycle hooks continue to short-circuit correctly. Switch tool-surface reads to `current_session_id`.

PR forthcoming.

GodsBoy:
## Summary

- Track the most recent non-ignored, non-stateless binding in a new `_foreground_session_id` field; expose it via `current_session_id` / `current_session_platform` properties on `LCMEngine`.
- Tool-surface code paths (`lcm_status`, `lcm_grep` default scope, `_get_session_node`, `_get_externalized_payload`, `_expand_message_sources`, `_expand_child_nodes`, `_serialize_loaded_message`, `lcm_describe`, `lcm_expand`, `lcm_expand_query`, `lcm_doctor`, `_status_text`, `_doctor_retention_text`) now read `current_session_id` so a cron-style ignored session that briefly rebinds `_session_id` cannot divert lcm_status / lcm_grep / etc. away from the operator's foreground conversation.
- Lifecycle paths inside the engine (`compress`, `should_compress_preflight`, `_ingest_messages`, `handle_tool_call`) continue to read `_session_id` directly, so cron's own calls still short-circuit on `_session_ignored=True` and cannot leak cron transcripts into the foreground store.
- `lcm_status` surfaces the diagnostic `session_filters.side_channel_active` flag (and `side_channel_session_id` when active) so operators can still observe a cron tick that has rebound the engine. `/lcm status` reports the same flag.

## Why

`hermes-lcm` registers a single `LCMEngine` per gateway process. When a cron job runs in the same process as a Telegram (or any other) conversation, cron's `AIAgent.__init__` calls `engine.on_session_start("cron_...")`. That unconditionally sets `self._session_id = cron_xxx` before `_refresh_session_filters` evaluates the ignore patterns, so every LCM tool that defaults to "the current session" reports the cron id from concurrent foreground tool calls (e.g., `lcm_status` returning `session_id=cron_xxx, ignored=true, store.messages=0` while the foreground Telegram session has rows in `lcm.db`). Full diagnosis in #126.

The first iteration of this fix tried to skip the rebind for ignored sessions, but that introduced a worse regression: with `_session_ignored=False` (because the engine stayed bound to the foreground), cron's own `_ingest_messages` no longer short-circuited and cron transcripts leaked into the foreground store. An empirical probe confirmed 4 cron messages writing through to a 1-row foreground session.

The shadow-field approach in this PR avoids the regression by keeping the existing rebind path intact (so cron's lifecycle gating keeps working) while exposing a separate `current_session_id` view for tools. Bound state and foreground state are decoupled.

## Validation

- [x] Focused validation: `pytest tests/test_lcm_engine.py tests/test_lcm_command.py -q` -> `318 passed in 17.85s`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> 449 passed
  - [x] `pytest -q` -> 497 passed
  - [x] `python -m compileall -q .` -> ok
  - [x] `python -m py_compile scripts/import_lossless_claw.py` -> ok
  - [x] `bash -n scripts/install.sh scripts/update.sh` -> ok
  - [x] `git diff --check` -> ok
- [x] Cron-leak regression probe: ran `engine.on_session_start("telegram-fg")` -> append 1 row -> `engine.on_session_start("cron_xxx")` -> `engine.compress(cron_messages)` + `engine.should_compress_preflight(cron_messages)`. Result: telegram store stays at 1 row, cron store at 0 rows. Pre-fix probe (skip-rebind variant) showed 5 telegram rows.

## New tests

- `test_ignored_session_does_not_rebind_foreground_view` -- foreground view stays stable when an ignored session rebinds the bound state.
- `test_ignored_session_compress_does_not_leak_into_foreground_store` -- regression test for the rejected first-iteration approach: cron messages must not leak into the foreground store.
- `test_lcm_status_stays_on_foreground_after_cron_tick` -- end-to-end via the `lcm_status` tool, with `side_channel_active` and `side_channel_session_id` diagnostics.
- `test_lcm_status_reports_bound_session_when_no_foreground_yet` -- cron-only deployment: `current_session_id` falls back to `_session_id` so the surface remains observable.
- `test_stateless_session_does_not_rebind_foreground_view` -- the same protection applies to stateless side-channel sessions.
- `test_lcm_command_status_text_is_consistent_with_lcm_status_during_cron_tick` -- `/lcm status` operator surface stays consistent with `lcm_status` output.
- `test_lcm_doctor_retention_targets_foreground_during_cron_tick` -- `/lcm doctor retention` reports retention for the foreground session, not the cron-bound one.
- `test_foreground_view_advances_when_a_real_foreground_arrives` -- `_foreground_session_id` advances forward through real foreground bindings (e.g., compression rollovers) and is never set to an ignored or stateless id.
- `test_lcm_tool_status_forwards_filter_config_to_agent_surface` updated to expect the new `side_channel_active` key.

## Notes

- Reviewed twice via parallel multi-persona review (correctness, testing, maintainability, project-standards, agent-native, kieran-python, learnings-researcher). The first round caught the cron-leak regression on the rejected approach; the second round on the shadow-field approach surfaced the `_status_text` consistency gap, the operator-protection comments, and the `side_channel_active` encapsulation, all addressed in this PR.
- Out of scope: the deeper concurrency property -- a cron tick that rebinds `_session_id` to an ignored session also pauses foreground compaction during the cron run window, because the foreground thread's `compress` / `_ingest_messages` calls see `_session_ignored=True` until the next foreground `on_session_start`. Fixing that requires per-task session resolution (ContextVars or explicit thread-local push that propagates through `contextvars.copy_context()`), which is an architectural change outside this PR's scope. The window is bounded (cron runs are typically short, and a fresh `AIAgent` per inbound foreground message rebinds the engine back), and no data corrupts.
- `command.py` retention/cleanup sites (`_scan_clean_candidates`, `_delete_clean_candidates_atomically`, the `protected` flag in `_scan_retention_candidates`) intentionally read `_session_id` (bound), with comments explaining the contract: the actively-bound session is the row receiving live writes from the concurrent run, so deletion / cleanup must protect it regardless of the foreground view.

Refs #126

Closes #126

### DIFF
--- a/command.py
+++ b/command.py
@@ -89,7 +89,7 @@ def _status_text(engine) -> str:
     db_path = Path(engine._store.db_path)
     db_exists = db_path.exists()
     db_size = db_path.stat().st_size if db_exists else 0
-    session_bound = bool(engine._session_id)
+    session_bound = bool(engine.current_session_id)
     source_stats = status.get("source_lineage") or {}
     runtime_identity = status.get("runtime_identity") or {}
     source_stats = {
@@ -113,8 +113,8 @@ def _status_text(engine) -> str:
         f"plugin_git_branch: {runtime_identity.get('plugin_git_branch') or '(unavailable)'}",
         f"plugin_git_dirty: {runtime_identity.get('plugin_git_dirty') if runtime_identity.get('plugin_git_dirty') is not None else '(unavailable)'}",
         f"hermes_home: {runtime_identity.get('hermes_home', '') or '(unset)'}",
-        f"session_id: {engine._session_id or '(unbound)'}",
-        f"session_platform: {status.get('session_platform') or ('(unbound)' if not session_bound else '(unknown)')}",
+        f"session_id: {engine.current_session_id or '(unbound)'}",
+        f"session_platform: {engine.current_session_platform or ('(unbound)' if not session_bound else '(unknown)')}",
         f"database_path: {db_path}",
         f"database_path_source: {runtime_identity.get('database_path_source', '(unknown)')}",
         f"database_exists: {_fmt_bool(db_exists)}",
@@ -128,8 +128,13 @@ def _status_text(engine) -> str:
         f"last_cache_write_tokens: {status.get('last_cache_write_tokens', 0)}",
         f"last_reasoning_tokens: {status.get('last_reasoning_tokens', 0)}",
         f"cache_read_ratio: {float(status.get('cache_read_ratio', 0.0) or 0.0) * 100:.1f}%",
-        f"session_ignored: {_fmt_bool(status.get('session_ignored'))}",
-        f"session_stateless: {_fmt_bool(status.get('session_stateless'))}",
+        # Filter classification for current_session_id (the foreground view).
+        # When a side channel is in flight, get_status() reports the bound
+        # session's flags; we read the engine properties instead so this row
+        # stays consistent with the session_id row above.
+        f"session_ignored: {_fmt_bool(engine.current_session_ignored)}",
+        f"session_stateless: {_fmt_bool(engine.current_session_stateless)}",
+        f"side_channel_active: {_fmt_bool(engine.side_channel_active)}",
         f"conversation_id: {runtime_identity.get('conversation_id', '') or '(unbound)'}",
         f"lifecycle_current_session_id: {runtime_identity.get('lifecycle_current_session_id', '') or '(none)'}",
         f"lifecycle_last_finalized_session_id: {runtime_identity.get('lifecycle_last_finalized_session_id', '') or '(none)'}",
@@ -220,6 +225,12 @@ def _scan_clean_candidates(engine) -> dict[str, Any]:
             stateless_count += 1
         if not matched_classes:
             continue
+        # Protect the actively-bound session from cleanup, not the foreground
+        # view. While a cron tick has rebound the engine, _session_id points
+        # at the cron session and the engine is actively writing through it
+        # via lifecycle hooks; deleting that data mid-flight would corrupt
+        # the cleanup pass. current_session_id (foreground) is the wrong
+        # field here.
         if session_id == getattr(engine, "_session_id", ""):
             protected_count += 1
             continue
@@ -245,7 +256,13 @@ def _scan_clean_candidates(engine) -> dict[str, Any]:
 def _scan_retention_candidates(engine) -> dict[str, Any]:
     conn = engine._store._conn
     now = datetime.now().timestamp()
-    session_id = getattr(engine, "_session_id", "")
+    # SQL is scoped to the foreground session so /lcm doctor retention
+    # reports the operator's real conversation rather than whatever side
+    # channel (cron tick, debug probe) currently owns engine._session_id.
+    # The "protected" flag below still keys off engine._session_id (the
+    # actively-bound row) because that is the row receiving live writes
+    # from the concurrent run.
+    session_id = engine.current_session_id
     if not session_id:
         return {
             "error": None,
@@ -340,6 +357,8 @@ def _scan_retention_candidates(engine) -> dict[str, Any]:
         first_activity_at = min(float(ts) for ts in (first_message_at, first_node_at) if ts is not None)
         last_activity_at = max(float(ts) for ts in (last_message_at, last_node_at) if ts is not None)
         age_days = max(0.0, (now - last_activity_at) / 86400.0)
+        # Bound (not foreground): protect the live session from retention
+        # bookkeeping while the engine may still be writing to it.
         protected = session_id == getattr(engine, "_session_id", "")
         total_footprint_tokens = int(token_total) + int(node_token_total)
         if protected:
@@ -927,6 +946,10 @@ def _delete_clean_candidates_atomically(engine, session_ids: set[str]) -> dict[s
     avoid half-cleaned state if a later table delete fails.
     """
     conn = engine._store._conn
+    # Protect the actively-bound session id, not current_session_id. While a
+    # cron tick has rebound the engine, _session_id is the row the engine is
+    # actively writing to via lifecycle hooks; deleting it during cleanup
+    # would race with that ingest.
     protected_session_ids = {getattr(engine, "_session_id", "")}
     protected_session_ids = {str(s) for s in protected_session_ids if s}
     session_ids = {str(s) for s in session_ids if s and str(s) not in protected_session_ids}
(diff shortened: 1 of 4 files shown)


Tosko4 (approved):
I hit one issue during review: the foreground/session split was working for the top-level `lcm_status` fields, but some diagnostic fields still came from the bound cron session, so the output could say “foreground” in one place and “cron” in another.

I pushed `07609b3` to keep `runtime_identity`, lifecycle, source-lineage, `/lcm status`, and store/DAG counts on the same foreground view, while still exposing the bound side-channel session explicitly for diagnostics.

Rechecked the side-channel shape after that:

- ignored/stateless side channels no longer steal the user-facing current session
- cron writes are still gated and do not leak into the foreground store
- `lcm_status`, `/lcm status`, and default `lcm_grep` stay on the foreground session during the side-channel window
- no-foreground cron-only behavior still falls back to the bound session

Validation on the pushed head `07609b3`:

- focused PR #127 regression slice: 9 passed
- `tests/test_lcm_engine.py tests/test_lcm_command.py`: 318 passed
- broad slice including core/engine/command/packaging/importer: 497 passed
- full suite: 497 passed
- extra adversarial probe passed

Approved. This is the right shape for #126: separate the operator-facing foreground view from the bound lifecycle state instead of trying to skip the cron rebind entirely.

PR #132, opened 2026-05-08
fix: preserve latest user objective during compaction

mehalter:
## Summary
- Keep the latest real user message as an explicit active-context anchor when LCM compaction would otherwise omit it from the fresh tail.
- Skip Hermes' synthetic preserved-todo context when choosing that anchor, so the real user request remains dominant.
- Add a regression test for a tool-heavy turn where assistant/tool messages push the current user request outside the normal fresh-tail window.

## Bug
In long gateway sessions, a user can change objectives and then the assistant may run many tool calls before LCM compacts. Because the fresh tail is message-count based, those assistant/tool messages can push the latest user request out of the active context. The next post-compaction prompt can then contain old summarized goals plus mechanical tool traces, but not the current user request verbatim. In practice this can make the agent regress to an older task after a compression/session split.

A user-side workaround is to increase the fresh-tail count, but that is not a rigorous fix: tool-heavy turns can still exceed the configured window, and a larger tail increases context/cost for every session. The engine should preserve the active objective directly.

## Fix
The change is intentionally small:

1. During normal `compress()`, temporarily provide `_assemble_context()` with the original pre-compaction message window.
2. `_assemble_context()` keeps the newest non-synthetic user message if it is not already present in the selected tail.
3. The anchor is inserted after summaries and before the fresh tail, so provider-safe assistant/tool sequencing remains intact.
4. Direct `_assemble_context()` callers and overflow recovery paths keep their existing behavior because anchoring is opt-in during normal compaction only.

## Testing
- `pytest tests/test_lcm_engine.py::TestEngineCompress::test_compress_preserves_latest_user_request_outside_fresh_tail -q`
  - Verified RED before the implementation: failed because the latest user request was absent from compressed active context.
  - Verified GREEN after the implementation: passed.
- `pytest tests/test_lcm_engine.py::TestEngineCompress -q`
  - `18 passed`
- `git diff --check`
- `pytest tests -q`
  - `523 passed, 39 warnings`

_NOTE:_ This was put together in conjunction with gpt-5.5 through hermes-agent mainly as a tool for identifying the problem through log investigations with human guidance for applying the fix. Feel free to leave feedback on the implementation, I naturally do not have deep experience with the codebase but just wanted to use my experiences to help forward this great project! Thanks in advance for the consideration.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -181,6 +181,9 @@ _SYNTHETIC_ASSISTANT_NOISE = {
     "pong",
 }
 
+_PRESERVED_TODO_CONTEXT_PREFIX = "[Your active task list was preserved across context compression]"
+_PRESERVED_OBJECTIVE_CONTEXT_PREFIX = "[Current user objective preserved from compacted history]"
+
 
 def _tool_call_id(tool_call: Any) -> str:
     if not isinstance(tool_call, dict):
@@ -323,6 +326,10 @@ class LCMEngine(ContextEngine):
         self.summary_model = self._config.summary_model
         self._last_overflow_recovery_failed = False
         self._last_condensation_suppressed_reason = ""
+        # Temporary source window used only while compress() assembles context.
+        # _assemble_context also serves tests and recovery paths directly, so
+        # keep anchoring opt-in rather than changing its public behavior.
+        self._pending_context_anchor_messages: Optional[List[Dict[str, Any]]] = None
         self._logged_filter_config = False
         self._pending_reset_session_id: str = ""
         self._pending_reset_conversation_id: str = ""
@@ -735,11 +742,15 @@ class LCMEngine(ContextEngine):
 
         # Step 7: Assemble new active context
         self._refresh_raw_backlog_debt(working_messages)
-        compressed = self._assemble_context(
-            working_messages[0],
-            working_messages[1:],
-            assembly_cap_override=recovery_assembly_cap,
-        )
+        self._pending_context_anchor_messages = messages[1:]
+        try:
+            compressed = self._assemble_context(
+                working_messages[0],
+                working_messages[1:],
+                assembly_cap_override=recovery_assembly_cap,
+            )
+        finally:
+            self._pending_context_anchor_messages = None
         self.compression_count += 1
         if recovery_assembly_cap is None:
             self._last_overflow_recovery_failed = False
@@ -1840,6 +1851,8 @@ class LCMEngine(ContextEngine):
                 "[Note: This conversation uses Lossless Context Management (LCM)." in content
                 and "Earlier turns have been compacted into hierarchical summaries below." in content
             )
+        if content.lstrip().startswith(_PRESERVED_OBJECTIVE_CONTEXT_PREFIX):
+            return True
         if "[Expand for details:" not in content:
             return False
         return bool(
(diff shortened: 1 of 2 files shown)


Tosko4 (changes requested):
I agree with the bug this is trying to fix, but I do not think this implementation is safe to merge yet.

The first blocker is budgeting. `_assemble_context()` selects the tail and summaries under `assembly_cap`, then appends the preserved user anchor afterward. That lets the anchor bypass `max_assembly_tokens` and the overflow recovery cap. I reproduced this with `max_assembly_tokens=120`: compaction returned a 248 token active context and set `_last_overflow_recovery_failed=True` only because the omitted user anchor was added after the budget was already spent. The anchor needs to be part of the budget, or be dropped/truncated when it cannot fit, with a final cap check before returning.

The second blocker is restart ingestion. The inserted anchor is an older real user message pulled forward ahead of the retained tail, while messages between that user request and the retained tail may have been compacted away. After a restart, replay reconciliation tries to match the filtered active context against the durable store suffix. Because the anchor plus tail is no longer a contiguous suffix, reconciliation can fall back to the scaffold cursor and re-ingest the anchor and retained tail. I reproduced this with a small capped context: the store had 8 rows before restart, then after replay plus one new user message it had 14 rows, with the anchored user request and retained tail duplicated.

So I think the shape needs one more design pass. Either the anchor needs explicit replay/scaffold semantics that reconciliation understands, or reconciliation needs to support this non-contiguous anchored context safely. And the cap path needs a regression that proves the preserved anchor cannot make overflow recovery return over budget.

For context, the current targeted test and full suite pass locally on `da7090a`, but they do not exercise either of these invariants. I also ran the full suite without the probes: `523 passed`.

stephenschoettler:
Agreed, this should block merge until it is fixed or we have evidence it is unrelated.

mehalter:
Thanks for the review! Did a revision here based on the feedback. Made sure the preserved objective is now labeled as synthetic scaffolding, the budget remains under the `max_assembly_tokens`, and is covered by the restart-reply/cap regressions. Also made sure the tests cover this as well for regression testing moving forward

Tosko4 (changes requested):
Re-reviewed the current head, `e84fff1`. The two issues from my first review look much closer now: the anchor is charged against the summary budget, and restart replay ignores the preserved-objective scaffold instead of re-ingesting it as a non-contiguous user message.

I still do not think this is safe to merge yet.

The remaining blocker is repeated compaction. After the first compaction, the objective is no longer a real user message in active context. It is synthetic summary scaffolding. On the next compaction, `_latest_user_context_anchor()` only scans `role == "user"` messages, so if the agent keeps running tools and compacts again before the user speaks, the preserved objective can be summarized away and no longer appears verbatim in active context.

I reproduced this on `e84fff1`: first compaction returned the expected `[Current user objective preserved from compacted history]` block with `LATEST OBJECTIVE: increase autonomy`; second compaction, after only assistant/tool messages, returned `second_compaction_has_latest=False`.

That is the same user-facing failure, just one compaction later. I think the fix needs to either carry forward an existing preserved-objective scaffold as the next anchor source, or reconstruct the latest objective from durable raw history instead of only scanning current active `role=user` messages. It also needs a regression for repeated compaction with no new user turn.

Validation on current head:

```text
python -m pytest -q tests/test_lcm_engine.py
306 passed in 3.64s

python -m pytest -q
525 passed in 16.96s

manual repeated-compaction probe
second_compaction_has_latest=False
```

Keeping changes requested until that repeated-compaction invariant is covered.

stephenschoettler:
Agreed with the latest review. This should block merge until it is fixed or we have evidence it is unrelated.

Tosko4 (approved):
Re-reviewed the rebased branch on `a30ebae`. The repeated-compaction blocker from my last review is covered now: an existing preserved-objective scaffold is carried forward when no newer real user turn exists, so the active objective does not disappear one compaction later.

I also resolved the latest `main` conflicts, including the #146 search fix, and reran the guardrails that mattered for this PR. The anchor remains budgeted, restart replay still ignores the scaffold, and CI is green across 3.11, 3.12, and 3.13.

Approving from my side.

PR #136, opened 2026-05-09
fix: handle multimodal first message assembly

### LINKED ISSUE #131: LCM context assembly crashes when first Hermes message is multimodal user content
## Bug Description

`hermes-lcm` can crash during context assembly for Hermes gateway sessions whose first persisted conversation message is multimodal/structured content.

Observed gateway/user-facing error:

```text
Sorry, I encountered an error (TypeError).
can only concatenate list (not "str") to list
```

Stack trace points at `_assemble_context()` appending the LCM note:

```text
run_agent.py -> _compress_context(...)
plugins/hermes-lcm/engine.py -> compress(...)
plugins/hermes-lcm/engine.py -> _assemble_context(...)
TypeError: can only concatenate list (not "str") to list
```

Relevant code on current `origin/main` still appears to do a raw `content + str` append:

```python
# engine.py:_assemble_context
sys_msg = system_msg.copy()
if self.compression_count == 0 and include_lcm_note:
    sys_content = sys_msg.get("content", "")
    sys_msg["content"] = (
        sys_content
        + "\n\n[Note: This conversation uses Lossless Context Management (LCM). "
        "Earlier turns have been compacted into hierarchical summaries below. "
        "Use lcm_grep to search history, lcm_describe to inspect the DAG, "
        "and lcm_expand to recover original details from any summary.]"
    )
```

## Why this happens

In the failing Hermes Discord thread, `messages[0]` was **not** a system prompt. It was the first user message, and because Discord routed attached images natively to a vision-capable model, its content was OpenAI-style structured/multimodal content:

```python
{
    "role": "user",
    "content": [
        {"type": "text", "text": "..."},
        {"type": "image_url", "image_url": {"url": "..."}},
        {"type": "image_url", "image_url": {"url": "..."}},
    ],
}
```

LCM appears to assume `working_messages[0]` / `system_msg` is always a string-content system message:

```python
# engine.py:compress
# Protect system prompt (always index 0)
candidate_raw = working_messages[1:fresh_tail_start]
...
compressed = self._assemble_context(
    working_messages[0],
    working_messages[1:],
    ...
)
```

But Hermes core passes the conversation `messages` list to context engines directly, without prepending the system prompt to that list:

```python
# run_agent.py:_compress_context
compressed = self.context_compressor.compress(
    messages,
    current_tokens=approx_tokens,
    focus_topic=focus_topic,
)
```

So for gateway sessions, index 0 can be the first user message. If that message is text-only, the bug is partially masked because `str + str` succeeds, although the LCM note is still attached to a user message. If it is multimodal, Python raises `list + str`.

## Steps to Reproduce

1. Use Hermes Agent with `hermes-lcm` enabled and selected as the context engine.
2. Start a new Discord/gateway thread with a first user message containing text plus image attachments, such that the persisted first message has `content` as a list of content parts.
3. Continue the session until context compression triggers.
4. Observe `_assemble_context()` treating the first user message as `system_msg` and attempting `list + str`.

A minimal unit-style repro should be possible by calling `LCMEngine.compress()` / `_assemble_context()` with a first message shaped like:

```python
messages = [
    {
        "role": "user",
        "content": [
            {"type": "text", "text": "hello"},
            {"type": "image_url", "image_url": {"url": "data:image/png;base64,..."}},
        ],
    },
    # enough additional messages to trigger compaction/assembly...
]
```

## Expected Behavior

LCM should not assume the active conversation list always starts with a system message.

Possible acceptable behavior:

- If `messages[0].get("role") == "system"`, append the LCM note using a structured-content-safe append helper.
- If there is no system message in the active message list, either:
  - insert a separate valid system/developer/user scaffold message for the LCM note, or
  - omit the note, or
  - otherwise handle the no-system case explicitly.
- Structured/multimodal content should never be concatenated directly with a string.

## Actual Behavior

LCM assumes index 0 is a system prompt and directly concatenates its `content` with a string. When `content` is a list, compression crashes and the Hermes gateway returns an error to the user.

## Environment

Observed locally:

- Hermes Agent: `v0.12.0 (2026.4.30)`
- Runtime: macOS launchd gateway, Discord thread session
- Installed `hermes-lcm`: `0.8.0` at commit `0f60d72`
- Local LCM was behind `origin/main` by 20 commits at investigation time
- Checked current LCM `origin/main` and the same raw-concat `_assemble_context()` code path still appears present

Related but distinct prior fix:

- #77 fixed SQLite persistence of list-form structured content (`Error binding parameter ... type 'list' is not supported`). This issue is a later assembly-time failure: `list + str` in `_assemble_context()`.

## Proposed Fix

1. Add a regression test for compression/assembly with no system message and a multimodal first user message.
2. Make `_assemble_context()` role-aware:

```python
sys_msg = system_msg.copy()
if sys_msg.get("role") == "system":
    # append LCM note safely
else:
    # do not treat it as a system prompt
```

3. Use `normalize_content_value()` or a structured-content-safe append helper before appending text.
4. Consider clarifying the context-engine contract with Hermes core: either context engines receive the system prompt as `messages[0]`, or they must handle message lists that start with user/assistant/tool messages.

Tosko4:
## Summary

Fixes context assembly when Hermes gateway sessions do not include a leading system prompt in the message list.

The change keeps the existing leading-message anchor behavior, but only appends the LCM note when that leading message is actually `role: system`. If the first message is a multimodal user message, LCM leaves it unchanged instead of treating it like a system prompt. Structured system content is also handled safely by appending the note as a text content part rather than concatenating a string onto a list.

Fixes #131

## Why

Hermes core can pass only conversation messages to context engines. In gateway sessions, that means `messages[0]` can be the first user message. When that first user message contains image parts, `_assemble_context()` was doing `list + str` while adding the LCM note, which raised `TypeError` during compression.

## Validation

- [x] Focused validation: `pytest -q tests/test_lcm_engine.py::TestEngineCompress::test_compress_handles_multimodal_first_user_message_without_system tests/test_lcm_engine.py::TestEngineCompress::test_assemble_context_appends_lcm_note_to_structured_system_content tests/test_lcm_engine.py::TestEngineCompress::test_compress_preserves_system_and_tail tests/test_lcm_engine.py::TestEngineCompress::test_compress_leaf_node_tracks_source_ids_for_content_part_messages` -> `4 passed`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `478 passed`
  - [x] `pytest -q` -> `524 passed`
  - [x] `git diff --check`

## Notes

No new configuration. The no-system case omits the LCM note rather than injecting a synthetic system message, because that keeps the returned message shape closest to what Hermes supplied.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -631,11 +631,14 @@ class LCMEngine(ContextEngine):
             n = len(working_messages)
             fresh_tail_start = max(0, n - self._config.fresh_tail_count)
 
-            # Protect system prompt (always index 0)
+            # Keep the leading message anchored. It is usually the system
+            # prompt, but Hermes gateway sessions may pass only conversation
+            # messages here, so index 0 can be a user message.
             if fresh_tail_start <= 1:
                 break
 
-            # Skip system prompt (index 0), candidate raw backlog is indices 1..fresh_tail_start
+            # Skip the leading anchor, candidate raw backlog is indices
+            # 1..fresh_tail_start.
             candidate_raw = working_messages[1:fresh_tail_start]
             if not candidate_raw:
                 break
@@ -2361,6 +2364,24 @@ class LCMEngine(ContextEngine):
 
     # -- Internal: context assembly ----------------------------------------
 
+    @staticmethod
+    def _append_lcm_note_to_content(content: Any) -> Any:
+        note = (
+            "\n\n[Note: This conversation uses Lossless Context Management (LCM). "
+            "Earlier turns have been compacted into hierarchical summaries below. "
+            "Use lcm_grep to search history, lcm_describe to inspect the DAG, "
+            "and lcm_expand to recover original details from any summary.]"
+        )
+        if isinstance(content, str):
+            return content + note
+        note_part = {"type": "text", "text": note.lstrip()}
+        if content is None:
+            return note.lstrip()
+        if isinstance(content, list):
+            return list(content) + [note_part]
+        normalized = normalize_content_value(content) or ""
+        return normalized + note
+
     def _assemble_context(
         self,
         system_msg: Dict[str, Any],
@@ -2371,24 +2392,26 @@ class LCMEngine(ContextEngine):
         """Build the active context from DAG summaries + fresh tail.
 
         Structure:
-          [system prompt (with LCM note)]
+          [leading anchor, normally system prompt]
           [highest-depth summary nodes first, then lower]
           [fresh tail messages]
         """
         result = []
 
-        # System prompt with LCM annotation
-        sys_msg = system_msg.copy()
-        if self.compression_count == 0 and include_lcm_note:
-            sys_content = sys_msg.get("content", "")
-            sys_msg["content"] = (
-                sys_content
-                + "\n\n[Note: This conversation uses Lossless Context Management (LCM). "
-                "Earlier turns have been compacted into hierarchical summaries below. "
-                "Use lcm_grep to search history, lcm_describe to inspect the DAG, "
-                "and lcm_expand to recover original details from any summary.]"
+        # Leading anchor with optional LCM annotation. Only append the note to
+        # an actual system message; gateway sessions can start with a structured
+        # multimodal user message, and that content must not be treated as a
+        # system prompt or concatenated with text.
+        leading_msg = system_msg.copy()
+        if (
+            leading_msg.get("role") == "system"
+            and self.compression_count == 0
+            and include_lcm_note
+        ):
+            leading_msg["content"] = self._append_lcm_note_to_content(
+                leading_msg.get("content", "")
             )
-        result.append(sys_msg)
+        result.append(leading_msg)
 
         assembly_cap = (
             assembly_cap_override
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Reviewed and approved.

Validation I checked locally on `443948d`:

```text
python -m pytest -q tests/test_lcm_engine.py::TestEngineCompress::test_compress_handles_multimodal_first_user_message_without_system tests/test_lcm_engine.py::TestEngineCompress::test_assemble_context_appends_lcm_note_to_structured_system_content tests/test_lcm_engine.py::TestEngineCompress::test_compress_preserves_system_and_tail tests/test_lcm_engine.py::TestEngineCompress::test_compress_leaf_node_tracks_source_ids_for_content_part_messages
4 passed

python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q
478 passed

python -m pytest -q -o addopts=
524 passed

python -m compileall -q .
git diff --check
all passed
```

The first full-suite run hit a transient local `sqlite3.OperationalError: unable to open database file`; the failing test passed on direct rerun, and the full suite passed after rerun with a raised fd limit.

This fixes the no-system multimodal first-message path without mutating the first user message, and structured system content gets the LCM note as a text part.

PR #143, opened 2026-05-09
chore: release v0.9.3

### LINKED ISSUE #108: Feature: push lcm_grep filter set into the search layer
## Summary

Add `role`, `time_from`, `time_to` (and tighten existing `source`) filters to `lcm_grep` so they push into the FTS / data-layer query rather than running as Python post-filters over a bounded candidate fetch. The previous attempt to add these filters in PR #102 was reverted because the post-fetch shape silently lost valid results when matches sat outside the first candidate window. This issue tracks the proper implementation.

## Why

PR #102 added `role`, `time_from`, `time_to` as Python post-filters applied to the result list returned by `MessageStore.search` / `SummaryDAG.search`. Maintainer review (#102) reproduced the failure:

> `lcm_grep` fetches only `source_limit = max(limit * 4, limit, 20)` from the store/DAG, then applies role/time filters in Python. That means valid filtered matches outside the first candidate window disappear.

Concretely, with one older matching `assistant` row and 25 newer matching `user` rows filling the candidate window, `role: assistant` returned `total_results == 0` instead of the older assistant row. The same shape reproduces with `time_to` and any other post-filter that excludes most of the early candidates.

The filters were removed from PR #102 entirely rather than fixed inline, since pushing them into the data layer is a meaningfully larger change.

## Scope

- Push `role`, `time_from`, `time_to` into the FTS / WHERE-clause layer of `MessageStore.search` (and the equivalent paths in `SummaryDAG.search` if/when summary cross-session expansion lands).
- Either: (a) accept the filters as SQL predicates so the FTS engine returns post-filter candidates directly, or (b) keep the post-filter shape but extend the adaptive-widening loop to keep paging until enough post-filter matches collect, bounded by a hard candidate cap.
- Surface the filters back on `lcm_grep` (and update the schema) once the data-layer contract is solid.
- Naive ISO 8601 inputs (no `Z`, no offset) must be rejected so cross-machine reproducibility holds.

## Test contract

The maintainer's exact reproduction is the regression bar:

```python
# one older matching assistant row
assistant_id = engine._store.append(\"test-session\", {\"role\": \"assistant\", \"content\": \"docker target assistant\"})
engine._store._conn.execute(\"UPDATE messages SET timestamp=? WHERE store_id=?\", (1000.0, assistant_id))

# 25 newer matching user rows fill the pre-filter candidate window
for i in range(25):
    sid = engine._store.append(\"test-session\", {\"role\": \"user\", \"content\": f\"docker noisy user {i}\"})
    engine._store._conn.execute(\"UPDATE messages SET timestamp=? WHERE store_id=?\", (2000.0 + i, sid))
engine._store._conn.commit()

result = json.loads(engine.handle_tool_call(\"lcm_grep\", {\"query\": \"docker\", \"role\": \"assistant\", \"limit\": 1, \"sort\": \"recency\"}))
```

Required: \`result[\"total_results\"] >= 1\` and the assistant row appears in \`result[\"results\"]\`. The same shape with \`time_to\` set to a value that excludes the 25 newer rows must return the older assistant row.

## Related

- PR #102 — the cross-session search PR that originally introduced and then removed these filters
- Maintainer review on #102 (CHANGES_REQUESTED, commit \`19d05f8\`) — the source of Bug 2 and the test contract above

### LINKED ISSUE #110: Bug: lcm_expand returns empty sources for carried-over summary nodes
## Summary

After `/new` or session rollover carries retained summary nodes into the new session, `lcm_grep` can surface those carried-over summaries as current-session hits, but `lcm_expand(node_id=...)` returns an empty `expanded` list for the same node.

## Reproduction

1. Create an old-session raw row with source `discord`.
2. Create a retained summary node in `old-session` with `source_ids` pointing to that raw row and `source_type="messages"`.
3. Run `carry_over_new_session_context("old-session", "new-session")` with `LCM_NEW_SESSION_RETAIN_DEPTH=2`.
4. Set the engine session to `new-session`.
5. Run `lcm_grep` for the carried-over summary. It returns the summary under `new-session`.
6. Run `lcm_expand` on that returned `node_id`.

## Expected behavior

`lcm_expand(node_id=...)` should recover the raw message sources behind the carried-over summary, or it should return an explicit contract error explaining why carried-over summary sources are not expandable.

## Actual behavior

`lcm_expand(node_id=...)` returns `expanded: []` even though the node has `source_ids` and `pagination.total_sources == 1`.

Observed repro result on latest `main` (`6b0269b`) and PR #102 head (`c837e9b`):

```json
{
  "grep_types": ["summary"],
  "grep_sessions": ["new-session"],
  "expand_count": 0,
  "expand": {
    "source_type": "messages",
    "expanded": [],
    "pagination": {
      "total_sources": 1,
      "returned_sources": 0,
      "has_more": false
    }
  }
}
```

## Likely cause

`_expand_message_sources()` drops rows whose stored `session_id` does not match `engine._session_id`. Carry-over reassigns the summary node to the new session while preserving descendant raw-message source lineage in the original session, so the expansion helper filters out the exact rows it needs.

## Notes

This appears pre-existing on `main`, not introduced by PR #102. PR #102 only made the issue easier to notice while re-reviewing retrieval contracts.

stephenschoettler:
## Summary

- Bumps `hermes-lcm` from 0.9.2 to 0.9.3.
- Prepares the tag-driven release `v0.9.3`.
- Includes the retrieval/rollover fixes from #135 directly in this release branch, pending final review on the standalone PR.

Fixes #108
Fixes #110

## Changes since v0.9.2

- fix: restore filtered grep and carried source expansion (#135)
- fix: handle multimodal first message assembly (#136)
- fix: handle interrupt during session-end cleanup (#138)

## Version files

- `plugin.yaml`
- `README.md`
- `tests/test_lcm_command.py`
- `tests/test_lcm_engine.py`
- `tests/test_packaging_install.py`
- `.github/ISSUE_TEMPLATE/bug_report.yml`

## Validation

- `python -m pytest tests/test_lcm_command.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=`: `363 passed`
- `python -m pytest tests -q -o addopts=`: `537 passed`
- `python -m compileall -q .`
- `python -m py_compile scripts/import_lossless_claw.py`

## Release path

After this PR lands on `main` and CI is green, the release watcher will push annotated tag `v0.9.3`. The existing release workflow will create the GitHub release from that tag.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -44,7 +44,7 @@ body:
     attributes:
       label: Version and branch
       description: hermes-lcm version, branch, commit, or release tag.
-      placeholder: v0.9.2, main at d6d4e15, or commit SHA
+      placeholder: v0.9.3, main at 3a2a7e7, or commit SHA
     validations:
       required: true
 
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ for earlier separate sessions or broad cross-session history.
 
 | Tool | Use |
 |------|-----|
-| `lcm_grep` | Search current-session raw messages and summaries. Opt into `session_scope='all'` or `session_scope='session'` (with `session_id`) for bounded archive recovery over rows already present in `lcm.db`, including externally backfilled rows that may carry source strings such as `openclaw-lcm:*`; broader scopes return raw-message hits only. Use `session_search` for earlier separate sessions or broad cross-session recall. |
+| `lcm_grep` | Search current-session raw messages and summaries. Opt into `session_scope='all'` or `session_scope='session'` (with `session_id`) for bounded archive recovery over rows already present in `lcm.db`, including externally backfilled rows that may carry source strings such as `openclaw-lcm:*`; broader scopes return raw-message hits only. Raw-message filters `role`, `time_from`, and `time_to` are pushed into the search query; when any of them is supplied, summary hits are omitted so the filter contract stays exact. Use `session_search` for earlier separate sessions or broad cross-session recall. |
 | `lcm_load_session` | Load one ordered raw-message transcript page for an explicit `session_id`. This is not search: it returns raw rows in `store_id` order, bounded by `limit`, with per-message content bounded by `max_content_chars`, and continues with `after_store_id` from `next_cursor`. |
 | `lcm_describe` | Inspect the current-session DAG or preview an `externalized_ref` without loading full content. |
 | `lcm_expand` | Recover source messages, child summaries, or externalized payloads with pagination. Use `store_id` to fetch a single raw message regardless of session, suitable for drilling into a cross-session `lcm_grep` result. |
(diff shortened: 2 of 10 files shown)


Tosko4 (approved):
Reviewed after adding the #135 retrieval/rollover fixes into the release branch.

Verified locally:

- `python -m pytest tests/test_lcm_command.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=`: `363 passed`
- `python -m pytest tests -q -o addopts=`: `537 passed`
- compile/shell syntax checks and `git diff --check`

CI is green on 3.11, 3.12, and 3.13.

stephenschoettler:

PR #159, opened 2026-05-10
fix: refresh plugin metadata from manifest on status reads

WeirdOldman:
Summary
- refresh hermes-lcm plugin metadata from plugin.yaml on status reads
- make runtime status tools reflect hot-updated plugin checkouts in long-lived Hermes processes
- retain cached metadata only as a fallback when the manifest is not readable

Why
lcm_status.runtime_identity.plugin_version could stay stale after a plugin repo hot-update because _plugin_metadata() returned cached values once initialized. Refreshing metadata from plugin.yaml keeps runtime status aligned with the actual checked-out plugin version.

Validation
- restarted the gateway and confirmed runtime_identity.plugin_version updated to 0.10.2
- confirmed hermes plugins list shows hermes-lcm 0.10.2
- ran:
  `PYTHONPATH=/root/.hermes/hermes-agent python3 -m pytest tests/test_lcm_engine.py -q -o addopts=`
- result: 360 passed

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -84,10 +84,12 @@ def _strip_metadata_scalar(value: str) -> str:
 
 
 def _plugin_metadata() -> dict[str, str]:
-    """Return plugin identity from the loaded code tree."""
+    """Return plugin identity from the loaded code tree.
+
+    Always re-read the manifest from disk when available so status tools reflect
+    hot-updated plugin checkouts even in long-lived Hermes processes.
+    """
     global _PLUGIN_METADATA
-    if _PLUGIN_METADATA is not None:
-        return dict(_PLUGIN_METADATA)
 
     metadata = {"name": "hermes-lcm", "version": "unknown"}
     manifest = _PLUGIN_ROOT / "plugin.yaml"
@@ -99,10 +101,13 @@ def _plugin_metadata() -> dict[str, str]:
             key = key.strip()
             if key in {"name", "version"}:
                 metadata[key] = _strip_metadata_scalar(raw_value)
+        _PLUGIN_METADATA = metadata
+        return dict(metadata)
     except OSError:
         logger.debug("LCM plugin manifest not readable at %s", manifest)
 
-    _PLUGIN_METADATA = metadata
+    if _PLUGIN_METADATA is not None:
+        return dict(_PLUGIN_METADATA)
     return dict(metadata)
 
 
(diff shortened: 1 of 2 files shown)


stephenschoettler:

Please add:
- Summary
- Why
- Validation

stephenschoettler:
Thanks. CI is green on `a2bcb00`, and I've asked @Tosko4 for review. If @Tosko4 approves and checks stay green, I'll merge this.

Tosko4 (dismissed):
I think the implementation is the right shape, but I don’t want this to merge without a regression test. The current suite would still pass with the old cached implementation, because it only checks the static manifest value.

Please add a small focused test for the runtime identity path. It should prove that a later status read in the same process sees a changed `plugin.yaml`, and that a missing or unreadable manifest falls back to the last good metadata. The first read with no manifest should still return `hermes-lcm` and `unknown`.

I checked the code path locally and did not find a correctness issue. This is a coverage blocker for an operator-facing status invariant.

```text
PYTHONPATH=/home/nabu/.hermes/hermes-agent python -m pytest tests/test_packaging_install.py tests/test_lcm_command.py -q -o addopts=  # 45 passed
PYTHONPATH=/home/nabu/.hermes/hermes-agent python -m pytest tests/test_lcm_engine.py -q -o addopts=  # 360 passed
PYTHONPATH=/home/nabu/.hermes/hermes-agent python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=  # 540 passed
PYTHONPATH=/home/nabu/.hermes/hermes-agent python -m pytest -q -o addopts=  # 587 passed
bash -lc 'ulimit -n 1024 && PYTHONPATH=/home/nabu/.hermes/hermes-agent python -m pytest -q -o addopts='  # 587 passed
python -m compileall -q .
```

stephenschoettler:
I agree with Tosko4's review. I am holding merge until this is fixed or there is clear evidence the concern does not apply here.

stephenschoettler (approved):
Re-reviewed current head a0c57e after updating it onto current main.

Tosko4's blocker is addressed now: the PR includes focused coverage proving same-process manifest refresh, last-good fallback after a manifest disappears, and first-read fallback when no manifest exists.

Local validation:
- `git diff --check origin/main...origin/pr-159-review`: clean
- `python -m pytest tests/test_lcm_engine.py -q -o addopts= -k 'plugin_metadata or runtime_identity or lcm_doctor_json_includes_runtime_identity or get_status_exposes_runtime_identity'`: 5 passed
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=`: 548 passed

CI is green. Approved.

PR #163, opened 2026-05-11
fix: route custom LCM models and report no-ops

stephenschoettler:
## Summary
- Fix LCM model override parsing so `custom:<named-provider>/<model>` works for resolvable named custom providers.
- Preserve canonical built-in provider safeguards when splitting provider/model overrides.
- Track compression status/no-op reasons and surface them through `/lcm status` and the `lcm_status` tool.
- Prevent sanitized active-context cleanup from being misreported as a no-op.

## Why
- Local llama.cpp deployments can hot-swap models by the `model` field, but LCM was not routing `custom:LCPP/...` overrides the same way Hermes auxiliary compression does.
- Manual and automatic compression no-ops were hard to diagnose because the public status surfaces did not explain why compression skipped work.
- Content-only sanitization can change active context without changing message count, so length-only no-op detection was wrong.

## Validation
- [x] Focused validation: `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py tests/test_lcm_command.py -q -o addopts=''` -> `582 passed, 40 warnings`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> covered by focused superset above
  - [x] `pytest -q` -> `592 passed, 40 warnings`
  - [x] `ulimit -n 1024; python -m pytest -q -o addopts=''` -> `592 passed, 40 warnings`

## Notes
- No workflow files changed, so `actionlint` was not run.
- Also ran an added-line static scan for obvious secret, shell, eval/exec, pickle, and SQL injection patterns: passed.
- Independent review: approved.

Refs #

### DIFF
--- a/command.py
+++ b/command.py
@@ -120,6 +120,8 @@ def _status_text(engine) -> str:
         f"database_exists: {_fmt_bool(db_exists)}",
         f"database_size: {_fmt_size(db_size) if db_exists else 'missing'}",
         f"compression_count: {engine.compression_count}",
+        f"last_compression_status: {status.get('last_compression_status', 'idle')}",
+        f"last_compression_noop_reason: {status.get('last_compression_noop_reason', '') or '(none)'}",
         f"threshold_tokens: {engine.threshold_tokens if session_bound else '(uninitialized)'}",
         f"cache_metrics_available: {_fmt_bool(status.get('cache_metrics_available'))}",
         f"last_input_tokens: {status.get('last_input_tokens', 0)}",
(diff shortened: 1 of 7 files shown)


Tosko4 (approved):
Reviewed current head `9a5ce90` against latest main. This looks good to me.

The custom model routing keeps the important guardrails intact: named custom providers like `custom:LCPP/...` split into an explicit provider plus stripped model, while canonical built-ins and OpenRouter namespace slugs stay model-only. The new compression status fields also line up with the actual outcomes I checked, including no-op, sanitized active context cleanup, compacted, and forced-overflow paths.

Validation:

```text
python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py tests/test_lcm_command.py -q -o addopts=''
589 passed

python -m pytest -q
657 passed

python -m compileall -q .
```

I also ran a read-only blocker audit focused on the model routing and no-op/status paths. No blockers found.

PR #167, opened 2026-05-13
fix: auto-derive conversation_id when gateway omits or mirrors session_id

### LINKED ISSUE #133: Feature: auto-link sessions into conversations when gateway doesn't pass conversation_id
## Problem

When the host/gateway starts an LCM session without a stable `conversation_id`, LCM binds that session as its own conversation. Across `/new`, restarts, or session rotation, related sessions can become separate lifecycle islands.

That breaks the intended behavior for:

- lifecycle rollover
- retained summary carry-over
- bounded cross-session search within the same real conversation

## Current understanding

LCM lifecycle binding uses the `conversation_id` passed to `on_session_start(...)`. If that value is missing, LCM falls back to `session_id`.

#167 was a prior attempt, but it was closed unmerged after review showed the derivation happened too late and only affected status/display output. It should be treated as background, not as the current implementation path.

## Desired contract

The host should pass a stable conversation identity into LCM at session start:

```python
on_session_start(session_id, conversation_id=<stable host conversation id>, ...)
```

That identity should come from host/gateway routing context rather than LCM guessing after lifecycle binding.

The id must also be scoped to sub-conversations where the platform has them:

- Telegram simple chat or DM: same chat stays in the same LCM conversation
- Telegram forum topic/thread: each topic/thread is isolated
- Discord channel/thread: thread scope is preserved according to the gateway session identity
- no thread/topic: existing chat/channel-level behavior is unchanged

## Related tracking

- Thread/topic acceptance criterion: #171
- Closed prior attempt: #167, not merged

ether-btc:
When Hermes gateway does not pass a stable conversation_id (or sets it equal to session_id), cross-session search and session rollover break.

## Problem
Issue #133: gateway occasionally omits `conversation_id` from runtime_identity or sets it equal to `session_id`. This breaks:
- `rollover_session()` — cannot link sessions across restarts
- Cross-session search — queries return no results because they're in separate 'islands'

## Fix
Auto-derives a platform-specific conversation_id from runtime identity metadata:

- **Discord**: `discord:{guild_id}:{channel_id}:{user_id}`
- **Telegram**: `telegram:{chat_id}`
- **Other**: `{platform}:{session_id}`

The fix triggers when `conversation_id` is `None`, empty string, or equals `engine.current_session_id`.

## Changes
- `command.py`: Added derivation logic in `_status_text()` runtime identity block
- 18 lines added, 0 removed
- All tests pass, syntax verified

Fixes: https://github.com/stephenschoettler/hermes-lcm/issues/133

### DIFF
--- a/command.py
+++ b/command.py
@@ -105,6 +105,24 @@ def _status_text(engine) -> str:
     session_bound = bool(engine.current_session_id)
     source_stats = status.get("source_lineage") or {}
     runtime_identity = status.get("runtime_identity") or {}
+
+    # Auto-derive conversation_id when gateway omits it or mirrors session_id
+    # Fixes https://github.com/stephenschoettler/hermes-lcm/issues/133
+    current_conv_id = runtime_identity.get('conversation_id')
+    if current_conv_id is None or current_conv_id == '' or current_conv_id == engine.current_session_id:
+        platform = engine.current_session_platform or 'unknown'
+        if platform == 'discord':
+            guild_id = runtime_identity.get('guild_id', 'unknown')
+            channel_id = runtime_identity.get('channel_id', 'unknown')
+            user_id = runtime_identity.get('user_id', 'unknown')
+            current_conv_id = f"discord:{guild_id}:{channel_id}:{user_id}"
+        elif platform == 'telegram':
+            chat_id = runtime_identity.get('chat_id', 'unknown')
+            current_conv_id = f"telegram:{chat_id}"
+        else:
+            current_conv_id = f"{platform}:{engine.current_session_id}"
+        runtime_identity['conversation_id'] = current_conv_id
+
     source_stats = {
         "messages_total": int(source_stats.get("messages_total", 0) or 0),
         "attributed_messages": int(source_stats.get("attributed_messages", 0) or 0),


stephenschoettler:
Thanks. CI is green on `daf64a6`, and I've asked @Tosko4 for review. If @Tosko4 approves and checks stay green, I'll merge this.

Tosko4 (changes requested):
I do not think this fixes #133 yet.

The derivation is happening in `command.py` inside `_status_text()`, after `engine.get_status()` has already read the engine and lifecycle state. That only changes the local dict used to render `/lcm status`; it never feeds back into `on_session_start()`, `_bind_lifecycle_state()`, or `lcm_lifecycle_state`. In the broken path, `on_session_start()` still calls `_bind_lifecycle_state(session_id, conversation_id=kwargs.get("conversation_id"))`, and `bind_session()` still falls back to `session_id` when that value is absent.

There is a second issue too: the platform IDs used here are not present in `runtime_identity`. `get_runtime_identity()` does not expose `chat_id`, `guild_id`, `channel_id`, or `user_id`, so a Telegram repro ends up displaying `conversation_id: telegram:unknown` while the lifecycle table still has separate rows for `session-a` and `session-b`.

Minimal repro I used locally:

```python
engine.on_session_start("session-a", platform="telegram", chat_id="12345")
engine.on_session_end("session-a", [{"role": "user", "content": "first"}])
engine.on_session_start("session-b", platform="telegram", chat_id="12345")
```

Observed lifecycle rows after that are still two islands:

```text
session-a, current_session_id=None, last_finalized_session_id=session-a
session-b, current_session_id=session-b, last_finalized_session_id=None
```

Calling `/lcm status` after that displays `telegram:unknown`, but the DB rows do not change.

I think the fix needs to normalize or derive the fallback `conversation_id` during session start, before lifecycle binding, and only from explicit host metadata that LCM actually receives. `_status_text()` should display persisted engine state, not invent lifecycle identity.

```text
python -m pytest tests/test_lcm_command.py tests/test_lcm_engine.py -q
406 passed

python -m pytest -q
657 passed
```

ether-btc:
Thanks for the thorough review -- you're absolutely right on both points. I did some more analysis and here's what I found:

**Point 1: Derivation is display-only, never feeds lifecycle binding**

Confirmed. The derivation in `_status_text()` only mutates the `runtime_identity` dict used to render `/lcm status` output. By that point, `on_session_start()` has already called `_bind_lifecycle_state(session_id, conversation_id=kwargs.get("conversation_id"))`, and when `conversation_id` is None, `bind_session()` in `lifecycle_state.py:126` falls back to `session_id`. The lifecycle DB rows are already split into separate conversations before `_status_text()` ever runs.

The minimal repro confirms this: two sessions from the same Telegram chat still produce two `lcm_lifecycle_state` rows because the binding happens too early for any display-time derivation to matter.

**Point 2: Platform IDs don't exist in `runtime_identity`**

Confirmed. I audited `get_runtime_identity()` (engine.py:1774-1804) -- it returns `session_id`, `session_platform`, `conversation_id`, plugin metadata, and git info. It never exposes `chat_id`, `guild_id`, `channel_id`, or `user_id`. Those fields are never passed from the gateway into the LCM engine at all. So `runtime_identity.get('chat_id', 'unknown')` always evaluates to `'unknown'`, producing `conversation_id: telegram:unknown`.

**Root cause analysis**

This fix belongs in the Hermes gateway, not in LCM. The gateway is where `chat_id`, `guild_id`, `channel_id`, and `user_id` are actually available. The correct fix is: gateway computes `conversation_id = f"telegram:{chat_id}"` or `f"discord:{guild_id}:{channel_id}:{user_id}"` and passes it to `engine.on_session_start(session_id, conversation_id=derived_id)`. That way `_bind_lifecycle_state()` receives a valid `conversation_id` and the lifecycle DB stays unified.

An LCM-side mitigation could be done later if gateway changes are slow -- but it would need to intercept at `on_session_start()` before lifecycle binding, not in the display layer. And with only `session_id` and `platform` as inputs (no user/channel IDs), the best an LCM-side derivation could do is `f"{platform}:shared"` which is wrong for multi-user setups like Discord.

Closing this PR for now. The fix should target the Hermes gateway where `conversation_id` was first identified as a missing kwarg.

Issue #175, opened 2026-05-16
feat: add /lcm rotate for in-place tail-preserving session compaction


GodsBoy:
### Problem

Long-running LCM sessions accumulate raw messages that have already been
compacted into summary DAG nodes, but the lifecycle frontier marker only
advances inside the compaction loop. There is no operator-level command
to compact a session in place, on demand, without changing session
identity. Today the operator's choices are:

- wait for automatic compaction to drag the frontier forward (no
  guarantee of timing on quiet sessions)
- `/new` at the Hermes layer, which starts a fresh session and loses
  in-session continuity
- `/lcm doctor clean` (with apply gate), which targets junk sessions,
  not an in-place compact of an active session

`lossless-claw` upstream of `hermes-lcm` has `/lcm rotate` for this
exact case. `hermes-lcm` does not yet, even though the README explicitly
positions it as inspired by lossless-claw.

### User or operator impact

Operators running heavy interactive sessions (Telegram/WhatsApp gateways,
long Hermes CLI conversations) cannot deterministically compact an
active session and reclaim the active-context budget that pre-tail
raw messages currently occupy. The frontier may not advance for many
turns even after summarization has covered the pre-tail range, leaving
the assembled prompt larger than necessary.

Smaller, secondary impact: there is no operator-visible signal for
"when was this session last manually compacted." Diagnostics
(`/lcm status`, `lcm_status` tool) cannot answer that question today.

### Proposed direction

Add `/lcm rotate` and `/lcm rotate apply` on the existing slash-command
surface (already gated behind `LCM_ENABLE_SLASH_COMMAND`):

- **`/lcm rotate`** (read-only preview): reports the planned frontier
  advance, fresh-tail boundary, total/pre-tail message counts, and the
  rolling rotate-latest backup path that `apply` would touch. No disk
  writes, no lifecycle mutation.
- **`/lcm rotate apply`** (mutating, backup-first): advances the
  lifecycle frontier past pre-tail raw messages, writing a rolling
  `{db_stem}-rotate-latest.sqlite3` snapshot atomically via tmp +
  `Path.replace` before mutating state. The rolling slot keeps disk
  usage bounded across repeated rotates.

Identity invariants:

- `session_id` unchanged
- `conversation_id` unchanged
- Raw messages remain in the SQLite store and remain recoverable
  through `lcm_load_session` and `lcm_expand` (confirmed against
  `tools.py` — both are frontier-unaware). The lossless raw recovery
[...]

PR #176, opened 2026-05-16
feat: add /lcm rotate for in-place tail-preserving session compaction

GodsBoy:
## Summary

- Adds `/lcm rotate` (read-only preview) and `/lcm rotate apply`
  (backup-first mutation) on the existing slash-command surface.
  Rotate advances the lifecycle frontier marker past pre-tail raw
  messages without changing `session_id` or `conversation_id`. Raw
  messages remain in the SQLite store and stay recoverable through
  `lcm_load_session` and `lcm_expand` so the lossless raw recovery
  contract is preserved.
- Rolling `{db_stem}-rotate-latest.sqlite3` snapshot written
  atomically via tmp + `Path.replace`. No-op rerun short-circuits
  before any disk write so the previous known-good snapshot is
  preserved across idempotent retries.
- `lcm_status` / `/lcm status` surface `last_rotate_at` (UTC ISO-8601
  with explicit `+00:00`), `rotate_backup_path`, and
  `rotate_backup_size`. `LCM_STATUS` tool schema description updated.

## Why

`hermes-lcm` is positioned as inspired by `lossless-claw`, which has
`/lcm rotate` for the in-session compact case. Operators running
heavy interactive sessions today have no deterministic way to compact
an active session in place: automatic compaction is timing-driven,
`/new` at the Hermes layer loses in-session continuity, and
`/lcm doctor clean` targets junk sessions, not an in-place compact of
an active one. This closes that gap with a backup-first, refusal-aware
command that follows the existing `/lcm doctor *` pattern.

Refs #175 for the full problem framing, alternatives considered, and
backward-compatibility notes.

## Validation

- [x] Focused validation: `pytest tests/test_lcm_rotate.py -q` -> `19 passed`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `564 passed`
  - [x] `pytest -q` -> `695 passed`
  - [x] `bash -lc 'ulimit -n 1024 && pytest -q'` -> `695 passed`
  - [x] `python -m compileall -q .` -> `OK`
  - [x] `python -m py_compile scripts/import_lossless_claw.py` -> `OK`
  - [x] `bash -n scripts/install.sh scripts/update.sh` -> `OK`
  - [x] `git diff --check` -> `clean`
- [ ] Workflow validation, if workflows changed: `actionlint` (n/a — no workflow changes)

## Notes

**Scope discipline.** I deliberately constrained this PR to the rotate
surface and the smallest engine and command-side additions needed to
support it. Refusal/no-op reason codes are documented on
`rotate_active_session` so future readers do not have to grep for the
bare strings.

**Refusal codes covered:** `no_active_session`, `session_ignored`,
`session_stateless`, `no_pre_tail_content`, `empty_tail`,
`frontier_already_ahead`, `stale_lifecycle_state`. The
`stale_lifecycle_state` code surfaces when `advance_frontier` returns
the unchanged state because the row's `current_session_id` no longer
matches this engine; rotate detects that case explicitly rather than
masking it as a successful frontier advance.

**Internals consolidated, no behavior change for existing callers:**

- `engine.backup_dir()` centralizes the backup directory derivation
  so the timestamped `/lcm backup` slot and the new rolling rotate
  slot cannot drift on `hermes_home` or `db_path` changes
- `command._flush_engine_connections()` extracted to share the
  three-connection commit between `/lcm backup` and `/lcm rotate apply`
- `get_status()` uses a single `stat()` on the rolling backup path,
  avoiding a TOCTOU window between separate mtime and size reads

**Pre-existing concerns surfaced during review, intentionally out of scope:**

- `lifecycle_state.advance_frontier` does a non-atomic read-then-update
  under autocommit (`lifecycle_state.py:549-572`). Pre-existing, not
  introduced by this PR, but rotate adds another caller. Happy to
  file a separate follow-up PR converting it to a single-statement
  `MAX()` update if you would like that addressed independently.
- `last_rotate_at` is derived from the rolling backup's mtime and is
  therefore spoofable by anything that touches the file. The README
  documents this; persisting the timestamp in `lcm_lifecycle_state`
  is a defensible follow-up if operational pain shows up.

**Multi-reviewer pass.** Before opening the PR I ran correctness,
testing, maintainability, project-standards, security, adversarial,
and python-idioms reviews on the working diff. All actionable findings
were addressed: `empty_tail` KeyError safety, persist-then-mutate
order for the in-process `_last_compacted_store_id` marker,
backup-only-after-noop-check (so idempotent rerun does not destroy the
previous known-good snapshot), standardized `rotate_backup_path` key
name across preview/apply/status surfaces, timezone-aware UTC
formatting on `last_rotate_at`, `pathlib`-native `Path.replace` for
the atomic rename, single `stat()` call in `get_status` for the
rolling backup, and refusal-codes documented on the engine method.

**No agent-tool surface change.** No new agent tool is added, no
existing tool schema parameters change. The `LCM_STATUS` description
is widened to mention the new fields so the model understands what
it sees.

Refs #175

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -2016,6 +2016,26 @@ class LCMEngine(ContextEngine):
             )
         except Exception as exc:  # pragma: no cover - defensive
             status["lifecycle_fragmentation"] = {"error": str(exc), "read_only": True}
+        try:
+            rotate_backup_path = self.rotate_backup_path()
+            status["rotate_backup_path"] = str(rotate_backup_path)
+            # Single stat() to avoid a TOCTOU window where the rolling slot
+            # could be atomically replaced between separate mtime and size reads.
+            try:
+                rotate_stat = rotate_backup_path.stat()
+            except FileNotFoundError:
+                rotate_stat = None
+            if rotate_stat is not None:
+                status["last_rotate_at"] = rotate_stat.st_mtime
+                status["rotate_backup_size"] = rotate_stat.st_size
+            else:
+                status["last_rotate_at"] = None
+                status["rotate_backup_size"] = 0
+        except Exception as exc:  # pragma: no cover - defensive
+            status["rotate_backup_path"] = None
+            status["last_rotate_at"] = None
+            status["rotate_backup_size"] = 0
+            status["rotate_backup_error"] = str(exc)
         if session_id:
             status["store_messages"] = self._store.get_session_count(session_id)
             status["dag_nodes"] = len(self._dag.get_session_nodes(session_id))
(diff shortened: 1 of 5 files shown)


stephenschoettler:
Thanks. CI is green on `539fe3e`, and I've asked @Tosko4 for review. If @Tosko4 approves and checks stay green, I'll merge this.

engine.py
```diff
@@ -3606,6 +3626,163 @@ def _extract_expand_hint(summary: str) -> str:
...
+        )
+        if persisted_frontier < new_frontier:
+            return {
+                **{k: v for k, v in result.items() if k != "ok"},
+                "ok": False,
+                "noop": False,
+                "reason": "stale_lifecycle_state",
+                "applied_frontier_store_id": persisted_frontier,
+            }
+        self._last_compacted_store_id = max(
```
Tosko4: Advancing the live `_last_compacted_store_id` here is unsafe unless the command also rewrites or reconciles the current active context. The persisted lifecycle frontier is for future bootstrap, but the same running process can still have pre-rotate messages in memory. On the next `compress()`, `_get_store_ids_for_messages()` filters out rows at or below this marker, so it can create a summary node whose text covers old messages while `source_ids` points only at a post-frontier row. I reproduced that with an ingested `system + msg-0..msg-9` session: rotate moved the marker to `8`, then the next compaction summarized `msg-0..msg-7` with `source_ids == [9]`.

Tosko4 (changes requested):
I agree with the command shape and the backup-first/no-op handling is mostly clean, but I do not think this is safe to merge yet.

The apply path advances the live engine's `_last_compacted_store_id` even though the host's current active context is not rewritten by the command. That leaves a same-process window where the next normal `compress()` can summarize pre-rotate messages while `_get_store_ids_for_messages()` refuses to look at rows at or below the rotated frontier.

I reproduced this locally on the current head. After ingesting `system + msg-0..msg-9`, running `/lcm rotate apply` moved the frontier to `8`. Continuing with the same in-memory active messages plus one assistant message, then running `compress()` with a deterministic summarizer produced a DAG node whose summary covered `msg-0` through `msg-7`, but `source_ids` was only `[9]`. That breaks the source lineage and expansion contract, and later condensation would be building on a poisoned node.

I think the fix needs to keep the persisted/bootstrap frontier separate from the live source-mapping frontier unless the active context is also rebuilt or reconciled. The regression should cover same-process rotate followed by normal compaction, and assert that any summary node's `source_ids` match the raw messages it summarized.

```text
pytest tests/test_lcm_rotate.py -q
19 passed

564 passed

pytest -q
695 passed
```

So the existing suite is green, but it misses this active-context/frontier interaction.

engine.py
GodsBoy: Fixed in 19912d7. The in-process source-mapping marker no longer advances on rotate, only the persisted lifecycle frontier moves. The two markers are now explicitly documented as separate concerns in `rotate_active_session`'s docstring:

- **Persisted frontier** (`lifecycle.current_frontier_store_id`) — bootstrap signal for the next process start. Rotate advances this.
- **In-process source-mapping marker** (`self._last_compacted_store_id`) — tracks rows the *current process* has moved into summary nodes; used by `_get_store_ids_for_messages` as a candidate filter. Rotate deliberately does **not** touch this. On next process start, `_bind_lifecycle_state` reads the persisted frontier into the marker against a freshly-built active context, so the contract holds.

Regression test (`test_rotate_apply_does_not_corrupt_source_lineage_on_next_compress`) reproduces your exact scenario: `system + msg-0..msg-9`, `/lcm rotate apply`, append one assistant message, stub the summarizer for determinism, run `compress()`, then assert that the resulting summary node's `source_ids` contains at least one row at or below the rotate frontier. Reverting the one-line fix in isolation makes the regression fail with a severed-lineage assertion; the fix makes it pass.

I also updated the existing assertion `engine._last_compacted_store_id >= 7` in `test_rotate_apply_advances_frontier_and_writes_rolling_backup` it was inadvertently encoding the bug as desired behavior. It now pins the marker at `0`.

Validation re-run:

```
pytest tests/test_lcm_rotate.py -q                                          -> 20 passed
pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q  -> 564 passed
pytest -q                                                                   -> 696 passed
bash -lc 'ulimit -n 1024 && pytest -q'                                      -> 696 passed
python -m compileall -q .                                                   -> OK
git diff --check                                                            -> clean
```

Thanks for catching this and for the reproducer, the explicit `source_ids == [9]` shape made the failure mode unambiguous.

Tosko4 (approved):
Re-reviewed `19912d7`. This clears my blocker.

`/lcm rotate apply` now keeps the persisted lifecycle frontier separate from the in-process source-mapping marker, so same-process compaction still maps pre-rotate active messages back to the correct raw rows. I reran the original repro shape locally and the next `compress()` produced `source_ids == [2, 3, 4, 5, 6, 7, 8, 9]` for the summarized `msg-0..msg-7` range, instead of the previous severed-lineage `[9]` case.

Validation on `19912d7`:

```text
pytest tests/test_lcm_rotate.py -q
20 passed

564 passed

pytest -q
696 passed

python -m compileall -q .
OK

git diff --check
clean
```

Approving from my side.

stephenschoettler:

PR #180, opened 2026-05-19
fix: avoid data URI scaffold false positives

duclamvan:
## Summary
- Avoid treating literal diagnostic scaffolds like `data:%;base64,%` as real inline data URIs in `lcm_doctor` payload checks.
- Avoid treating code examples like `DATA_URI = "data:image/png;base64," + DATA_PAYLOAD` as real inline data URI payloads.
- Pre-filter broadly in SQL, then apply the existing conservative Python data-URI regex used by ingest externalization.
- Add regression tests for both harmless scaffold text and code-prefix-only examples.

## Why
Doctor/tool output and tests can include diagnostic/code strings that mention a data URI prefix without embedding an actual payload. If those strings are re-ingested, broad SQL matching can make the warning self-reproducing even though there is no inline media payload to externalize.

## Test Plan
- `python -m pytest tests/test_ingest_protection.py::test_lcm_doctor_reports_largest_and_suspicious_payload_rows tests/test_ingest_protection.py::test_lcm_doctor_ignores_literal_data_uri_like_scaffold tests/test_ingest_protection.py::test_lcm_doctor_ignores_code_data_uri_prefix_without_payload -q`

### DIFF
--- a/ingest_protection.py
+++ b/ingest_protection.py
@@ -541,28 +541,38 @@ def scan_sqlite_payload_risks(conn, *, limit: int = 5) -> dict[str, Any]:
         """,
         (limit,),
     ).fetchall()
-    data_uri_content = conn.execute(
+    candidate_cap = max(limit * 20, limit)
+    # Pre-filter broadly in SQL, then apply the same conservative Python regex
+    # used by ingest externalization. This avoids false positives from code or
+    # doctor text that quotes scaffolds such as "data:%;base64,%" or
+    # `DATA_URI = "data:image/png;base64," + DATA_PAYLOAD`.
+    data_uri_content_candidates = conn.execute(
         """
         SELECT store_id, session_id, source, role, COALESCE(length(content), 0) AS content_len, content
         FROM messages
-        WHERE content LIKE '%data:%;base64,%'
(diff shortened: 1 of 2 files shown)


stephenschoettler:

Please add:
- Validation

stephenschoettler:
Thanks. CI is green on `ac24d6d`, and I've asked @Tosko4 for review. If @Tosko4 approves and checks stay green, I'll merge this.

Tosko4 (approved):
Reviewed the current head. The fix is in the right place: `lcm_doctor` now uses SQL only as a broad candidate filter, then applies the same conservative detector used by ingest protection. That removes the diagnostic scaffold and code-prefix false positives while real inline data URIs are still reported.

I also checked the candidate cap. For this diagnostic surface it is an acceptable bounded scan tradeoff, not a storage safety issue.

Validation on `ac24d6d`:

`python -m pytest tests/test_ingest_protection.py -q`: 51 passed

`python -m pytest -q`: 699 passed

`python -m compileall -q .`

`git diff --check`

Approving from my side.

stephenschoettler:

PR #181, opened 2026-05-20
fix: suppress preflight compression when LCM would noop

WXBR:
## Summary
- Adds a preflight eligibility check before LCM reports that compression should run.
- Suppresses preflight compression when the active context is over the global threshold but the only available raw backlog is protected by the fresh tail or smaller than the leaf chunk threshold.
- Preserves normal preflight compression when an eligible old raw leaf chunk exists.
- Preserves deferred/emergency maintenance preflight eligibility when existing raw-backlog debt is under critical budget pressure, matching `compress()`.

## Why
- Hermes can emit a user-visible "Compacting context" status before calling `compress()`.
- If LCM then immediately no-ops with `raw backlog outside fresh tail is below leaf chunk threshold`, the session can stay above threshold and repeatedly show compaction status without any compaction happening.
- The previous guard also suppressed deferred/emergency maintenance when raw backlog was below the leaf chunk threshold even though `compress()` would still compact under deferred maintenance debt plus critical budget pressure.
- This makes preflight match `compress()` instead of blocking a real maintenance compaction path.

## Validation
- [x] Focused preflight validation: `python -m pytest tests/test_lcm_engine.py::TestEngineABC::test_preflight_does_not_request_compaction_when_only_fresh_tail_is_over_threshold tests/test_lcm_engine.py::TestEngineABC::test_preflight_requests_compaction_for_deferred_maintenance_under_critical_pressure tests/test_lcm_engine.py::TestEngineABC::test_preflight_requests_compaction_when_old_backlog_has_leaf_chunk -q -o 'addopts='` -> `3 passed`
- [x] `python -m compileall -q engine.py tests/test_lcm_engine.py` -> passed
- [ ] Default validation:
  - [ ] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> earlier local macOS run: `565 passed, 1 skipped, 1 failed` due to existing SQLite lock timing assertion (`elapsed < 0.3`)
  - [ ] `pytest -q` -> earlier local macOS run: `695 passed, 1 skipped, 3 failed`; failures appeared unrelated to this patch: the SQLite lock timing assertion above and two `/var` vs `/private/var` temp path string-prefix assertions
  - [ ] Latest `python -m pytest tests/test_lcm_engine.py -q -o 'addopts='` -> `383 passed, 1 skipped, 1 failed`; same SQLite lock timing assertion (`elapsed < 0.3`, observed `0.376s`) reproduced locally on macOS
  - [ ] `bash -lc 'ulimit -n 1024 && pytest -q'` -> not run
  - [x] `python -m py_compile scripts/import_lossless_claw.py` -> passed earlier
  - [x] `bash -n scripts/install.sh scripts/update.sh` -> passed earlier
  - [x] `git diff --check` -> passed earlier
- [ ] Workflow validation, if workflows changed: not applicable, no workflow changes

## Notes
- Code intentionally stays minimal: preflight now mirrors `compress()` at both boundaries: normal leaf eligibility and deferred maintenance under critical budget pressure.

Refs #

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -537,6 +537,8 @@ class LCMEngine(ContextEngine):
             eligible, reason = self._leaf_compaction_candidate_status(messages)
             if eligible:
                 return True
+            if self._should_run_deferred_maintenance(messages, observed_tokens=rough):
+                return True
             self._last_compression_status = "noop"
             self._last_compression_noop_reason = reason
             logger.info("LCM preflight compression no-op: %s", reason)
--- a/engine.py
+++ b/engine.py
@@ -534,9 +534,54 @@ class LCMEngine(ContextEngine):
         if self._should_force_overflow_recovery(observed_tokens=rough):
             return True
         if self.threshold_tokens > 0 and rough >= self.threshold_tokens:
-            return True
+            eligible, reason = self._leaf_compaction_candidate_status(messages)
+            if eligible:
+                return True
+            self._last_compression_status = "noop"
+            self._last_compression_noop_reason = reason
+            logger.info("LCM preflight compression no-op: %s", reason)
+            return False
         self._refresh_raw_backlog_debt(messages, observed_tokens=rough)
         return self._should_run_deferred_maintenance(messages, observed_tokens=rough)
+
+    def _leaf_compaction_candidate_status(
+        self,
+        messages: List[Dict[str, Any]],
+        *,
+        force_overflow: bool = False,
+    ) -> tuple[bool, str]:
+        """Return whether a normal leaf compaction pass can actually run.
+
+        The host asks ``should_compress_preflight`` before it emits user-visible
+        compression status. A session can be over the global context threshold
+        while all pressure sits in the protected fresh tail, or while the raw
+        backlog outside that tail is still smaller than the configured leaf
+        chunk. In that case ``compress()`` would immediately no-op, so preflight
+        should not advertise a compaction attempt yet.
+        """
+        if not messages:
+            return False, "empty message list"
+        n = len(messages)
+        fresh_tail_start = max(0, n - self._config.fresh_tail_count)
+        leading_anchor_count = self._leading_anchor_count(messages)
+        if fresh_tail_start <= leading_anchor_count:
+            return False, "no eligible raw backlog outside fresh tail"
+
+        candidate_raw = messages[leading_anchor_count:fresh_tail_start]
+        if not candidate_raw:
+            return False, "no eligible raw backlog outside fresh tail"
+
+        if force_overflow:
+            return True, "forced overflow recovery"
+
+        raw_tokens_outside_tail = count_messages_tokens(candidate_raw)
+        if self._config.dynamic_leaf_chunk_enabled:
+            working_leaf_chunk_tokens = self._working_leaf_chunk_tokens(raw_tokens_outside_tail)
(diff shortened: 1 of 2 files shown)


stephenschoettler:

Please add:
- Validation

WXBR:
Edited.

stephenschoettler (changes requested):
Thanks for updating the PR body. I approved the fork CI run after checking that the diff only touches `engine.py` and `tests/test_lcm_engine.py`.

Requesting one fix before merge: preflight now returns `False` for every over-threshold context whose raw backlog is below the leaf chunk threshold, but `compress()` can still compact that same backlog when deferred maintenance debt exists and critical budget pressure is reached.

I reproduced the mismatch locally on this head:
- `_should_run_deferred_maintenance(...)=True`
- `_critical_budget_pressure_reached(...)=True`
- `_leaf_compaction_candidate_status(...)=False`
- `should_compress_preflight(...)=False`

That suppresses the deferred/emergency maintenance path and can leave an over-budget session unable to compact even though `compress()` would not no-op.

Please make preflight use the same eligibility rule as `compress()` for the `deferred_maintenance_active and critical_budget_pressure` case, and add a regression for it.

Local validation otherwise passed on Linux:
- focused new tests: passed
- default slice: `567 passed`
- full suite: `699 passed`
- `python -m compileall -q engine.py tests/test_lcm_engine.py`: passed
- `python -m py_compile scripts/import_lossless_claw.py`: passed

WXBR:
Pushed a fix for the deferred maintenance mismatch in 4b2dea6. Preflight now returns true when deferred raw-backlog maintenance is active under critical budget pressure, even if the normal leaf candidate check reports below-threshold raw backlog. Added a regression for that path.

Validation:
- `python -m pytest tests/test_lcm_engine.py::TestEngineABC::test_preflight_does_not_request_compaction_when_only_fresh_tail_is_over_threshold tests/test_lcm_engine.py::TestEngineABC::test_preflight_requests_compaction_for_deferred_maintenance_under_critical_pressure tests/test_lcm_engine.py::TestEngineABC::test_preflight_requests_compaction_when_old_backlog_has_leaf_chunk -q -o 'addopts='` -> `3 passed`
- `python -m compileall -q engine.py tests/test_lcm_engine.py` -> passed

WXBR:
Rebased onto current main after the v0.11.0 release/docs changes. Latest head is a9eb75b; focused preflight regression tests and compileall still pass after the rebase.

stephenschoettler (approved):
Re-reviewed current head a9eb75b. The deferred-maintenance blocker from my earlier review is fixed: preflight now returns true when raw-backlog debt is active under critical budget pressure, matching the `compress()` escape hatch, and the new regression covers that path.

I also approved the current fork CI run after confirming the diff still only touches `engine.py` and `tests/test_lcm_engine.py`.

Validation on Linux:
- focused preflight tests: `3 passed`
- direct preflight/compress probe for deferred debt + critical pressure: passed
- default slice: `568 passed`
- full suite: `702 passed`
- low-FD full suite: `702 passed`
- `python -m py_compile scripts/import_lossless_claw.py`: passed
- `git diff --check origin/main...HEAD`: passed

CI is green across workflow-lint and Python 3.11, 3.12, 3.13, 3.14.

PR #185, opened 2026-05-21
fix: avoid union type arrays in lcm_grep schema

### LINKED ISSUE #182: lcm_grep schema uses JSON Schema union type that crashes Kimi/Moonshot models (TypeError: unhashable type: list)
## Bug

The `lcm_grep` tool schema in `schemas.py` uses JSON Schema union type arrays for the `time_from` and `time_to` parameters:

```python
"time_from": {
    "type": ["number", "string"],
    ...
},
"time_to": {
    "type": ["number", "string"],
    ...
},
```

This is valid JSON Schema, but it causes a `TypeError: unhashable type: 'list'` crash in Hermes Agent when using **Kimi K2.6** or any Moonshot-family model. The crash happens in `agent/moonshot_schema.py` in `_fill_missing_type()` at this check:

```python
if "type" in node and node["type"] not in {None, ""}:
```

When `node["type"]` is `["number", "string"]` (a Python list), the `not in {None, ""}` comparison tries to hash the list, which raises `TypeError: unhashable type: 'list'`. This is classified as a non-retryable client error, so the entire API call aborts immediately — the model becomes completely unusable.

## Reproduction

1. Configure Kimi K2.6 as the model (via `ollama-cloud` provider or any Moonshot endpoint).
2. Ensure the `hermes-lcm` plugin is installed (it registers `lcm_grep` as a tool).
3. Send any message — the agent immediately crashes with `unhashable type: 'list'`.

## Root Cause

`moonshot_schema.py`'s `_fill_missing_type()` function does not handle the case where `node["type"]` is a list (JSON Schema union type). Python cannot hash a list, so the `not in` comparison raises `TypeError`.

### 1. Fix `schemas.py` (plugin side — simplest)

Replace union type arrays with `anyOf`:

```python
"time_from": {
    "anyOf": [{"type": "number"}, {"type": "string"}],
    "description": "Optional inclusive minimum raw-message timestamp...",
},
```

This avoids the union type array entirely and is compatible with all OpenAI-compatible API providers.

### 2. Harden `moonshot_schema.py` (Hermes Agent side — defensive)

Add a guard before the hash comparison to flatten union type arrays into `anyOf`:

```python
def _fill_missing_type(node: Dict[str, Any]) -> Dict[str, Any]:
    """Infer a reasonable type if this schema node has none."""
    # Handle JSON Schema union type arrays: ["number", "string"]
    if "type" in node and isinstance(node["type"], list) and node["type"]:
        if "anyOf" not in node:
            node["anyOf"] = [{"type": t} for t in node["type"] if isinstance(t, str)]
        node["type"] = node["type"][0] if isinstance(node["type"][0], str) else "string"
    if "type" in node and node["type"] not in {None, ""}:
        return node
    ...
```

I have applied fix #2 locally and confirmed Kimi K2.6 works correctly with `anyOf` + scalar `type`.

## Impact

- **Severity**: High — makes Kimi/Moonshot models completely unusable when hermes-lcm is installed.
- **Affected models**: All Moonshot/Kimi models (kimi-k2.6, moonshot-v1, etc.) via ollama-cloud, direct Moonshot API, or any aggregator routing to Moonshot.
- **Workaround**: Without hermes-lcm installed, the crash does not occur (no `type: [list]` in any tool schema).

stephenschoettler:
## Summary
- Replaces `lcm_grep` timestamp schema union type arrays with `anyOf` entries.
- Adds a packaging regression that asserts `time_from` and `time_to` do not expose list-valued `type` fields.

## Why
- Kimi/Moonshot schema normalization can crash on JSON Schema `type: ["number", "string"]` because the provider adapter treats `type` as hashable scalar data.
- `anyOf` keeps the same accepted argument contract while avoiding that provider-specific failure path.

## Validation
- [x] Focused validation: `python -m pytest tests/test_packaging_install.py::test_lcm_grep_time_filters_use_anyof_not_union_type_arrays -q -o 'addopts='` -> `1 passed`
- [x] Focused validation: `python -m pytest tests/test_packaging_install.py -q -o 'addopts='` -> `9 passed`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `569 passed`
  - [x] `pytest -q` -> `703 passed`
  - [x] `bash -lc 'ulimit -n 1024 && pytest -q'` -> `703 passed`

## Notes
- No workflow files changed.
- No runtime parsing changes, this only changes the tool schema shape exposed to model/tool adapters.

Fixes #182

### DIFF
--- a/schemas.py
+++ b/schemas.py
@@ -73,14 +73,14 @@ LCM_GREP = {
                 "description": "Optional raw-message role filter. When supplied, lcm_grep returns raw message hits only.",
             },
             "time_from": {
-                "type": ["number", "string"],
+                "anyOf": [{"type": "number"}, {"type": "string"}],
                 "description": (
                     "Optional inclusive minimum raw-message timestamp. Accepts Unix seconds or timezone-aware ISO 8601; "
                     "naive ISO timestamps are rejected. When supplied, lcm_grep returns raw message hits only."
                 ),
             },
             "time_to": {
-                "type": ["number", "string"],
+                "anyOf": [{"type": "number"}, {"type": "string"}],
                 "description": (
                     "Optional inclusive maximum raw-message timestamp. Accepts Unix seconds or timezone-aware ISO 8601; "
                     "naive ISO timestamps are rejected. When supplied, lcm_grep returns raw message hits only."
--- a/tests/test_packaging_install.py
+++ b/tests/test_packaging_install.py
@@ -115,6 +115,18 @@ def test_install_script_refuses_to_replace_existing_non_symlink_path(tmp_path):
     assert target.is_dir()
 
 
+def test_lcm_grep_time_filters_use_anyof_not_union_type_arrays():
+    engine = _register_plugin_engine("hermes_lcm_schema_shape")
+    assert engine is not None
+    schemas = {schema["name"]: schema for schema in engine.get_tool_schemas()}
+    properties = schemas["lcm_grep"]["parameters"]["properties"]
+
+    for name in ("time_from", "time_to"):
+        field = properties[name]
+        assert field["anyOf"] == [{"type": "number"}, {"type": "string"}]
+        assert "type" not in field
+
+
 def test_plugin_entrypoint_registers_lcm_context_engine():
     engine = _register_plugin_engine("hermes_lcm_packaging_entrypoint")
 


Tosko4 (approved):
Reviewed the current head. This is the right plugin-side shape for #182: it removes the list-valued `type` fields while preserving number/string inputs through `anyOf`.

I also checked the emitted `lcm_grep` schema against Hermes' Moonshot schema sanitizer and confirmed there are no list-valued `type` entries left in the plugin schema.

Validation I ran locally:

```
python -m pytest tests/test_packaging_install.py::test_lcm_grep_time_filters_use_anyof_not_union_type_arrays -q -o 'addopts='
python -m pytest tests/test_packaging_install.py -q -o 'addopts='
pytest -q
python -m compileall -q .
```

Approving from my side.

PR #191, opened 2026-05-21
fix: ignore disabled Hermes compression threshold

### LINKED ISSUE #187: Bug Report: Compression Threshold Fallback Overrides LCM_CONTEXT_THRESHOLD
**Title:** `_hermes_compression_threshold()` ignores `compression.enabled: false`, reads threshold from config.yaml regardless

---

**Expected behavior**

When `compression.enabled: false` is set in `config.yaml` and `LCM_CONTEXT_THRESHOLD=0.75` is set in `.env`, LCM should use `context_threshold: 0.75`. The Hermes agent's built-in compressor settings should NOT leak into LCM's config when disabled.

---

**Actual behavior**

`_hermes_compression_threshold()` in `config.py` (lines 48-77) reads `compression.threshold` from `config.yaml` **without checking `compression.enabled`**. Even though the user has explicitly disabled compression, the value `threshold: 0.5` is injected as LCM's fallback `context_threshold`. This means:

- LCM compacts at 50% (~128K tokens on a 256K context) instead of the desired 75% (~192K tokens)
- When LCM compacts, the WebUI detects the compaction via `compression_count` (streaming.py line 4214-4216) and sends a `compressed` SSE event to the frontend
- The frontend shows a "compression card" and the user's current answer is swallowed — they see only tool output, no assistant text

---

**Minimal reproduction**

1. Configure `~/.hermes/config.yaml`:
```yaml
compression:
  enabled: false
  threshold: 0.5       # default
  compressor:
    enabled: false

context:
  engine: lcm
```

2. Set `~/.hermes/.env`:
```bash
LCM_CONTEXT_THRESHOLD=0.75
```

3. Start the WebUI (so .env is loaded into process env).

4. Run a conversation until ~128K+ tokens (50% of 256K context). LCM will compact.

5. Observe: LCM's `context_threshold` in `lcm_status` shows `0.5`, not `0.75`.

Bonus: the WebUI's `streaming.py` then detects the LCM compaction via `compression_count` and sends a `compressed` event, causing the ongoing assistant response to be lost.

---

**Version and branch**

- hermes-lcm v0.11.1
- Commit `71dd8a1` (fix: avoid union type arrays in lcm_grep schema (#185))
- Branch: main

---

**Environment**

- Hermes Agent: v0.14.0
- Python: 3.11.15
- OS: Linux (Proxmox LXC, Debian/Ubuntu)
- WebUI: nesquena/hermes-webui (runs agent process directly, no separate gateway)

---

**Relevant config**

```yaml
compression:
  enabled: false
  threshold: 0.5       # ← read by _hermes_compression_threshold() as fallback!
  target_ratio: 0.2
  protect_last_n: 20
  compressor:
    enabled: false

context:
  engine: lcm
```

```bash
# .env
LCM_CONTEXT_THRESHOLD=0.75    # ← overridden by fallback to 0.5
```

---

**Logs, traceback, or failing test**

`lcm_status` output showing the bug:

```json
{
  "context_threshold": 0.5,          // should be 0.75
  "threshold_tokens": 128000,        // should be 192000
  "compression_count": 1,
  "last_compression_status": "compacted"
}
```

After the workaround (`compression.threshold: 0.75` + WebUI restart):

```json
{
  "context_threshold": 0.75,
  "threshold_tokens": 192000,
  "compression_count": 0,
  "last_compression_status": "idle"
}
```

---

**Validation attempted**

1. Confirmed `LCM_CONTEXT_THRESHOLD=0.75` is set in the process environment (`echo $LCM_CONTEXT_THRESHOLD` → `0.75`)
2. Confirmed `_hermes_compression_threshold()` returns 0.5 from config.yaml despite `compression.enabled: false`
3. Verified workaround: setting `compression.threshold: 0.75` in config.yaml (even with `enabled: false`) makes LCM correctly use 0.75
4. Tested: after WebUI restart + fresh chat, `lcm_status` shows `context_threshold: 0.75` and `compression_count: 0`
5. Large chat no longer triggers premature compaction

stephenschoettler:
## Summary
- Ignore Hermes `compression.threshold` as an LCM fallback when `compression.enabled: false`.
- Preserve the existing fallback when Hermes compression is enabled or unspecified and `LCM_CONTEXT_THRESHOLD` is missing.
- Add regression coverage for the normal PyYAML path and the minimal no-PyYAML parser path.

## Why
- A disabled Hermes compressor should not pull LCM's `context_threshold` down to the disabled host compression threshold.
- This prevents premature LCM compaction in configurations that explicitly disable host compression and rely on LCM's own default or `LCM_CONTEXT_THRESHOLD`.

## Validation
- [x] Focused validation: `python -m pytest tests/test_lcm_core.py::TestConfig -q` -> `8 passed`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `571 passed`
  - [x] `pytest -q` -> `706 passed`
  - [ ] `bash -lc 'ulimit -n 1024 && pytest -q'` -> not run, local execution policy denied lowering the fd limit
  - [x] `git diff --check`

## Notes
- Regression tests were written first and failed as expected before the fix: the disabled Hermes threshold still produced `context_threshold == 0.5`.
- No workflow files changed.

Fixes #187

### DIFF
--- a/config.py
+++ b/config.py
@@ -45,13 +45,22 @@ def _parse_bool_env(key: str, default: bool) -> bool:
     return default
 
 
+def _config_bool_disabled(value) -> bool:
+    if isinstance(value, bool):
+        return value is False
+    if isinstance(value, str):
+        return value.strip().lower() in {"0", "false", "no", "off"}
+    return False
+
+
 def _hermes_compression_threshold(default: float) -> float:
     """Read Hermes compression.threshold when no LCM env override is present.
 
     Hermes gateways may load ``~/.hermes/config.yaml`` without exporting every
     setting into the process environment. Falling back to the main Hermes
     compression threshold keeps LCM aligned with the active agent config while
-    still allowing ``LCM_CONTEXT_THRESHOLD`` to override it explicitly.
+    still allowing ``LCM_CONTEXT_THRESHOLD`` to override it explicitly. Disabled
+    Hermes compression should not leak its threshold into LCM.
     """
     home = Path(os.environ.get("HERMES_HOME") or Path.home() / ".hermes")
     cfg_path = home / "config.yaml"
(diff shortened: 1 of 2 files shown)


Tosko4 (approved):
Rebased this onto current main and re-reviewed `4a1e34a`. The fallback now does the right thing: a disabled Hermes compressor no longer supplies its threshold to LCM, while explicit `LCM_CONTEXT_THRESHOLD` still wins and the existing enabled or unspecified fallback is preserved.

I reran direct config probes, `TestConfig`, the focused core, engine, and packaging set, the full suite twice including `ulimit -n 1024`, compile checks, shell syntax checks, and diff check. Current head CI is green too. Approving from my side.

PR #194, opened 2026-05-21
feat: add Codex GPT benchmark policy

stephenschoettler:
## Summary
- Add the `codex_gpt_long_context` benchmark candidate as a built-in policy and committed policy file.
- Add deterministic `--synthetic-fixture name:pairs:canaries:filler_words` support for large pressure probes without committing huge transcripts.
- Harden benchmark CLI and replay safety: bounded synthetic specs, input loading before output directory creation, explicit refusal to reuse non-empty per-run directories, and CLI argv regression coverage.
- Add benchmark tests for policy parity, synthetic fixture parsing, CLI behavior, replay isolation, and output safety.

## Why
- Moves #189 from “the harness can compare policies” toward “a candidate policy can beat baseline under pressure” while keeping this benchmark-only.
- Provides repeatable pressure evidence for GPT/Codex long-context tuning without runtime preset selection or live config mutation.

## Validation
- [x] Focused validation: `python -m pytest tests/test_benchmarking_cli.py tests/test_benchmarking_fixtures.py tests/test_benchmarking_replay.py tests/test_benchmarking_report.py tests/test_benchmarking_types.py -q` -> `33 passed`
- [x] Pressure smoke: `python scripts/lcm_benchmark.py --synthetic-fixture codex_pressure_probe:42:4:1000 --policy benchmarks/policies/baseline.yaml --policy benchmarks/policies/codex_gpt_long_context.yaml --output /tmp/hermes-lcm-codex-gpt-pressure-cli-* --allow-external-output --json` -> `runs 2`, `failures 0`, `codex_gpt_long_context@1 score 92.5`, `baseline_272k@1 score 72.5`
- [x] Default validation:
  - [x] `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `579 passed`
  - [x] `python -m pytest -q` -> `747 passed`
  - [ ] `bash -lc 'ulimit -n 1024 && pytest -q'` -> local execution policy blocked lowering the file descriptor limit, not retried
  - [x] `python -m compileall -q .`
  - [x] `python -m py_compile scripts/import_lossless_claw.py`
  - [x] `bash -n scripts/install.sh scripts/update.sh`
  - [x] `git diff --check`
- [x] Workflow validation, if workflows changed: not applicable, no workflow files changed

## Notes
- Benchmark-only: no `preset: auto`, no `/lcm preset suggest`, no `/lcm preset apply`, no live-provider tuning, and no automatic live config mutation.
[...]

PR #195, opened 2026-05-22
feat: add inspectable preset surface

### LINKED ISSUE #189: Proposal: benchmark-driven model-aware LCM tuning presets
## Summary

LCM behavior depends heavily on the model and provider route in use: effective context length, token accounting, tool-calling reliability, summarizer quality, latency, and cost all change the right compression strategy.

Instead of expecting users to hand-tune every knob, we should build a benchmark-driven tuning harness that produces recommended LCM presets for model families.

The goal is not to tune every model manually. The goal is to ship sensible defaults, model-family presets, and diagnostics that tell users what to change when LCM is too noisy, too aggressive, or not recovering enough context.

## Problem

Current LCM tuning is mostly manual.

Users can run into symptoms like:

- compaction happening every turn
- compaction landing too close to the threshold
- too little post-compaction headroom
- stale context not being recovered unless retrieval is used correctly
- weak tool-calling models failing to take advantage of LCM retrieval
- strong tool-calling models carrying more active context than they need
- provider-specific context lengths not matching model marketing numbers
- noisy status messages that are expected during compaction but confusing when they appear frequently

A single default cannot be optimal across:

- Codex/OpenAI-style long-context routes
- OpenRouter models
- Anthropic models
- Kimi/GLM/DeepSeek/Qwen models
- local llama.cpp/vLLM models
- strong tool-calling models
- weaker tool-calling models
- cheap summarizer routes

## Proposed direction

Build a benchmark and tuning harness that can test LCM configurations against fixed transcript suites and produce recommended presets.

The harness should follow an autoresearch-style loop:

1. Start from a baseline LCM policy.
2. Try bounded changes to config and prompt/policy knobs.
3. Replay fixed transcript fixtures.
4. Score behavior with deterministic metrics.
5. Keep improvements and reject regressions.
6. Export a recommended preset.

This should produce model-family presets first, not per-model bespoke configs.

## Initial preset targets

Start with broad classes:

- `codex_gpt_long_context`
- `strong_tool_calling_long_context`
- `strong_tool_calling_medium_context`
- `weak_or_local_tool_calling`
- `cheap_summarizer_route`
- `high_latency_local_route`

Specific models/providers can map onto these presets over time.

## Candidate tuning knobs

The harness should evaluate knobs such as:

- context threshold
- post-compaction target/headroom
- hysteresis band
- minimum turns between compactions
- fresh tail count
- leaf chunk token size
- condensation fan-in/depth
- retrieval result limits
- summarizer/extractor/expander model route
- summary prompt variants
- provider/model token-count correction factor
- retrieval guidance for weaker tool-calling models

## Metrics

Do not optimize only for smallest context. That would produce worse agent behavior.

Score at least:

- compaction count per replay
- repeated compaction/chatter rate
- prompt tokens before and after compaction
- post-compaction headroom
- old fact recall
- exact canary recovery through `lcm_grep` and `lcm_expand`
- answer correctness on long-session questions
- retrieval tool calls required
- latency
- estimated token cost
- orphaned tool-result or code-output blocks
- false confidence when retrieval was required
- failures/exceptions

Separate scores for:

- active-context recall
- retrieval-backed recall
- cost/latency
- compaction stability

## Benchmark fixtures

Use a mix of:

- synthetic long conversations with planted canaries
- repeated compaction scenarios
- code-heavy sessions
- large tool outputs
- noisy irrelevant turns
- tool call/tool result groups
- sanitized real-world transcripts
- held-out validation transcripts

The tuning agent should not see held-out answer keys. We need to avoid Goodharting the benchmark.

## Cost controls

The default benchmark mode should be cheap and safe:

- deterministic/mock summarizers
- synthetic fixtures
- no live provider calls by default
- no live config mutation
- all writes isolated to a benchmark output directory

Live-model tuning should be opt-in and budget-capped.

Example:

```bash
hermes-lcm tune --model current --budget 5.00 --dry-run
```

For the first live tuning target, I can use my existing GPT/Codex route and current subscription. That lets us tune the path I actually use first without adding extra third-party inference spend.

Other providers should be BYOK or community-contributed initially.

## Community contribution path

Add a way for users to run benchmarks locally with their own provider credentials or local models and export scrubbed metrics.

Example:

```bash
hermes-lcm benchmark --model current --export lcm-benchmark-result.json
```

The export should include:

- provider/model name
- effective context length
- tested policy settings
- compaction metrics
- recall metrics
- latency/cost estimates
- recommended preset

It should not include raw transcript contents by default.

This lets the community contribute results for Anthropic, OpenRouter, local models, and other providers without maintainers collecting API keys or paying for every model.

## Funding note

Do not require donations to start this work.

Build the harness and run a small pilot first. If broader live-provider coverage proves useful and expensive, we can later add a specific sponsorship note:

> Sponsorship funds live-model benchmark runs for LCM preset tuning.

If we do that, results should be transparent: model tested, approximate cost, date, benchmark version, and resulting preset/docs PR.

## Non-goals

- Do not require normal users to run autonomous research loops.
- Do not auto-edit live user config without explicit approval.
- Do not collect user API keys.
- Do not tune every branded model individually at first.
- Do not optimize purely for context size at the expense of recall.
- Do not make LCM behavior depend on hidden maintainer-only benchmarks.

## Proposed deliverables

- Benchmark harness for LCM replay tests
- Synthetic and sanitized transcript fixture format
- Deterministic/mock summarizer mode
- Optional live-provider mode with budget caps
- Metrics export format
- Initial GPT/Codex long-context preset
- Model-family preset docs
- Symptom-based tuning guide
- Optional community benchmark export workflow

## Acceptance criteria

- A dry-run benchmark can run without live provider calls.
- The benchmark reports compaction stability, recall, latency, and headroom metrics.
- The harness can compare at least two LCM policies against the same replay suite.
- The first preset is generated for the current GPT/Codex long-context route.
- Documentation explains which symptoms map to which tuning knobs.
- Community benchmark exports are scrubbed by default.
- No live Hermes config is changed unless explicitly requested by the user.

## Open questions

- What is the minimum fixture set needed before results are trustworthy?
- Which metrics should be hard gates versus weighted scores?
- How should model-family presets be selected automatically?
- How much hysteresis should be built into LCM itself versus treated as config?
- Should benchmark exports include optional encrypted/private transcript artifacts for maintainers, or metrics only?
- What live-model budget is acceptable for the first public pilot?

stephenschoettler:
## Summary
- Add an inspect-only preset catalog for `codex_gpt_long_context` with benchmark provenance and dry-run application helpers.
- Add `/lcm preset show`, `/lcm preset suggest`, and `/lcm preset apply <name> --dry-run` without mutating live config.
- Add scrubbed community benchmark export JSON, plus docs for preset provenance, export workflow, and symptom-to-knob tuning.
- Validate preset-managed env overrides before honoring them, so invalid `LCM_*` values are reported instead of treated as active runtime overrides.
- Constrain `--export` to the benchmark output directory and refuse existing export targets or `.git/**` paths.

## Why
- Finishes the remaining #189 slices after the candidate policy landed in #194.
- Gives operators inspectable evidence before copying benchmarked settings.
- Keeps `preset: auto`, live-provider tuning, and automatic live config edits explicitly deferred.
- Keeps dry-run guidance aligned with `LCMConfig.from_env` fallback behavior when env values are invalid.
- Preserves the benchmark safety contract that generated artifacts stay isolated to the requested output directory.

## Validation
- [x] Focused validation: `python -m pytest tests/test_lcm_preset_command.py tests/test_benchmarking_cli.py tests/test_benchmarking_report.py tests/test_benchmarking_types.py -q` -> `29 passed`
- [x] Default validation:
  - [x] `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `580 passed`
  - [x] `python -m pytest -q` -> `791 passed`
  - [x] `prlimit --nofile=1024:1024 -- python -m pytest -q` -> `791 passed`

## Notes
- No workflow files changed, so `actionlint` was not run. Refreshed onto `origin/main` after #197 and re-ran validation on head `92f6f79`.
- The preset surface is inspect-only. `apply` requires `--dry-run` and only prints env-var suggestions.
- Explicit parseable preset-managed `LCM_*` environment variables win over preset defaults.
- Invalid preset-managed `LCM_*` values are reported as invalid and are not treated as active explicit overrides.
- Community exports are aggregate-only. The local `--json` benchmark summary remains diagnostic output and can include local paths.
- `--export` paths are now output-contained. Relative export names outside `--output` are resolved under `--output`; existing files and `.git/**` are refused.
- `target_after_compaction=0.55` remains benchmark provenance only because it is not a live runtime config field.
- Deferred by design: `preset: auto`, live-provider tuning, automatic live config edits, and silent runtime mutation.

Closes #189

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,8 +256,9 @@ safe shipped suggestion for the current context window when one exists, and
 labels the current selector as `context-only` when the provider/model family is
 not available to the plugin. `/lcm preset apply ... --dry-run` previews env-var
 settings only; it does not
-write files, change process state, or override explicit preset-managed `LCM_*`
-environment variables.
+write files, change process state, or override explicit parseable
+preset-managed `LCM_*` environment variables. Invalid preset-managed env values
+are reported as invalid instead of being treated as active runtime overrides.
 
 The current `codex_gpt_long_context` dry-run preview is:
 
--- a/README.md
+++ b/README.md
@@ -239,6 +239,37 @@ If startup/status output shows a host-side compression percentage that disagrees
 with LCM, trust live LCM status after a normal message has initialized the
 session.
 
+### Preset inspection and dry-run suggestions
+
+Model-aware presets are inspectable, but they are not automatic live config
+mutations. With `LCM_ENABLE_SLASH_COMMAND=true`, use:
+
+```text
+/lcm preset show codex_gpt_long_context
+/lcm preset suggest
+/lcm preset apply codex_gpt_long_context --dry-run
+```
+
+`/lcm preset show` reports the shipped preset metadata, benchmark provenance,
+policy file, policy version, and metric summary. `/lcm preset suggest` chooses a
+safe shipped suggestion for the current context window when one exists, and
+labels the current selector as `context-only` when the provider/model family is
+not available to the plugin. `/lcm preset apply ... --dry-run` previews env-var
+settings only; it does not
+write files, change process state, or override explicit preset-managed `LCM_*`
+environment variables.
+
+The current `codex_gpt_long_context` dry-run preview is:
+
+```text
+LCM_CONTEXT_THRESHOLD=0.75
+LCM_FRESH_TAIL_COUNT=24
+LCM_LEAF_CHUNK_TOKENS=8000
+```
+
+`target_after_compaction=0.55` is still benchmark provenance, not a runtime
+setting, because the engine does not expose that live knob yet.
+
 ### FAQ: tuning LCM for large context windows
 
 Long-context models change the tuning problem. A 1M-token model does not mean
--- a/benchmarking/report.py
+++ b/benchmarking/report.py
@@ -7,9 +7,9 @@ from collections import defaultdict
 from datetime import UTC, datetime
 from pathlib import Path
 from statistics import mean
-from typing import Any, Iterable
+from typing import Any, Iterable, Mapping
 
-from .types import ReplayMetrics
+from .types import LCMPolicy, ReplayMetrics
 
 BENCHMARK_VERSION = "2"
 FRESH_TAIL_PRESSURE_THRESHOLD = 0.30
@@ -177,3 +177,59 @@ def write_summary(path: str | Path, metrics: Iterable[ReplayMetrics]) -> dict[st
     output_path.parent.mkdir(parents=True, exist_ok=True)
     output_path.write_text(json.dumps(summary, indent=2, sort_keys=True) + "\n", encoding="utf-8")
     return summary
+
+
+def _policy_settings(policies: Iterable[LCMPolicy]) -> dict[str, dict[str, object]]:
+    settings: dict[str, dict[str, object]] = {}
+    for policy in policies:
+        data = policy.to_dict()
+        data.pop("notes", None)
+        settings[f"{policy.name}@{policy.policy_version}"] = data
+    return settings
+
+
+def build_community_export(
+    summary: Mapping[str, Any],
+    *,
+    policies: Iterable[LCMPolicy],
+    provider: str = "",
+    model: str = "",
+) -> dict[str, object]:
+    """Build a scrubbed benchmark result export for community sharing.
+
+    The export intentionally omits per-run metrics because those contain local
+    paths such as ``database_path`` and ``hermes_home``. It includes only the
+    aggregate comparison/provenance fields needed to reproduce and discuss a
+    preset result.
+    """
+
+    return {
+        "schema_version": "1",
+        "benchmark_version": summary.get("benchmark_version", BENCHMARK_VERSION),
+        "generated_at_utc": summary.get("generated_at_utc", _utc_timestamp()),
+        "provider": provider,
+        "model": model,
+        "transcript_contents_included": False,
+        "fixtures": list(summary.get("fixtures", [])),
+        "fixture_suite": list(summary.get("fixture_suite", [])),
+        "policies": list(summary.get("policies", [])),
+        "policy_versions": dict(summary.get("policy_versions", {})),
+        "policy_settings": _policy_settings(policies),
+        "metric_summary": dict(summary.get("metric_summary", {})),
+        "policy_comparison": list(summary.get("policy_comparison", [])),
+    }
+
+
+def write_community_export(
+    path: str | Path,
+    summary: Mapping[str, Any],
+    *,
+    policies: Iterable[LCMPolicy],
+    provider: str = "",
+    model: str = "",
+) -> dict[str, object]:
+    export = build_community_export(summary, policies=policies, provider=provider, model=model)
+    output_path = Path(path)
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    output_path.write_text(json.dumps(export, indent=2, sort_keys=True) + "\n", encoding="utf-8")
+    return export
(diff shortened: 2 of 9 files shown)


stephenschoettler:
Addressed the Codex P2 in `aaecc14`.

What changed:
- parseable preset-managed `LCM_*` env values are still treated as explicit overrides
- unparsable values are reported under `invalid_overrides` and are not honored as active runtime config
- dry-run preview now shows the preset value that would replace the invalid env value, plus the current parsed runtime value
- added regression coverage for invalid `LCM_FRESH_TAIL_COUNT=abc` through both `preset suggest` and `preset apply --dry-run`

Validation on the follow-up head:
- `python -m pytest tests/test_lcm_preset_command.py tests/test_benchmarking_cli.py tests/test_benchmarking_report.py tests/test_benchmarking_types.py -q` -> `27 passed`
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `579 passed`
- `python -m pytest -q` -> `757 passed`
- compile checks, shell syntax, and `git diff --check` passed

Tosko4 (changes requested):
Found one blocker before merge.

The new `--export` path is only checked against repo containment, then `write_community_export()` writes directly to that path. That means `--export README.md`, `--export plugin.yaml`, or even `.git/config` is accepted as long as the path is inside the repo, and the file gets overwritten with JSON. I reproduced this locally with `--export README.md`: the command completed and replaced the tracked README contents.

That does not match the benchmark safety claim that writes stay isolated to the requested output directory. I’d constrain `--export` under the benchmark output directory, or at least refuse existing files and reject `.git/**` unless there is an explicit overwrite flag.

The rest of the review looked good: the preset surface is dry-run only, invalid preset-managed env values are no longer treated as active overrides, the scrubbed export omits local paths and transcript content, and the focused/full/fd-limited suites passed locally.

Tosko4:
@stephenschoettler

stephenschoettler:
Thanks, agreed. Pushed `eed6430` to address the export safety blocker.

What changed:
- `--export` is now constrained under `--output`
- existing export targets and `.git/**` are refused before replay/output writes
- added regressions for `README.md` overwrite, `.git/config`, and outside-output export paths
- refreshed the PR body with current-head validation

Validation on `eed6430`:
- `python -m pytest tests/test_lcm_preset_command.py tests/test_benchmarking_cli.py tests/test_benchmarking_report.py tests/test_benchmarking_types.py -q` -> `29 passed`
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `579 passed`
- `python -m pytest -q` -> `759 passed`
- `prlimit --nofile=1024:1024 -- python -m pytest -q` -> `759 passed`
- compile checks, shell syntax, and `git diff --check` passed
- GitHub CI is green: workflow-lint and test 3.11/3.12/3.13/3.14

Tosko4:
@codex review

Tosko4 (approved):
Re-reviewed current head 92f6f79 after the branch was refreshed onto latest main. The export containment blocker I raised is fixed now: export targets stay under the benchmark output directory, existing targets and `.git` paths are refused, and the inspect-only preset surface still matches the PR body.

Validation on the live head: focused preset and benchmark tests passed locally, the default and full suites passed locally including the fd-limited run, GitHub CI is green, there are no unresolved review threads, and the fresh Codex review on this head came back clean.

Approving from my side.

stephenschoettler:

PR #201, opened 2026-05-24
fix: register LCM tools via ctx.register_tool() so lcm_* appear in agent toolset

### LINKED ISSUE #200: Plugin advertises 7 lcm_* tools in provides_tools but register() never calls ctx.register_tool() — tools missing from agent
## Summary

The plugin's `plugin.yaml` declares 7 tools under `provides_tools:`, and `tools.py` implements them — but `__init__.py:register(ctx)` never calls `ctx.register_tool(...)` for any of them. As a result, none of the LCM tools are present on the agent at runtime, on either Linux/systemd or macOS/launchd hosts.

This appears to be a regression-from-birth: `git log -S "register_tool"` returns zero commits across the entire history. The `register()` function only registers a context engine (`ctx.register_context_engine(engine)`) and, when `LCM_ENABLE_SLASH_COMMAND=1`, the `/lcm` slash command. Per the design philosophy laid out in #129, those tools are part of the LCM "context-engine contract" — without them the model sees compacted summaries with recovery hints, but the recovery path itself isn't wired up.

## Reproduction

Tested on two independent Hermes installs:

- **Linux** (Hetzner VPS, hermes-agent latest, hermes-lcm v0.11.1, systemd-managed): `hermes tools list` shows no `lcm_*` tools. Only built-in toolsets plus `hermes-cli` plugin toolset and configured MCP servers appear.
- **macOS** (Mac Mini, hermes-agent fork @ 729a778af, hermes-lcm v0.11.1, launchd-managed): identical result.

In both cases the plugin loads cleanly (`hermes plugins ls` shows it as `enabled`, 0.11.1, `git` source) and the DAG context engine is active.

## Evidence (from installed plugin tree)

`plugin.yaml`:

\`\`\`yaml
provides_tools:
  - lcm_grep
  - lcm_load_session
  - lcm_describe
  - lcm_expand
  - lcm_expand_query
  - lcm_status
  - lcm_doctor
\`\`\`

`__init__.py:register(ctx)` (full body):

\`\`\`python
def register(ctx):
    config = LCMConfig.from_env()
    hermes_home = ...  # resolved
    engine = LCMEngine(config=config, hermes_home=hermes_home)

    ctx.register_context_engine(engine)

    register_command = getattr(ctx, "register_command", None)
    slash_enabled = _env_flag_enabled("LCM_ENABLE_SLASH_COMMAND", default=False)
    if callable(register_command) and slash_enabled:
        from .command import handle_lcm_command
        register_command("lcm", lambda raw_args: handle_lcm_command(raw_args, engine),
                         description="LCM status and diagnostics")
    # ... no register_tool() call anywhere
\`\`\`

`grep -RnE "ctx\\.register_tool|register_tool\\(" plugins/hermes-lcm/` → 0 hits.

`tools.py:1003` defines `def lcm_describe(args, **kwargs) -> str: ...`, and (per `provides_tools`) the other six handlers exist in the same module. They are simply never bound into the registry.

## Expected behavior

Per the official plugin guide (`build-a-hermes-plugin.md`) the `register()` function should wire each schema/handler pair via `ctx.register_tool(name=..., toolset=..., schema=..., handler=...)`. Without this, `provides_tools:` is metadata only — the agent never sees the tools.

## Suggested fix

Add a `schemas.py` (or co-locate schemas with handlers in `tools.py`) and extend `register()`:

\`\`\`python
def register(ctx):
    # ... existing engine + slash command code ...
    from . import tools, schemas
    for name in ("lcm_grep", "lcm_load_session", "lcm_describe",
                 "lcm_expand", "lcm_expand_query", "lcm_status", "lcm_doctor"):
        ctx.register_tool(
            name=name,
            toolset="lcm",
            schema=getattr(schemas, name.upper()),
            handler=getattr(tools, name),
        )
\`\`\`

Happy to send a PR if helpful — wanted to file the issue first to confirm the intended surface (e.g., toolset name, whether any of the 7 should be gated behind a config flag).

## Why this matters

Per the discussion on #129, LCM's compaction relies on the agent being able to recover detail through these tools. Right now agents on every install of v0.11.1 see compacted summaries with recovery hints that point at tools the agent doesn't have. That's the failure mode #129 was explicitly trying to prevent.

BowmanStephen:
## Summary

Fixes #200 — the plugin's `plugin.yaml` declared 7 `lcm_*` tools, but `register()` never called `ctx.register_tool()`. As a result, the tools were invisible to the global tool registry and only accessible through the ContextCompressor fallback path, which does not run for all agent configurations.

## Changes

- Added `_make_wrapped_handler()` to bind the `LCMEngine` instance into each `lcm_*` handler.
- Extended `__init__.py:register()` to import all 7 schemas and handlers, then call `ctx.register_tool(name, toolset="lcm", schema=..., handler=...)` for each.
- The context engine registration (`ctx.register_context_engine()`) and slash command registration are unchanged.

## Tool Registration Table

| Tool | Schema | Handler |
|------|--------|---------|
| `lcm_grep` | `LCM_GREP` | `lcm_tools.lcm_grep` |
| `lcm_load_session` | `LCM_LOAD_SESSION` | `lcm_tools.lcm_load_session` |
| `lcm_describe` | `LCM_DESCRIBE` | `lcm_tools.lcm_describe` |
| `lcm_expand` | `LCM_EXPAND` | `lcm_tools.lcm_expand` |
| `lcm_expand_query` | `LCM_EXPAND_QUERY` | `lcm_tools.lcm_expand_query` |
| `lcm_status` | `LCM_STATUS` | `lcm_tools.lcm_status` |
| `lcm_doctor` | `LCM_DOCTOR` | `lcm_tools.lcm_doctor` |

## Verification

- `python3 -c "import ast; ast.parse(open('__init__.py').read())"` passes (syntax OK).
- Manual review confirms all 7 names in `plugin.yaml` `provides_tools` are now wired.

## Follow-up (deferred)

- Item 2 from the #200 debt list: `engine.handle_tool_call()` is still called by `tool_executor.py` via `agent.context_compressor.handle_tool_call()`. This path is still the primary dispatch for context-engine tools. Removing it is a larger refactor that needs the test suite passing and maintainer sign-off — deferred.
- Item 3: full `pytest tests/` run needs a local hermes-agent environment with correct PYTHONPATH.
- Item 6: removing `get_tool_schemas()`/`handle_tool_call()` from engine should only happen after registry routing is proven stable across all configs.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -1,3 +1,4 @@
+
 """Hermes LCM Plugin — Lossless Context Management.
 
 Replaces the built-in ContextCompressor with a DAG-based context engine
@@ -19,10 +20,27 @@ def _env_flag_enabled(name: str, default: bool = False) -> bool:
     return value.strip().lower() in {"1", "true", "yes", "on"}
 
 
+def _make_wrapped_handler(handler, engine):
+    """Wrap a raw lcm_* handler so kwargs['engine'] is always bound."""
+    def _wrapped(args, **kwargs):
+        return handler(args, engine=engine, **kwargs)
+    return _wrapped
+
+
 def register(ctx):
-    """Plugin entry point — register the LCM context engine."""
+    """Plugin entry point — register the LCM context engine and tools."""
     from .config import LCMConfig
     from .engine import LCMEngine
+    from . import tools as lcm_tools
+    from .schemas import (
+        LCM_GREP,
+        LCM_LOAD_SESSION,
+        LCM_DESCRIBE,
+        LCM_EXPAND,
+        LCM_EXPAND_QUERY,
+        LCM_STATUS,
+        LCM_DOCTOR,
+    )
 
     config = LCMConfig.from_env()
 
@@ -39,6 +57,27 @@ def register(ctx):
 
     # Register as the context engine (replaces ContextCompressor)
     ctx.register_context_engine(engine)
+
+    # Register tools via the plugin registry so they are discoverable
+    # by the global tool system (not just the context-engine fallback).
+    _TOOLS = {
+        "lcm_grep": LCM_GREP,
+        "lcm_load_session": LCM_LOAD_SESSION,
+        "lcm_describe": LCM_DESCRIBE,
+        "lcm_expand": LCM_EXPAND,
+        "lcm_expand_query": LCM_EXPAND_QUERY,
+        "lcm_status": LCM_STATUS,
+        "lcm_doctor": LCM_DOCTOR,
+    }
+    for name, schema in _TOOLS.items():
+        handler = getattr(lcm_tools, name)
+        ctx.register_tool(
+            name=name,
+            toolset="lcm",
+            schema=schema,
+            handler=_make_wrapped_handler(handler, engine),
+            description=schema.get("description", f"LCM tool: {name}"),
+        )
 
     register_command = getattr(ctx, "register_command", None)
     slash_enabled = _env_flag_enabled("LCM_ENABLE_SLASH_COMMAND", default=False)


stephenschoettler:

Please add:
- Why
- Validation

stephenschoettler (changes requested):
Thanks. I’m holding CI approval on this head because the current diff has a load-time blocker.

Please guard `ctx.register_tool` the same way `/lcm` guards `register_command`. I reproduced the host/plugin-collection shape that exposes `register_context_engine` but not `register_tool`:

- current `main`: `register(ctx)` succeeds and registers the engine
- this branch: `AttributeError: 'Ctx' object has no attribute 'register_tool'`

That means the plugin can fail to load entirely on hosts or collection contexts where the context-engine path would otherwise keep working.

One more thing to preserve before this becomes the dispatch path: Hermes currently passes live `messages=...` through `agent.context_compressor.handle_tool_call(...)` for `lcm_*` calls. Direct registry handlers do not receive that argument today, so make sure this PR does not shadow the context-engine dispatch path and lose current-turn ingest.

After that, please rebase/update from `main` and fill in `Why` and `Validation`. I’ll approve the fork workflow on the next head.

BowmanStephen:
Thanks for the review. Closing this for now — I won't be able to get to the requested changes in the near term. Happy for anyone to pick it up from here.

PR #206, opened 2026-05-25
Add structured focus briefs

ai-ag2026:
## Summary

Adds a small structured Focus brief for summary prompts when `focus_topic` is supplied:

- normalizes and bounds focus-topic text before injecting it into prompts
- gives L1 summarization a concrete preservation contract for focus-related decisions, constraints, files, commands, identifiers, and current state
- gives L2 summarization an aggressive bullet-preservation contract while keeping unrelated blockers/handoff state
- preserves existing empty-focus and custom-instructions behavior

This is intentionally prompt-only and provider-free: it does not change runtime storage, DAG shape, model routing, or benchmark policy files.

## Why

The current focus guidance is a single ad-hoc line. A structured, bounded Focus brief gives the summarizer clearer retention priorities for focus-related context while reducing prompt-injection and noise risk from long or multi-line topics.

## Validation

- [x] RED verified: the new Focus-brief tests failed against the existing ad-hoc focus guidance.
- [x] `python3 -m pytest tests/test_lcm_core.py::TestEscalation -q`
- [x] `python3 -m compileall -q escalation.py tests/test_lcm_core.py`
- [x] `git diff --check`
- [x] Added-line scans for secret-like strings, private/local markers, shell/eval/pickle/SQL patterns: 0 hits

## Notes

`tests/test_lcm_engine.py` still does not collect in this standalone checkout because the host `agent.context_engine` module is absent. This PR is scoped to `escalation.py`, so the local gate uses the escalation prompt tests directly.

PR #207, opened 2026-05-25
Improve budget-aware context assembly

ai-ag2026:
## Summary
- Make active-context assembly skip over-budget assistant/tool tail turns instead of stopping tail selection before older prompt-bearing user context can fit.
- Make summary selection skip an over-budget summary part and keep scanning for later summaries that fit the remaining budget.
- Add focused regression tests and update assembly guardrail tests to document the intended contract change: user turns remain prompt-bearing stop points; derived assistant/tool turns may be evicted when they block budgetable context.

## Why
- A single bulky assistant/tool turn or summary can otherwise evict all older budgetable context behind it, including the current user objective or smaller summaries that would still fit.
- User turns are treated as prompt-bearing stop points; derived assistant/tool material is allowed to be evicted when it blocks useful fit candidates.
- This keeps the behavior deterministic and local to assembly while preserving hard caps and explicit overflow failure reporting for irreducible user-prompt overflows.

## Validation
- [x] RED: `python3 -m pytest tests/test_lcm_core.py::TestAssemblyBudgetSelection -q` -> 2 failed on `origin/main` before the implementation (`KEEP_USER_DECISION` / `SMALL_RECENT_SUMMARY` absent).
- [x] Focused review-fix validation: `python3 -m pytest tests/test_lcm_core.py::TestAssemblyBudgetSelection tests/test_lcm_engine.py::TestAssemblyGuardrails::test_max_assembly_tokens_caps_recent_tail tests/test_lcm_engine.py::TestAssemblyGuardrails::test_reserve_tokens_floor_caps_recent_tail tests/test_lcm_engine.py::TestAssemblyGuardrails::test_max_assembly_tokens_skips_droppable_assistant_gap_with_varied_sizes tests/test_lcm_engine.py::TestAssemblyGuardrails::test_summary_budget_skips_oversized_summary_and_keeps_later_fit_part tests/test_lcm_engine.py::TestAssemblyGuardrails::test_max_assembly_tokens_drops_oversized_newest_assistant_and_keeps_user_prompt tests/test_lcm_engine.py::TestAssemblyGuardrails::test_forced_overflow_tail_capping_updates_bookkeeping_without_middle_compaction tests/test_lcm_engine.py::TestAssemblyGuardrails::test_forced_overflow_recovery_flags_irreducible_single_tail_overflow tests/test_lcm_engine.py::TestAssemblyGuardrails::test_overflow_recovery_failure_flag_resets_after_successful_compression tests/test_packaging_install.py::test_plugin_entrypoint_registers_lcm_context_engine -q` -> 11 passed.
- [x] Hosted-CI mirror validation after rebasing on `origin/main`: create the same minimal `agent.context_engine` stub as `.github/workflows/ci.yml`, then `ulimit -n 1024 && python3 -m pytest tests/ -q` -> 804 passed, 56 warnings.
- [x] Syntax validation: `python3 -m compileall -q engine.py tests/test_lcm_core.py tests/test_lcm_engine.py` -> passed.
- [x] Script validation: `python3 -m py_compile scripts/import_lossless_claw.py` -> passed.
- [x] Shell validation: `bash -n scripts/install.sh scripts/update.sh` -> passed.
- [x] Whitespace validation: `git diff --check` -> passed.
- [x] Added-line secret scan: regex scan over `git diff origin/main...HEAD --unified=0` -> only false-positive `token` substrings in test names/config fields; no credentials or secret values found.
- [ ] Hosted GitHub Actions after review-fix force-push: queued/running on `bbb399e`.

## Notes
- This branch has been rebased on current `origin/main` after #206 landed.
- No runtime provider calls, new env vars, or external services are introduced.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -2775,6 +2775,20 @@ class LCMEngine(ContextEngine):
                 and len(candidate_messages) >= raw_session_count
                 and raw_session_count > 1
             )
+            has_preserved_objective_scaffold = any(
+                str(msg.get("role") or "") != "system"
+                and (normalize_content_value(msg.get("content")) or "").lstrip().startswith(
+                    _PRESERVED_OBJECTIVE_CONTEXT_PREFIX
+                )
+                for msg in candidate_messages
+            )
+            candidate_suffix_has_user_turn = any(identity[0] == "user" for identity in candidate_prefix)
+            has_scaffold_suffix_replay = (
+                matches_sanitized_tail
+                and has_preserved_objective_scaffold
+                and not candidate_suffix_has_user_turn
+                and cursor < len(messages)
+            )
             has_raw_cleanup_replay = (
                 matches_raw_tail
                 and has_scaffold_evidence
@@ -2782,7 +2796,7 @@ class LCMEngine(ContextEngine):
                 and len(candidate_prefix) >= max(1, self._config.fresh_tail_count)
                 and raw_suffix_needs_cleanup_equivalence
             )
-            if has_effective_full_replay or has_raw_full_replay or has_raw_cleanup_replay:
+            if has_effective_full_replay or has_raw_full_replay or has_scaffold_suffix_replay or has_raw_cleanup_replay:
                 return cursor
         return empty_prefix_cursor if allow_empty_prefix else None
 
@@ -3749,6 +3763,8 @@ class LCMEngine(ContextEngine):
 
         tail_selected = tail_messages
         anchor_source = getattr(self, "_pending_context_anchor_messages", None)
+        if anchor_source is None:
+            anchor_source = tail_messages
         anchor_part: Optional[str] = None
         summary_budget = None
         if assembly_cap is not None:
@@ -3759,11 +3775,15 @@ class LCMEngine(ContextEngine):
                 tail_messages,
                 insert_missing_tool_stubs=False,
             )
+            skipped_tail_gap = False
             for msg in reversed(tail_for_selection):
                 msg_tokens = count_message_tokens(msg)
                 if used + tail_token_total + msg_tokens > assembly_cap:
                     if self._is_budget_droppable_tail_message(msg):
+                        skipped_tail_gap = True
                         continue
+                    break
+                if skipped_tail_gap and kept_tail_reversed:
                     break
                 kept_tail_reversed.append(msg)
                 tail_token_total += msg_tokens
--- a/engine.py
+++ b/engine.py
@@ -2787,7 +2787,6 @@ class LCMEngine(ContextEngine):
                 matches_sanitized_tail
                 and has_preserved_objective_scaffold
                 and not candidate_suffix_has_user_turn
-                and cursor < len(messages)
             )
             has_raw_cleanup_replay = (
                 matches_raw_tail
--- a/engine.py
+++ b/engine.py
@@ -3782,7 +3782,7 @@ class LCMEngine(ContextEngine):
                         skipped_tail_gap = True
                         continue
                     break
-                if skipped_tail_gap and kept_tail_reversed:
+                if skipped_tail_gap:
                     break
                 kept_tail_reversed.append(msg)
                 tail_token_total += msg_tokens
--- a/engine.py
+++ b/engine.py
@@ -3750,6 +3750,8 @@ class LCMEngine(ContextEngine):
             for msg in reversed(tail_for_selection):
                 msg_tokens = count_message_tokens(msg)
                 if used + tail_token_total + msg_tokens > assembly_cap:
+                    if self._is_budget_droppable_tail_message(msg):
+                        continue
                     break
                 kept_tail_reversed.append(msg)
                 tail_token_total += msg_tokens
@@ -3797,7 +3799,7 @@ class LCMEngine(ContextEngine):
                     if count_message_tokens(candidate_msg) > summary_budget:
                         if part == anchor_part:
                             continue
-                        break
+                        continue
                     selected_parts.append(part)
             if selected_parts:
                 combined = "\n\n---\n\n".join(selected_parts)
@@ -3832,6 +3834,24 @@ class LCMEngine(ContextEngine):
             result = self._sanitize_active_context_messages(trimmed_result)
 
         return result
+
+    def _is_budget_droppable_tail_message(self, message: Dict[str, Any]) -> bool:
+        """Return whether an over-budget tail message may be evicted.
+
+        User turns are prompt-bearing context and stop tail selection when they
+        cannot fit. Assistant/tool turns are derived context; if one bulky turn
+        blocks older prompt material, skip it and keep scanning for budgetable
+        user intent or compact status that still fits.
+        """
+        role = message.get("role")
+        if role not in {"assistant", "tool"}:
+            return False
+        content = normalize_content_value(message.get("content")) or ""
+        if _PRESERVED_TODO_CONTEXT_PREFIX in content:
+            return False
+        if _PRESERVED_OBJECTIVE_CONTEXT_PREFIX in content:
+            return False
+        return True
 
     def _finalize_forced_overflow_result(
         self,
--- a/tests/test_lcm_core.py
+++ b/tests/test_lcm_core.py
@@ -3152,6 +3152,86 @@ class TestEscalation:
         assert "Additional instructions:" not in l2
 
 
+class TestAssemblyBudgetSelection:
+    def _engine(self, tmp_path: Path, monkeypatch, *, max_assembly_tokens: int = 120):
+        if "agent.context_engine" not in sys.modules:
+            agent_mod = ModuleType("agent")
+            agent_mod.__path__ = []
+            context_engine_mod = ModuleType("agent.context_engine")
+
+            class ContextEngine:
+                def __init__(self, **kwargs):
+                    self.compression_count = 0
+                    self.last_prompt_tokens = 0
+
+                def get_status(self):
+                    return {}
+
+            setattr(context_engine_mod, "ContextEngine", ContextEngine)
+            monkeypatch.setitem(sys.modules, "agent", agent_mod)
+            monkeypatch.setitem(sys.modules, "agent.context_engine", context_engine_mod)
+
+        from hermes_lcm.engine import LCMEngine
+
+        config = LCMConfig(
+            database_path=str(tmp_path / "assembly.db"),
+            max_assembly_tokens=max_assembly_tokens,
+        )
+        engine = LCMEngine(config=config, hermes_home=str(tmp_path / "hermes"))
+        engine._session_id = "assembly-session"
+        return engine
+
+    def test_assembly_skips_oversized_assistant_turn_to_preserve_user_prompt(self, tmp_path, monkeypatch):
+        engine = self._engine(tmp_path, monkeypatch, max_assembly_tokens=120)
+        huge_assistant = "oversized assistant tool chatter " * 400
+
+        assembled = engine._assemble_context(
+            {"role": "system", "content": "System anchor."},
+            [
+                {"role": "user", "content": "KEEP_USER_DECISION: continue with prompt-aware assembly."},
+                {"role": "assistant", "content": huge_assistant},
+                {"role": "assistant", "content": "Latest compact status."},
+            ],
+        )
+
+        contents = "\n".join(str(msg.get("content", "")) for msg in assembled)
+        assert "KEEP_USER_DECISION" in contents
+        assert "Latest compact status" in contents
+        assert "oversized assistant tool chatter" not in contents
+
+    def test_assembly_skips_oversized_summary_and_keeps_later_fit_summary(self, tmp_path, monkeypatch):
+        engine = self._engine(tmp_path, monkeypatch, max_assembly_tokens=140)
+        engine._dag.add_node(SummaryNode(
+            session_id="assembly-session",
+            depth=2,
+            summary="HUGE_DURABLE_SUMMARY " * 400,
+            token_count=800,
+            source_token_count=2000,
+            source_ids=[1],
+            source_type="messages",
+            expand_hint="durable huge",
+        ))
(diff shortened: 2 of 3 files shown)


stephenschoettler (dismissed):
Changes requested.

Hosted CI is red on `3e1b16d`, so this needs to be fixed before review routing.

Key failure classes from the failed `test (3.11)` job:
- compression tests no longer hit the patched summarizer path, for example `test_compress_accepts_focus_topic` and adaptive leaf rescue tests
- assembly guardrail expectations changed around tail contiguity and summary-budget ordering
- `test_plugin_entrypoint_registers_lcm_context_engine` now hits `AttributeError: 'super' object has no attribute 'get_status'`

Because this PR intentionally changes assembly selection, please either preserve the current guardrail/contiguity contract or update that contract with explicit tests and rationale. I am not routing this while hosted CI is red.

ai-ag2026:
Addressed the red hosted CI on the previous head.

Changes made:
- Fixed the test isolation issue from `TestAssemblyBudgetSelection` so it no longer replaces `hermes_lcm.engine` or leaves an `agent` module stub in `sys.modules` for later tests.
- Updated the existing assembly guardrail tests to make the intended contract change explicit: prompt-bearing user turns remain stop points; derived assistant/tool turns and over-budget summary parts may be skipped when they block later budgetable context.
- Rebasing included current `main` after #206 landed.

Local validation on the new head `bbb399e`:
- `python3 -m pytest ...review-fix focused selection... -q` -> 11 passed
- CI mirror with the same minimal `agent.context_engine` stub and `ulimit -n 1024 && python3 -m pytest tests/ -q` -> 804 passed, 56 warnings
- `python3 -m compileall -q engine.py tests/test_lcm_core.py tests/test_lcm_engine.py` -> passed
- `python3 -m py_compile scripts/import_lossless_claw.py` -> passed
- `bash -n scripts/install.sh scripts/update.sh` -> passed
- `git diff --check` -> passed

Hosted Actions are now rerunning on the updated head.

ai-ag2026:
Follow-up on the earlier `CHANGES_REQUESTED` review: the branch was updated after that review, and hosted CI is now green on the current head `bbb399e` across workflow-lint and Python 3.11-3.14.

No additional code changes in this comment; just calling out that the original blocker was addressed and the PR is ready for re-review.

Tosko4 (changes requested):
I agree with the problem this is trying to fix, and the direction is close: bulky assistant or tool material should not be able to crowd out the current user objective. But I do not think this implementation is safe to merge yet.

The new tail selection can return a non-contiguous active tail made of real durable messages. In the repro I used, compaction returned the older `role=user` objective and the latest assistant status while dropping the oversized assistant message between them. That looks good for the immediate prompt, but after a restart `_reconcile_ingest_cursor_from_store()` cannot match that active context to a durable suffix, so the old user and assistant turns get persisted again as new rows.

Concrete repro shape:

```python
messages = [
    {"role": "system", "content": "sys"},
    {"role": "user", "content": "old compactable question"},
    {"role": "assistant", "content": "old compactable answer"},
    {"role": "user", "content": "KEEP_USER_DECISION: continue prompt-aware assembly"},
    {"role": "assistant", "content": "oversized assistant tool chatter " * 400},
    {"role": "assistant", "content": "Latest compact status."},
]
```

With deterministic test tokenization and summarization, `max_assembly_tokens=450`, `fresh_tail_count=3`, and `leaf_chunk_tokens=1`, the returned active context is:

```text
system: sys + LCM note
assistant: [Recent Summary ...]
user: KEEP_USER_DECISION: continue prompt-aware assembly
assistant: Latest compact status.
```

After creating a new engine on the same DB and ingesting that active context plus one new user message, the store grew by 3 rows instead of 1. The reconciliation state was `persisted ambiguous delta`, and both the old user objective and latest assistant status were duplicated.

That is the blocker for me. This changes active prompt selection in a way that is locally useful, but it breaks the lossless store and restart idempotency contract. Strategically, this needs to stay aligned with the existing anchor/scaffold design: if an older user objective is pulled forward across compacted or skipped material, it should be carried as active-context scaffolding or otherwise reconciled explicitly, not replayed as a real non-contiguous raw tail message.

The PR’s focused tests pass locally, and GitHub Actions are green on `38978a5`, so this is not a CI issue. The missing coverage is the restart/re-ingest invariant for the new non-contiguous tail shape.

stephenschoettler:
Thanks @Tosko4. I reproduced this on current head `38978a5`, so I am keeping this blocked.

The failure mode I see is the same restart/re-ingest contract break:

- Compaction can now return a non-contiguous durable tail after skipping the oversized assistant turn: `system + KEEP_USER_DECISION + Latest compact status`
- After restart, ingesting that active context plus one new user message grew the store from 6 rows to 9 rows. Expected growth was 1 row.
- The old `KEEP_USER_DECISION: continue prompt-aware assembly` user turn and `Latest compact status.` assistant turn were duplicated.
- Reconciliation recorded `advanced cursor` / `skipped scaffold-only prefix` with `cursor=1`, which only skipped the system LCM scaffold and then persisted the pulled-forward durable messages again.

Please fix this before re-review. The fix needs to preserve the lossless store/restart idempotency contract, either by carrying pulled-forward older user intent as active-context scaffolding or by teaching reconciliation this non-contiguous shape explicitly. Add a focused regression for this restart case, not just the immediate prompt-selection behavior.

ai-ag2026:
Addressed the restart/re-ingest blocker on the current head `1572220`.

Changes made:
- Preserved older user objectives are now emitted as active-context scaffolding when a droppable assistant/tool gap would otherwise create a non-contiguous raw durable tail.
- Restart reconciliation now treats that preserved-objective scaffold plus a matching durable suffix as replay evidence, without broadening generic summary/system-note scaffolding; existing repeated-tail delta tests remain intact.
- Added the focused regression for the reproduced shape: system + summary scaffold + preserved `KEEP_USER_DECISION` + latest status + one new user grows the durable store by exactly one row, not by duplicated pulled-forward rows.

Local validation:
- `python3 -m pytest tests/test_lcm_core.py::TestAssemblyBudgetSelection -q` -> 3 passed
- CI mirror with minimal `agent.context_engine` stub: `python3 - <<'PY' ... pytest.main(['tests/', '-q']) ... PY` -> 816 passed, 57 warnings
- `python3 -m compileall -q engine.py tests/test_lcm_core.py tests/test_lcm_engine.py` -> passed
- `python3 -m py_compile scripts/import_lossless_claw.py` -> passed
- `bash -n scripts/install.sh scripts/update.sh` -> passed
- `git diff --check` -> passed

ai-ag2026:
Follow-up on the preserved-objective replay fix; new head is `3753249`.

I found one over-broad edge in the previous head: preserved-objective scaffold plus a new user turn identical to the durable tail could be treated as replay and skipped. This update tightens that path so preserved-objective scaffold suffix replay only applies when the matching suffix contains no user turns. Assistant/tool suffix replay still works for the restart idempotency case, while repeated user deltas are preserved.

Added regression:
- `test_preserved_objective_scaffold_does_not_skip_new_repeated_user_tail`

Validation:
- Focused replay/assembly regressions -> `4 passed, 28 warnings`
- Full standalone suite with minimal `agent.context_engine` stub -> `817 passed, 248 warnings`
- `python3 -m compileall -q engine.py tests/test_lcm_core.py tests/test_lcm_engine.py`
- `python3 -m py_compile scripts/import_lossless_claw.py`
- `git diff --check`
- Added-line secret/private-marker scan -> clean
- Independent focused reviewer: no blockers; noted the mixed user suffix path now conservatively preserves data rather than deduping.

Tosko4 (changes requested):
I rechecked the current head `3753249`. The new restart case with `active_context + one new user` is fixed, but the same preserved-objective shape still is not idempotent when the restarted process ingests the active context by itself.

Repro shape is the one from the earlier blocker: persisted `system, old user, old assistant, KEEP_USER_DECISION, oversized assistant, latest status`, then assemble the preserved-objective scaffold plus `Latest compact status.`. If a new engine on the same DB ingests just that active context, the durable store grows from 6 rows to 7 and `Latest compact status.` is duplicated.

The reconciliation state I see is:

```text
advanced cursor / skipped scaffold-only prefix, cursor=2, incoming=3, effective_incoming=1
```

So the scaffold and one suffix row are skipped, but the final durable status row is treated as a new delta. That is the same lower-level problem in a narrower path: a replayed active context should be safe to ingest before any new delta arrives.

The focused branch tests and full suite pass locally, including the rebased shape on latest `main` (`822 passed`). The remaining issue is missing coverage for preserved-objective active replay with no appended user/tool delta. I think this should add a regression for `after_restart._ingest_messages(active_context)` and make that path grow by zero rows before this merges.

ai-ag2026:
Addressed the preserved-objective replay no-delta idempotency blocker reported on `3753249`; new head is `3e97879`.

What changed:
- Added a regression to replay the assembled preserved-objective active context by itself after restart and assert the durable store does not grow.
- Kept the existing appended-delta regression for `active_context + new user`.
- Kept the inverse guard for repeated user deltas: preserved-objective scaffold replay still does **not** apply when the matching suffix contains a user turn.
- Minimal code change: scaffold suffix replay no longer requires an appended delta (`cursor < len(messages)`), because a pure replay should also be idempotent.

Validation:
- RED before fix: focused preserved-objective replay test failed with `AssertionError: assert 7 == 6`.
- Focused regressions: `2 passed, 28 warnings`.
- `TestAssemblyBudgetSelection`: `4 passed, 28 warnings`.
- Full suite with standalone `agent.context_engine` stub: `817 passed, 248 warnings`.
- Hygiene: `compileall`, `py_compile scripts/import_lossless_claw.py`, `bash -n scripts/install.sh scripts/update.sh`, `git diff --check`, and added-line secret/private-marker scan all clean.

Tosko4 (changes requested):
I rechecked current head `3e97879`. The no-delta preserved-objective replay blocker I reported on `3753249` is fixed, but there is still a sibling non-contiguous raw-tail replay path.

Minimal shape:

```python
messages = [
    {"role": "system", "content": "sys"},
    {"role": "user", "content": "repeat user intent"},
    {"role": "assistant", "content": "huge assistant output " * 400},
]
```

With `max_assembly_tokens=160`, `_assemble_context(messages[0], messages[1:])` returns:

```text
system: sys + LCM note
user: repeat user intent
```

The oversized newest assistant is skipped, but the older durable user turn is emitted as a raw tail message. After restart, ingesting that active context plus one new user grows the durable store from 3 rows to 5 rows instead of 4. `repeat user intent` is duplicated.

Observed reconciliation:

```text
advanced cursor / skipped scaffold-only prefix, cursor=1, incoming=3, effective_incoming=2
```

So this is the same contract issue in the “oversized newest assistant” branch: the immediate prompt keeps the user objective, but restart replay cannot prove the raw user turn is a durable suffix because a skipped assistant row sits after it in storage.

Evidence:

```text
active: system + raw user repeat user intent
before: 3
after: 5
expected: 4
counts: repeat user intent = 2, new user after restart = 1
```

The existing changed test `test_max_assembly_tokens_drops_oversized_newest_assistant_and_keeps_user_prompt` appears to encode this raw-user behavior, but it does not cover restart/re-ingest. I think this path needs the same treatment as the preserved-objective gap case: if retaining an older user across a droppable assistant/tool gap would make the active context non-contiguous, carry it as preserved-objective scaffolding or otherwise make reconciliation explicitly idempotent. Add a focused restart regression for this exact newest-assistant shape before merge.

Validation I ran:

```text
current PR head focused tests: 4 passed, 14 passed
current PR head full suite: 817 passed, 1 warning
manual previous no-delta repro: fixed, 6 rows stayed 6
manual previous appended-delta repro: fixed, 6 rows grew to 7
rebased onto latest upstream/main: clean rebase, 822 passed, 1 warning
manual newest-assistant repro: fails on both current PR head and rebased shape
```

ai-ag2026:
Addressed the sibling non-contiguous raw-tail replay blocker reported on current head `3e97879`; new head is `18af0a6`.

What changed:
- Added a focused regression for the minimal shape from the review: `system, user repeat intent, oversized newest assistant`, then restart ingest of `active_context + one new user`.
- Tightened assembly tail selection so once an over-budget assistant/tool gap is skipped, older durable tail rows are not emitted as raw conversation messages. The latest omitted user intent is still preserved through the existing preserved-objective scaffold path, which keeps restart ingestion from duplicating it.
- Updated the existing newest-oversized-assistant guardrail expectation to assert the user objective remains present but is no longer returned as a raw `role=user` tail row when a later assistant/tool gap was skipped.

Validation:
- RED before fix: the new focused regression failed because `repeat user intent` was returned as a raw user row and duplicated after restart.
- Focused replay/guardrail tests: `6 passed, 28 warnings`.
- Full suite with standalone `agent.context_engine` stub: `818 passed, 248 warnings`.
- Hygiene: `compileall`, `py_compile scripts/import_lossless_claw.py`, `bash -n scripts/install.sh scripts/update.sh`, `git diff --check`, and added-line secret/private-marker scan all clean.

Tosko4 (approved):
Re-reviewed `18af0a6`. The restart/idempotency blockers I raised are fixed now. The newest oversized assistant case no longer brings the older user objective back as a raw durable tail; it is carried as preserved-objective scaffold, so replaying the active context with a new user grows by exactly one real row and does not duplicate the older user.

I also reran the earlier preserved-objective status replay cases and a realistic summary-scaffold path through `compress()`. Those stayed idempotent.

Validation on current head:

```text
TestAssemblyBudgetSelection + TestAssemblyGuardrails: 19 passed
full suite: 818 passed, 1 warning
manual newest-assistant replay: 3 -> 4, repeat user intent count stayed 1
manual preserved-objective/status replay: 6 -> 7, Latest compact status count stayed 1
rebase onto latest upstream/main: clean
rebased focused tests: 19 passed
rebased full suite: 823 passed, 1 warning
```

Approving from my side.

Tosko4:
@stephenschoettler can you review mate?

Tosko4 (approved):

stephenschoettler:

PR #212, opened 2026-05-27
fix(store): serialize MessageStore writes on threading.RLock

spcoombe:
## Summary
- Serialize all `MessageStore` write paths on a new `threading.RLock` (`self._write_lock`). Defense-in-depth against on-disk corruption observed in deployments running heavy concurrent in-process HTTPS API clients alongside LCM writes.

## Why

`MessageStore` opens its SQLite connection with `check_same_thread=False` and shares it across all threads in the host process. SQLite's own C-level mutex protects engine-internal state, but Python's `sqlite3` module releases the GIL while the C call runs. Under heavy thread contention with concurrent HTTPS API clients in the same process, downstream operators have observed on-disk corruption.

### Observed signature

Two independent LCM databases in the same deployment, corrupted at different times, presented an identical byte pattern in the first 32 bytes of the file:

```
Healthy SQLite:  53 51 4c 69 74 65 20 66 6f 72 6d 61 74 20 33 00   "SQLite format 3\0"
Both corrupt:    53 51 4c 69 74 17 03 03 00 13 XX XX XX XX XX XX
                 └─SQLit─┘ └────── TLS 1.2 Application Data ──────┘
```

- Bytes 0-4 (`SQLit`) — intact
- Bytes 5-9 (`17 03 03 00 13`) — exact TLS 1.2 ContentType.application_data record header with length=19
- Bytes 10-28 — 19 bytes of TLS ciphertext payload (differs between the two corrupt files, as expected for different sessions)
- Bytes 29+ — original SQLite header intact in both cases

SQLite cannot produce these bytes. The only source in the affected process at the time of corruption was a Python HTTPS client (`openai`/`httpx`). The pattern is consistent with concurrent in-process TLS buffer state intersecting SQLite's write/flush path while the GIL was released during a C-level SQLite write.

Both databases were fully recoverable by splicing a canonical 100-byte SQLite header back in — the SQLite pages themselves were untouched, only the file header. Row counts and `PRAGMA integrity_check` matched pre-corruption snapshots exactly.

## Fix

Add `self._write_lock = threading.RLock()` to `MessageStore.__init__` and wrap every call site that mutates `self._conn`:

- `append`
- `append_batch`
- `reassign_session_messages`
- `delete_session_messages`
- `gc_externalized_tool_result`
- `pin`
- `unpin`
- `normalize_legacy_blank_sources`

The lock is re-entrant so nested write call sites are safe. Where existing `with self._conn:` transaction managers are present, the new lock is acquired alongside (`with self._write_lock, self._conn:`) so the SQLite-level transaction and the Python-level mutex enter/exit together.

Single-threaded callers see no behavior change; the lock adds one uncontended `RLock.acquire`/`release` pair per write operation.

`unpin` had no `commit()` previously, while `pin` did — `commit()` is now consistent across both. Functional behavior is unchanged because SQLite autocommit on simple `UPDATE` matches what `commit()` does explicitly, but the explicit `commit` makes the symmetry obvious.

## Out of scope

`LifecycleStateStore` in `lifecycle_state.py` uses the same shared `check_same_thread=False` connection pattern and would benefit from an identical lock. Keeping that as a follow-up so this PR stays small and easy to review per `CONTRIBUTING.md`.

## Validation

- [x] Focused validation: `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` → **588 passed** (includes new `test_write_lock_serializes_concurrent_appends`)
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` → 588 passed
  - [x] `pytest -q` → 817 passed, 2 failures in `test_ingest_protection.py` (pre-existing on `main` HEAD `b4a64e4`, unrelated to this change — verified by running them against unpatched main with the patch stashed)
  - [x] `bash -lc 'ulimit -n 1024 && pytest -q'` (same result as above)
  - [x] `python -m compileall -q .` clean
  - [x] `python -m py_compile scripts/import_lossless_claw.py` clean
  - [x] `bash -n scripts/install.sh scripts/update.sh` clean
  - [x] `git diff --check` clean
- [ ] Workflow validation, if workflows changed: N/A (no workflow changes)

### New test

`tests/test_lcm_core.py::TestMessageStore::test_write_lock_serializes_concurrent_appends`

- asserts `MessageStore._write_lock` exists and is re-entrant
- spins up 8 worker threads doing 25 single appends + 5 batches of 5 each
- asserts no exceptions raised, exact expected row count (400 rows), and `PRAGMA quick_check = ok` afterward

## Notes

This is defense-in-depth. The exact race that caused the observed TLS-byte injection is not nailed down to a single line of CPython, but the fix removes one whole class of risk (Python-side concurrent writers to the same `sqlite3.Connection`) with negligible performance cost. The full forensic write-up of the incident is in the deploying organization's internal records — happy to share more detail if maintainers want it.

### DIFF
--- a/store.py
+++ b/store.py
@@ -213,6 +213,23 @@ class MessageStore:
         self._ingest_protection_config = ingest_protection_config or LCMConfig(database_path=str(self.db_path))
         self._hermes_home = hermes_home or str(self.db_path.parent)
         self._conn: Optional[sqlite3.Connection] = None
+        # ``self._conn`` is shared across threads (the connection is opened with
+        # ``check_same_thread=False``). SQLite's own C-level mutex serializes
+        # statements at the engine layer, but the Python ``sqlite3`` module
+        # releases the GIL while the C call runs. Under heavy thread contention
+        # with concurrent HTTPS clients in the same process, downstream
+        # operators have observed on-disk corruption that is consistent with
+        # external bytes landing inside SQLite's write path (e.g. the first
+        # 28 bytes of the database file replaced with a TLS record header +
+        # ciphertext while the "SQLit" magic remains intact).
+        #
+        # This re-entrant lock is defense-in-depth: it forces all write call
+        # sites that use ``self._conn`` to be serialized at the Python layer,
+        # eliminating any window where Python-side buffer reuse or memory
+        # aliasing could intersect SQLite's flush of a write. It does not
+        # change semantics for single-threaded callers and adds only a single
+        # uncontended ``RLock.acquire``/``release`` pair per operation.
+        self._write_lock = threading.RLock()
         self._init_db()
 
     def _init_db(self):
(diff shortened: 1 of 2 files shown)


stephenschoettler:

stephenschoettler:
Thanks. CI is green on `ec9ee82`, and I've asked @Tosko4 for review. If @Tosko4 approves and checks stay green, I'll merge this.

Tosko4 (approved):
Reviewed current head `ec9ee82` against current `main`.

The branch is now based on the latest `main`, satisfying Stephen's earlier rebase request. The diff is narrow: `MessageStore` gets a re-entrant Python write lock around the mutating `self._conn` paths, plus a focused concurrency regression. This fits the project direction as a small defensive hardening change in the plugin's own storage layer, without changing host/plugin ownership or broad runtime behavior.

I treated the TLS-byte forensic detail as motivation rather than something this PR has to prove line-by-line. The merge criterion here is narrower and the patch meets it: all normal `MessageStore` write paths on the shared `check_same_thread=False` connection are serialized, existing transaction-manager writes keep transaction scope, and the read/query behavior is otherwise unchanged. Keeping `LifecycleStateStore` for a follow-up is reasonable for this PR size.

Validation:

```text
new focused regression: 1 passed
tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py: 600 passed, 1 warning
full suite: 841 passed, 1 warning
low-fd full suite: 841 passed, 1 warning
compileall / import_lossless_claw py_compile / install+update shell syntax: pass
GitHub Actions on current head: green across workflow-lint and Python 3.11-3.14
```

Scope/security check: no workflow, dependency, install-script, network, credential, persistence-location, or external execution surface changes. Only SQLite store write serialization and its regression test.

Approving from my side.

Tosko4:
@stephenschoettler can you review mate?

stephenschoettler:

PR #209, opened 2026-05-26
feat: register LCM diagnostic tools with Hermes host

ruangraung:
## Summary

Register all 7 LCM diagnostic tools (`lcm_grep`, `lcm_load_session`, `lcm_describe`, `lcm_expand`, `lcm_expand_query`, `lcm_status`, `lcm_doctor`) as Hermes agent tools in the plugin entrypoint (`__init__.py`).

The `plugin.yaml` manifest declares these 7 tools under `provides_tools`, but the `register(ctx)` function only called `ctx.register_context_engine()` and optionally registered the `/lcm` slash command — the tools themselves were never exposed to the Hermes agent host via `ctx.register_tool()`.

This change:
- Imports all 7 tool schemas from `schemas.py`
- Registers each tool under the `context_engine` toolset with `ctx.register_tool()`
- Routes each handler through `engine.handle_tool_call()` to preserve the engine's message-ingest step (current-turn search), as required by the context-engine dispatch path
- Guards `register_tool` via `getattr(ctx, 'register_tool', None)` so the plugin loads gracefully on hosts or collection contexts without this method — same pattern used by the `/lcm` slash command guard
- Updates test `_Ctx` stubs to support the new `register_tool()` interface

**Note:** These tools also require `context_engine` to be listed in `platform_toolsets.<platform>` where they should appear (e.g., `- context_engine` under `platform_toolsets.telegram`).

## Why

`plugin.yaml` and the README both advertise 7 agent tools, but the runtime entrypoint never registered them with the Hermes host. Running `hermes plugins` showed `hermes-lcm v0.12.0 (7 tools)` from the manifest, but the tools were absent from the agent's tool schema, making them unavailable to the model. See issue #200.

This is a complementary fix to the host-side PR [`NousResearch/hermes-agent#31194`](https://github.com/NousResearch/hermes-agent/pull/31194) (which auto-enables the `context_engine` toolset for non-default context engines). The plugin-side registration ensures tools are available even on Hermes Agent versions that predate the host-side fix.

## Validation

- [x] Focused validation: `pytest tests/test_packaging_install.py::test_plugin_manifest_lists_all_registered_tools -q` -> `1 passed`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` — `166 passed`
  - [x] `pytest -q` — `783 passed`
  - [x] `python -m compileall -q .` — no errors
[...]

Issue #168, opened 2026-05-13
Make preflight compression messaging context-engine aware for LCM


barronlroth:
## Summary

When `hermes-lcm` is the selected context engine (`context.engine: lcm`) and LCM is functioning normally, Hermes can still surface the generic user-facing status message:

```text
📦 Preflight compression...
```

That wording is confusing in an LCM-backed session because it implies the built-in lossy compressor is running or that something is wrong, even when LCM is healthy and doing the right thing.

## Observed behavior

In a Telegram gateway session with LCM active:

- `plugins.enabled` includes `hermes-lcm`
- `context.engine: lcm`
- `compression.enabled: true` remains enabled as the host-level compaction gate
- `lcm_status` reports:
  - `engine: lcm`
  - `plugin_version: 0.10.4`
  - `context_threshold: 0.75`
  - `fresh_tail_count: 128`
- `lcm_doctor` passes the important integrity checks:
  - database integrity
  - FTS integrity
  - DAG orphan check
  - config validation
  - context pressure

Despite this, the user can still see the host's generic `Preflight compression` status before/around LCM handling. In practice that reads like a warning, even though LCM is the selected context engine and this may simply be the normal preflight path.

## Expected behavior

When the selected context engine is LCM, user-facing preflight messaging should avoid implying lossy/built-in compression unless that is actually happening.

Possible acceptable outcomes:

1. Suppress the generic `📦 Preflight compression...` user-facing status when `context.engine: lcm` and LCM reports a healthy/no-op/handled state.
2. Replace it with LCM-specific wording, e.g.:
   - `Checking LCM context budget...`
   - `LCM compacting older context...`
   - `LCM context maintenance...`
3. Expose a plugin-side integration/status hook so Hermes core can choose a context-engine-specific preflight label and avoid emitting the generic compression text for plugin engines.

## Why this matters

LCM's value proposition is that context management is lossless/recoverable via the LCM store and DAG. A generic `Preflight compression` message is operationally misleading for users/operators because it suggests:

- Hermes may have fallen back to built-in compression
- the session may be repeatedly compressing due to a fault
- LCM is not actually active

The correct diagnosis requires checking `lcm_status` and `lcm_doctor`, which is more operator work than necessary.

## Environment where observed

```text
Hermes Agent: v0.13.0 (2026.5.7)
hermes-lcm: v0.10.4
Platform: Telegram gateway
Config:
  plugins.enabled: [hermes-lcm]
  context.engine: lcm
[...]

PR #217, opened 2026-05-29
fix: prevent compression cascade on session split

ruangraung:
## Summary
- Repair compression-boundary skip handling so an unproven session split does not reuse stale `_last_compacted_store_id` or `_ingest_cursor` from the previously bound session.
- Bind the new session against its own lifecycle frontier, schedule ingest-cursor reconciliation, and keep fresh-session messages ingestible.
- Share the boundary-skip cooldown between `should_compress()` and `should_compress_preflight()` so the mutating preflight path cannot immediately re-trigger compression.
- Update/add rollover regressions for stale cursor/frontier loss, lossless preflight ingest during cooldown, and preflight replay-diff cooldown suppression.

## Why
- The cascade bug is real, but the earlier fix preserved progress from a session that had not been proven to be the same logical continuation.
- That could avoid the cascade by skipping fresh messages for the new session, which breaks the lossless-store invariant.
- The documented host path is preflight, so the cooldown has to cover `should_compress_preflight()`, not just `should_compress()`.

## Validation
- [x] Focused validation: `python -m pytest tests/test_lcm_engine.py::TestSessionRollover -q` -> `65 passed`
- [x] Focused validation: `python -m pytest tests/test_lcm_engine.py -k 'compression_boundary or rollover_session or preflight' -q` -> `26 passed, 368 deselected`
- [x] Default validation:
  - [x] `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `605 passed`
  - [x] `python -m pytest -q` -> `854 passed`
  - [x] `bash -lc 'ulimit -n 1024 && python -m pytest -q'` -> `854 passed`
- [x] Workflow validation, if workflows changed: not applicable, PR diff does not change `.github/workflows/*`.

## Notes
- Maintainer follow-up pushed on top of the contributor branch after Tosko4's review.
- The original PR direction is preserved, but the unsafe stale-progress preservation is replaced with fresh/unproven-session progress reset plus cursor reconciliation.
- Remote head is now based on current `main` after #220/#221.
- This also addresses the state/boundary half of #168. The remaining user-visible status behavior stays tracked upstream in NousResearch/hermes-agent#25115.

Refs #209
Refs #168


Tosko4 (changes requested):
I do not think this is safe to merge yet. The bug is real, but the current fix preserves state from the previously bound session in the exact path where the code has decided that bound session is not a safe carry-over source.

1. In `_continue_compression_boundary()`, the mismatch path at `engine.py:1763` to `engine.py:1797` can set `source_session_id = ""` and `source_state = None` because the bound session did not match `old_session_id` and could not be proven safe. The new skip branch then saves `_last_compacted_store_id` and `_ingest_cursor` from that bound session, binds the new session, and restores those values at `engine.py:1872` to `engine.py:1873`.

That can make the next ingest skip fresh messages for the new session. I reproduced it locally by binding `session-a`, setting `_ingest_cursor = 3`, taking the compression-boundary skip into `session-b`, then ingesting two fresh `session-b` messages:

```text
session_id= session-b
restored_cursor= 3
last_compacted_store_id= 42
session_b_store_count= 0
session_b_rows= []
```

So the patch can avoid the cascade by not ingesting the new session's messages at all. That is worse than the original symptom for the lossless store.

2. The cooldown is currently only in `should_compress()`. The documented host path is `should_compress_preflight()`, and that path still ingests messages and can return `True` during the cooldown window. Local repro with `_last_boundary_skip_time = time.time()`:

```text
should_compress= False
should_compress_preflight= True
store_count_after_preflight= 10
```

So the safety net does not cover the path that actually decides and mutates during preflight.

I ran the focused rollover suite and the full local suite. They pass, but the new tests assert the stale cursor preservation directly, so they are not proving the invariant we need here. I would expect the fix to only preserve progress when the source session is proven to be the same logical continuation, otherwise bind the new session with its own frontier and reconcile its cursor before trusting it. If cooldown stays, it also needs to cover preflight, not just `should_compress()`.

stephenschoettler:
Pushed maintainer follow-up `a60d54e` for the two review blockers:

- unproven boundary skip now resets stale compaction progress and schedules cursor reconciliation for the new session
- cooldown now covers `should_compress_preflight()` as well as `should_compress()`

Validation is green on the new head. GitHub CI passed, and local validation included full suite plus low-FD full suite (`854 passed` for both). I also refreshed the PR body to match the current diff and re-requested Tosko4 review.

Tosko4 (approved):
Looks good to merge. My previous changes request is cleared on the current head.

I re-ran the two blocker repros against `a60d54e`:
- unproven compression-boundary skip no longer reuses stale `_last_compacted_store_id` / `_ingest_cursor`; fresh `session-b` messages are persisted under the new session
- cooldown now covers `should_compress_preflight()` as well as `should_compress()`: preflight still ingests fresh messages, but does not immediately request another compression during the boundary-skip window, including the replay-diff path

Validation:
- `python -m pytest tests/test_lcm_engine.py::TestSessionRollover -q` -> 65 passed
- `python -m pytest tests/test_lcm_engine.py -k 'compression_boundary or rollover_session or preflight' -q` -> 26 passed, 368 deselected
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> 605 passed, with only the existing discord `audioop` deprecation warning
- `python -m pytest -q` -> 854 passed, with only the same warning
- `git diff --check upstream/main...HEAD` -> passed

I also reviewed the `.gitignore` supply-chain surface. It is broad local-artifact hygiene, but I do not see a merge blocker there. No dependency, workflow, install, or runtime supply-chain surface changes.

This fits the LCM-side state/boundary follow-up for the repeated compression cascade. The remaining user-visible preflight/status wording is still host-side work, as tracked separately.

stephenschoettler:

PR #225, opened 2026-05-31
test: add lifecycle soak stress scenario

stephenschoettler:
## Summary
- Adds the deterministic `lifecycle_soak_and_profile_rebinds` stress scenario for `/new` rollover, restart/rebind, Hermes home profile rebinding, SQLite WAL growth, and externalized payload integrity.
- Adds a `soak` stress tier with lifecycle-specific knobs for longer manual durability runs.
- Documents the lifecycle-only soak command in the benchmark README.
- Tightens the review blockers: rollover now fails when no summary nodes carry into the new session/current scope, and externalized payload checks verify the planted canary/value through `lcm_expand(externalized_ref=...)`.

## Why
- P0 gave us deterministic release smoke coverage. P1 extends that into lifecycle durability boundaries where context can be lost across session rollover, process restart, profile rebinds, or accumulated externalized payloads.
- The scenario stays offline, deterministic, and sandboxed under the requested output directory.
- The follow-up makes the scenario fail on the two lifecycle regressions called out in review instead of only recording them.

## Validation
- [x] Review regression proof: focused test first failed on missing `rollover_0_current_scope_probe`, then passed after the scenario recorded and enforced the new assertions.
- [x] Focused validation: `python -m pytest tests/test_stress_release_check.py -q` -> `7 passed`
- [x] Stress/benchmarking slice: `python -m pytest tests/test_stress_release_check.py tests/test_benchmarking_cli.py tests/test_benchmarking_replay.py -q` -> `27 passed`
- [x] Lifecycle smoke: `python scripts/lcm_stress_check.py --output /tmp/hermes-lcm-pr225-reviewfix-smoke-20260531-234106-27264 --tier smoke --scenario lifecycle_soak_and_profile_rebinds --json` -> `failure_count: 0`, `rollover_count: 1`, `externalized_payload_integrity_checked: 4/4`, `old_canary_current_scope_ok: true`
- [x] Lifecycle release: `python scripts/lcm_stress_check.py --output /tmp/hermes-lcm-pr225-reviewfix-release-20260531-234106-27264 --tier release --scenario lifecycle_soak_and_profile_rebinds --json` -> `failure_count: 0`, `rollover_count: 3`, `externalized_payload_integrity_checked: 24/24`, `old_canary_current_scope_ok: true`
- [x] Lifecycle soak: `python scripts/lcm_stress_check.py --output /tmp/hermes-lcm-pr225-reviewfix-soak-20260531-234106-27264 --tier soak --scenario lifecycle_soak_and_profile_rebinds --json` -> `failure_count: 0`, `rollover_count: 5`, `externalized_payload_integrity_checked: 120/120`, `old_canary_current_scope_ok: true`
- [x] Default validation:
  - [x] `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `605 passed`
  - [x] `python -m pytest -q` -> `861 passed`
  - [x] `prlimit --nofile=1024:1024 -- python -m pytest -q` -> `861 passed`
  - [x] `python -m py_compile scripts/import_lossless_claw.py scripts/lcm_stress_check.py` -> passed

## Notes
- No workflow files changed, so `actionlint` was not run.
- Low-fd validation used `prlimit --nofile=1024:1024` instead of the shell `ulimit` form from the template.
- Stress artifacts were written under `/tmp/...`; live Hermes profile/config state was not mutated.
- Related PR search found no open or closed exact overlaps for `lifecycle soak`, `profile rebind`, or `stress`.
- Related: #223

### DIFF
--- a/benchmarking/stress.py
+++ b/benchmarking/stress.py
@@ -828,6 +828,8 @@ def _case_lifecycle_soak_and_profile_rebinds(run: StressRun) -> None:
     def lifecycle_messages_for_session(session_idx: int) -> list[dict[str, Any]]:
         return [{"role": "system", "content": f"lifecycle system anchor session={session_idx}"}]
 
+    expected_externalized_payloads: dict[str, str] = {}
+
     def append_turn(messages: list[dict[str, Any]], session_idx: int, turn_idx: int) -> None:
         cid = f"CANARY_LIFECYCLE_{session_idx:02d}_{turn_idx:03d}"
         val = f"VALUE_LIFECYCLE_{session_idx:02d}_{turn_idx:03d}"
@@ -836,6 +838,7 @@ def _case_lifecycle_soak_and_profile_rebinds(run: StressRun) -> None:
             "content": f"session={session_idx} turn={turn_idx} {cid} = {val} lifecycle restart rollover payload " + ("alpha " * 20),
         })
         if turn_idx % max(1, run.tier.lifecycle_externalize_every) == 0:
+            expected_externalized_payloads[cid] = val
             tool_call_id = f"lifecycle-call-{session_idx}-{turn_idx}"
             blob = base64.b64encode((f"{cid} {val} " + ("PAYLOAD " * 220)).encode()).decode()
             messages.append({
@@ -899,6 +902,33 @@ def _case_lifecycle_soak_and_profile_rebinds(run: StressRun) -> None:
                 )
                 rollovers += 1
                 run.record(case, f"rollover_{session_idx}_carried_nodes", carried)
+                if carried <= 0:
+                    run.fail(
+                        case,
+                        "rollover_carry_over_missing",
+                        "Lifecycle rollover carried no summary nodes into the new session",
+                        {"old_session": old_session, "new_session": new_session, "carried_nodes": carried},
+                    )
+                current_scope_ok, current_scope_probe = _tool_recall_contains(
+                    run,
+                    engine,
+                    "CANARY_LIFECYCLE_00_000",
+                    "CANARY_LIFECYCLE_00_000",
+                    "VALUE_LIFECYCLE_00_000",
+                    args={"limit": 10},
+                )
+                run.record(
+                    case,
+                    f"rollover_{session_idx}_current_scope_probe",
+                    {"ok": current_scope_ok, "probe": current_scope_probe},
+                )
+                if not current_scope_ok:
+                    run.fail(
+                        case,
+                        "rollover_immediate_current_scope_lost_early_canary",
+                        "Lifecycle rollover lost early canary from the new session's default/current scope",
+                        {"old_session": old_session, "new_session": new_session, "probe": current_scope_probe},
+                    )
                 current_session = new_session
                 messages = lifecycle_messages_for_session(session_idx + 1)
         messages = engine.compress(messages, current_tokens=max(4_200, count_messages_tokens(messages)))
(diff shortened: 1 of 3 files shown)


Tosko4 (changes requested):
Requesting changes because the new stress scenario is useful, but the assertions do not yet fail on two of the lifecycle regressions it is meant to catch.

Directionally this is good: offline, deterministic, sandboxed, and focused on rollover, restart, profile rebinds, WAL growth, and large-output payloads. That is the right shape for P1 lifecycle soak coverage.

Two gaps should be tightened before this becomes a release gate.

1. `/new` carry-over is recorded but not enforced

`benchmarking/stress.py:893-903` records `rollover_0_carried_nodes`, but nothing fails if it is zero.

The later old-canary checks at `benchmarking/stress.py:907-938` use `session_scope: "all"` and explicit old-session scope. Those can still pass from raw old-session storage even if rollover carry-over into the new/current session regresses. For a lifecycle/session-drift soak, the important failure mode is that old context is not available from the new active session after `/new` unless the caller explicitly searches all or the old session.

Please add one of these:

* fail when `carry_over_context=True` returns zero carried nodes for this scenario
* probe the early canary from the current/default scope after rollover while bound to the new session
* assert a carried node/result exists under the new session

The focused test should assert that too, not only `rollover_count >= 1`.

2. Externalized payload integrity only checks shape, not integrity

`benchmarking/stress.py:940-953` verifies that payload JSON files exist, parse, and have non-empty `content` plus `session_id`. That is a presence/metadata check, not payload integrity.

A regression that writes the wrong payload, truncates the expected canary/value, or stores unrelated content would still pass as long as `content` is non-empty.

Since the PR summary explicitly says this covers externalized payload integrity, the scenario should verify the planted lifecycle canary/value in the externalized content, ideally through the normal externalized-ref recovery path rather than only direct JSON reads.

Validation I ran locally on `f0d86bd`:

```text
python -m pytest tests/test_stress_release_check.py -q
7 passed

python scripts/lcm_stress_check.py --output /tmp/hermes-lcm-pr225-lifecycle-smoke-mainreview --tier smoke --scenario lifecycle_soak_and_profile_rebinds --json
failure_count: 0
```

The harness runs cleanly. The issue is that it can still go green while missing the specific lifecycle and payload regressions it claims to guard.

stephenschoettler:
Updated in e427eaf.

Addressed both blockers:
- Carry-over now fails if `rollover_session(... carry_over_context=True)` carries zero nodes, and the scenario probes the old canary from the new session's default/current scope after rollover plus at final state.
- Payload integrity now matches every externalized payload to its planted canary/value and verifies recovery through `lcm_expand(externalized_ref=...)` while bound to the payload's session.

Validation is refreshed in the PR body. Key gates: `tests/test_stress_release_check.py` 7 passed, stress/benchmarking slice 27 passed, lifecycle smoke/release/soak all `failure_count: 0` with payload integrity checked `4/4`, `24/24`, and `120/120`, full suite 861 passed including low-fd.

Re-requesting review.

Tosko4 (approved):
The follow-up commit addresses my previous blockers.

The lifecycle soak now fails when `/new` carries no nodes, and it checks the early canary from the new/current scope instead of only proving all-scope or explicit old-session recall. The externalized payload path also now verifies planted canary/value integrity and exercises `lcm_expand` against each externalized ref.

I re-reviewed the updated diff on `e427eaf` and did not find new scope or supply-chain concerns. This remains an isolated deterministic stress-harness addition, with no runtime `engine.py`/`config.py` behavior changes and no workflow/dependency changes.

Local validation:

```text
python -m pytest tests/test_stress_release_check.py -q
7 passed

python scripts/lcm_stress_check.py --output /tmp/hermes-lcm-pr225-rereview-lifecycle-smoke --tier smoke --scenario lifecycle_soak_and_profile_rebinds --json
failure_count: 0

python -m pytest tests/test_stress_release_check.py tests/test_benchmarking_cli.py tests/test_benchmarking_replay.py -q
27 passed, 1 warning

python -m pytest -q
861 passed, 1 warning

python -m compileall -q .
python -m py_compile scripts/import_lossless_claw.py scripts/lcm_stress_check.py
passed
```

GitHub checks are green on the current head. Approved.

stephenschoettler:

PR #226, opened 2026-06-01
fix: preserve active replay after storage protection

de-snake:
## Summary
- Preserve provider-facing active replay for structured image/media payloads while still externalizing those payloads before SQLite storage.
- Preserve provider-facing active replay for tool-result media payloads while keeping stored rows protected.
- Keep the existing compact active stub behavior for whole-message generic `raw_payload` externalization.
- Prevent media-only storage protection from making repeated preflight checks request no-op compaction.

## Why
- Storage-boundary protection should keep large inline media/base64 out of SQLite, FTS, WAL, and backups.
- That storage protection must not corrupt the active context sent back to providers. Image and tool-result media turns need their original payloads in active replay.
- Generic oversized raw text is different: the established behavior is to return the compact externalized stub in active context so the host does not keep replaying a huge opaque payload while durable storage contains only the protected row.
- This distinguishes those two contracts instead of treating every storage-protection rewrite as storage-only.

## Validation
- [x] RED before follow-up fix: `python -m pytest -q tests/test_lcm_core.py::TestIngestExternalization::test_compress_returns_externalized_stub_for_oversized_active_tail` failed with active replay returning `ACTIVE_RAW_NEEDLE` instead of the `raw_payload` stub.
- [x] Focused raw-payload regression: `python -m pytest -q tests/test_lcm_core.py::TestIngestExternalization::test_compress_returns_externalized_stub_for_oversized_active_tail` -> 1 passed
- [x] Focused media regressions: `python -m pytest -q tests/test_ingest_protection.py::test_ingest_preserves_provider_active_context_while_protecting_storage tests/test_ingest_protection.py::test_tool_result_ingest_preserves_active_context_while_protecting_storage tests/test_ingest_protection.py::test_preflight_storage_protection_does_not_force_noop_compaction` -> 3 passed
- [x] Targeted PR/review slice: `python -m pytest tests/test_ingest_protection.py tests/test_lcm_engine.py -q` -> 491 passed
- [x] Default validation:
  - [x] `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> 609 passed
  - [x] `python -m pytest -q` -> 868 passed
  - [x] `prlimit --nofile=1024:1024 -- python -m pytest -q` -> 868 passed
  - [x] `python scripts/lcm_stress_check.py --output "$OUT" --tier release --json` -> `failure_count: 0`, 6 cases
  - [x] `python -m py_compile scripts/import_lossless_claw.py scripts/lcm_stress_check.py` -> passed
  - [x] `git diff --check origin/main...HEAD` -> passed

## Notes
- Maintainer follow-up pushed in `e25ad61` after merging current `origin/main` into the PR branch.
- The follow-up changes only active replay selection after storage protection. Durable storage externalization semantics are unchanged.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -3142,15 +3142,14 @@ class LCMEngine(ContextEngine):
             estimates,
             source=self._session_platform,
         )
-        active_replay_messages = list(replay_messages)
-        for (absolute_idx, replay_msg), protected_msg in zip(messages_to_store_with_index, protected_messages):
-            active_msg = dict(protected_msg)
-            if "content" not in replay_msg and active_msg.get("content") in (None, ""):
-                active_msg.pop("content", None)
-            active_replay_messages[absolute_idx] = active_msg
         self._ingest_cursor = n
         logger.debug("Ingested %d messages into LCM store", len(messages_to_store_with_index))
-        return active_replay_messages
+        # ``protected_messages`` are storage-only: they may replace inline media,
+        # tool outputs, or base64 substrings with externalized placeholders. The
+        # provider-facing active replay must keep usable original payloads (after
+        # active-only quarantine/redaction above), otherwise image/tool turns can
+        # be corrupted and storage protection alone can trigger no-op compaction.
+        return replay_messages
 
     def _get_store_ids_for_messages(self, messages: List[Dict[str, Any]]) -> List[int]:
         """Map current raw messages back to store_ids in stable store order.
--- a/engine.py
+++ b/engine.py
@@ -3212,6 +3212,18 @@ class LCMEngine(ContextEngine):
             config=self._config,
             hermes_home=self._hermes_home,
         )
+        active_replay_messages = replay_messages
+        for (absolute_idx, _replay_msg), protected_msg in zip(
+            messages_to_store_with_index,
+            protected_messages,
+        ):
+            if self._protected_message_uses_raw_payload_active_stub(protected_msg):
+                if active_replay_messages is replay_messages:
+                    active_replay_messages = [dict(message) for message in replay_messages]
+                active_message = dict(active_replay_messages[absolute_idx])
+                active_message["content"] = protected_msg["content"]
+                active_replay_messages[absolute_idx] = active_message
+
         estimates = [count_message_tokens(m) for m in protected_messages]
         self._store.append_batch(
             self._session_id,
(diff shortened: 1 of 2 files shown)


stephenschoettler:

Please add:
- Why

Tosko4 (changes requested):
Thanks for the focused patch. I agree with the bug this is trying to fix: storage-boundary externalization for inline media/base64 should not corrupt provider-facing active replay. The new image/tool-result tests point at the right failure mode, and I did not find any suspicious supply-chain or hidden-execution surface in the diff.

I do need to request changes because the implementation is broader than that contract and breaks an existing active-context invariant for generic oversized raw payloads.

The change in `engine.py` now always returns `replay_messages` after `protect_messages_for_ingest(...)`, making every ingest-protection rewrite storage-only. That is correct for provider-usable media/image payloads, but not for the existing raw-payload fallback. For large opaque text payloads, the current behavior intentionally returns the compact externalized stub in active context so the host does not keep replaying a huge raw payload while SQLite/FTS stores only the protected row.

This now fails on the current PR head:

```text
python -m pytest -q tests/test_lcm_core.py::TestIngestExternalization::test_compress_returns_externalized_stub_for_oversized_active_tail
FAILED tests/test_lcm_core.py::TestIngestExternalization::test_compress_returns_externalized_stub_for_oversized_active_tail
```

The assertion that fails:

```text
assert active_content.startswith("[Externalized payload: kind=raw_payload;")
```

Observed active content is the original `ACTIVE_RAW_NEEDLE:` payload instead of the compact raw-payload stub.

I also ran the broader suite:

```text
python -m pytest -q
1 failed, 863 passed, 1 warning
```

The targeted PR tests do pass, and I also verified the intended media boundary manually:

```text
active_has_payload: true
store_has_payload: false
mapped_ids: [1, 2, 3]
ref_count: 2
store_expand_has_tool_calls_key: false
store_expand_has_raw_payload: false
```

So the direction is good, but the fix needs to distinguish active replay semantics by payload class instead of treating all storage protection as storage-only. Suggested shape:

* keep provider-facing originals for structured media/image/tool-result media payloads where replacing the payload corrupts the model turn;
* keep the existing compact active stub behavior for generic `raw_payload` / oversized opaque text payloads;
* keep the new preflight no-op regression for media payloads, and add/keep a regression proving oversized raw payloads still return an active stub and reconcile/idempotently replay against the protected stored row.

I am not approving the first-time contributor workflow run while the local full suite is failing.

stephenschoettler:
Pushed maintainer follow-up e25ad61 after merging current main into the branch.

This keeps the contributor's storage-only media fix, but restores the existing generic raw-payload contract: whole-message `raw_payload` externalization still returns the compact active stub, while structured media and tool-result media keep provider-usable originals in active replay.

Validation refreshed in the PR body. Key gates: targeted PR/review slice 491 passed, full suite 868 passed, low-fd full suite 868 passed, release stress `failure_count: 0` across 6 cases.

Requesting re-review once the latest workflow run is available.

Tosko4 (approved):
The follow-up fixes my previous blocker.

The PR now distinguishes the two active-context contracts correctly:

* provider-usable media/image/tool-result payloads stay intact in active replay, while storage remains protected;
* generic oversized `raw_payload` whole-message externalization still returns the compact active stub, so the host does not keep replaying huge opaque text.

I re-reviewed the updated diff on `e25ad61`. The change is still narrowly scoped to `engine.py` plus focused regressions, with no dependency, workflow, install-script, network, credential, or hidden-execution surface added.

Local validation:

```text
python -m pytest -q \
  tests/test_lcm_core.py::TestIngestExternalization::test_compress_returns_externalized_stub_for_oversized_active_tail \
  tests/test_ingest_protection.py::test_ingest_preserves_provider_active_context_while_protecting_storage \
  tests/test_ingest_protection.py::test_tool_result_ingest_preserves_active_context_while_protecting_storage \
  tests/test_ingest_protection.py::test_preflight_storage_protection_does_not_force_noop_compaction
4 passed

python -m pytest -q tests/test_ingest_protection.py tests/test_lcm_core.py tests/test_lcm_engine.py
686 passed, 1 warning

python -m pytest -q
868 passed, 1 warning

python scripts/lcm_stress_check.py --output /tmp/hermes-lcm-pr226-rereview-smoke --tier smoke --json
failure_count: 0

python -m compileall -q .
python -m py_compile scripts/import_lossless_claw.py scripts/lcm_stress_check.py
passed
```

I also manually checked the boundary behavior: media active replay retains the original data URI, SQLite rows do not contain raw media/raw sentinel payloads, raw oversized text returns the compact `raw_payload` active stub, and externalized refs remain recoverable.

GitHub checks are green on the current head. Approved.

stephenschoettler:

Issue #235, opened 2026-06-04
Slow startup: unconditional full FTS5 `integrity-check` on every launch scans a fragmented index (~81s on a 328MB DB)


ma-pony:
# Slow startup: unconditional full FTS5 `integrity-check` on every launch scans a fragmented index (~81s on a 328MB DB)

## Summary

On a long-lived `lcm.db` (328MB, ~54.6k messages), Hermes startup stalls for **~90s** inside LCM database initialization. The cost comes from the FTS5 **`'integrity-check'`** command, which runs **on every startup** with no caching and walks the *entire* FTS index. Because LCM's write pattern leaves the FTS5 index heavily **segment-fragmented**, that walk takes ~81s.

## Environment

- LCM DB: `~/.hermes/lcm.db`, **328MB**, **54,621 messages**
- Content is mostly tool output (terminal/read_file/web_fetch/MCP) + externalized payloads
- SQLite WAL mode (`PRAGMA journal_mode=WAL`, `busy_timeout=30000`)

## Evidence

Measured directly against the production DB:

| Observation | Result |
|---|---|
| FTS5 `integrity-check` on the live DB | **81.2 s** |
| `SELECT COUNT(*) FROM messages_fts` | **>60 s (timeout)** |
| Same `integrity-check` after the index was rebuilt | **2.3 s** |
| Disk reclaimed by the rebuild | only **~5MB** (328MB → 323MB) |

The startup log shows a ~90s black box between plugin registration and "LCM plugin loaded", matching the integrity-check cost.

Key signal: rebuilding the index reclaimed almost no disk (5MB) yet produced a **35x** speedup. So the cost is **not** disk fragmentation — it is **FTS5 inverted-index segment fragmentation**. Collapsing the many small segments into a compact layout is what makes the check fast again.

## Root Cause

Two compounding issues:

### 1. `integrity-check` runs on every startup, with no cache

`ensure_external_content_fts()` is called during DB init for each store and reaches `_fts_needs_rebuild()`:

```python
def _fts_needs_rebuild(conn, spec) -> bool:
    if _fts_needs_rebuild_structural(conn, spec):
        return True
    return check_external_content_fts_integrity(conn, spec)["status"] == "fail"
```

- `_fts_needs_rebuild_structural()` includes a content-vs-FTS **row-count comparison** (`COUNT(*)` over `*_fts`), which on a fragmented index is itself slow (>60s above).
[...]

PR #236, opened 2026-06-04
fix: throttle startup FTS integrity-check to cut launch time

ma-pony:
## Summary
- Throttle the startup FTS5 `integrity-check` (O(index size)) so it runs at most once per interval on the startup path, instead of on every startup of a structurally-sound index. New env `LCM_FTS_INTEGRITY_CHECK_INTERVAL_HOURS` (default `24`; `0` = every startup = previous behavior; negative = never). A per-table marker `fts_integrity_checked_at:<table>` is stored in the existing `metadata` table after a successful check or rebuild.
- The throttle is opt-in: `repair_external_content_fts()` / `_fts_needs_rebuild()` default to a forced deep check, so `/lcm doctor repair apply` (which calls `repair_external_content_fts()` directly) stays unthrottled; only `ensure_external_content_fts()` on the startup path passes `throttle=True`.
- Switch the structural content-vs-index row-count comparison from `COUNT(*) FROM <fts>` to `COUNT(*) FROM <fts>_docsize`.
- Reject non-finite `LCM_FTS_INTEGRITY_CHECK_INTERVAL_HOURS` values (`nan`/`inf`).

## Why
- On a long-lived DB (328MB, ~54.6k messages) the startup `integrity-check` measured ~81s and dominated launch time (issue #235). It guards against rare physical corruption; paying O(index) on every startup is the wrong frequency.
- `COUNT(*) FROM <fts>` on an external-content FTS5 table reads through to the content table, so it is both O(index size) **and** unable to reveal a lagging index (it always equals the content-table count). `<fts>_docsize` is a cheap ordinary-table count reflecting the true indexed-document count — faster, and it fixes a desync-detection blind spot.
- Explicit repair must stay unthrottled: same-row-count FTS drift passes structural checks but fails the integrity-check, so `/lcm doctor repair apply` needs the forced deep check to fix it.
- `nan`/`inf` would otherwise parse as a valid interval and suppress startup checks indefinitely once a marker exists.
- Safety nets are preserved: triggers maintain FTS consistency on every write, FTS query failures fall back to LIKE, structural checks run every startup, `/lcm doctor` offers explicit deep repair, and operators can set `0` to restore the prior every-startup behavior. A freshly rebuilt index is treated as verified (marker recorded).

## Validation
- [x] Full suite against the Hermes runtime: `pytest -q` -> `892 passed, 1 skipped, 2 failed`. The 2 failures (`test_path_containment`, `test_path_security`) are pre-existing macOS-only issues (`/tmp` -> `/private/tmp` symlink); they reproduce on clean `main` and exercise `externalize.py` path logic, not this change (diff is `db_bootstrap.py` + its tests + one README line).
- [x] Focused: `pytest tests/test_lcm_core.py tests/test_db_bootstrap_fts.py -q` -> `210 passed`
- [x] `python -m compileall -q .` -> OK

## Notes
- New `tests/test_db_bootstrap_fts.py` covers: existing index without marker runs the check and records a marker; fresh marker skips; expired marker reruns; `interval=0` checks every startup; negative interval never checks; structural breakage rebuilds despite a fresh marker; docsize-based desync detection; explicit repair fixes same-row-count corruption despite a fresh marker; `throttle=True` still skips; and `nan`/`inf` fall back to the default.
- Scope is intentionally limited to startup verification cost. Periodic FTS5 `'optimize'` (to prevent the segment fragmentation that makes integrity-check and queries degrade over time) is a separate concern and is not included here.

Refs #235

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,7 @@ environment variables:
 | `LCM_SUMMARY_TIMEOUT_MS` | `60000` | Timeout for one summarization call |
 | `LCM_EXPANSION_TIMEOUT_MS` | `120000` | Timeout for one `lcm_expand_query` synthesis call |
 | `LCM_DATABASE_PATH` | auto | SQLite database path, profile-scoped by default |
+| `LCM_FTS_INTEGRITY_CHECK_INTERVAL_HOURS` | `24` | Minimum hours between startup FTS5 deep integrity-checks (O(index size)). `0` checks every startup (previous behavior); a negative value never checks on startup. Structural checks always run regardless. |
 | `LCM_ENABLE_SLASH_COMMAND` | `false` | Enable the optional `/lcm` operator command surface |
 | `LCM_DOCTOR_CLEAN_APPLY_ENABLED` | `false` | Permit destructive `/lcm doctor clean apply` in trusted operator contexts |
 
--- a/db_bootstrap.py
+++ b/db_bootstrap.py
@@ -7,6 +7,7 @@ same schema-version marker, PRAGMA settings, and FTS repair behavior.
 from __future__ import annotations
 
 import logging
+import math
 import os
 import re
 import sqlite3
@@ -301,9 +302,14 @@ def _integrity_check_interval_hours() -> float:
     if raw is None:
         return DEFAULT_INTEGRITY_CHECK_INTERVAL_HOURS
     try:
-        return float(raw)
+        value = float(raw)
     except (TypeError, ValueError):
         return DEFAULT_INTEGRITY_CHECK_INTERVAL_HOURS
+    if not math.isfinite(value):
+        # nan/inf would suppress startup checks indefinitely once a marker
+        # exists; treat non-finite values as invalid.
+        return DEFAULT_INTEGRITY_CHECK_INTERVAL_HOURS
+    return value
 
 
 def _integrity_marker_key(spec: ExternalContentFtsSpec) -> str:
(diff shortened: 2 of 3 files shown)


stephenschoettler (changes requested):
Thanks, this targets the right startup cost, and the `_docsize` row-count switch makes sense. I need one repair-path fix before merge.

`repair_external_content_fts()` is now used by both startup and `/lcm doctor repair apply`, but the new marker gate lives inside `_fts_needs_rebuild()`. With a fresh marker, explicit repair skips the deep integrity check even when the index has same-row-count FTS drift/corruption that structural checks cannot see.

I reproduced this locally with a structurally valid external-content index, then a same-row-count stale update:

```text
structural after stale update False
integrity after stale update {'status': 'fail', 'detail': 'fts5: checksum mismatch for table "messages_fts"'}
repair result fresh marker {'rebuilt': False, 'degraded': False, 'triggers_recreated': False}
integrity after repair attempt {'status': 'fail', 'detail': 'fts5: checksum mismatch for table "messages_fts"'}
repair result forced interval 0 {'rebuilt': True, 'degraded': False, 'triggers_recreated': False}
integrity after forced {'status': 'pass', 'detail': 'ok'}
```

Please keep the throttle on the startup path only, or add an explicit unthrottled/force mode for doctor repair apply, with a regression for same-count stale-token corruption.

Non-blocking while touching this: `nan`/`inf` currently parse as floats for `LCM_FTS_INTEGRITY_CHECK_INTERVAL_HOURS`; with an existing marker they can suppress startup checks indefinitely. I would treat non-finite values as invalid and fall back to the default.

Local validation on this head: `pytest -q` passed, 892 tests.

ma-pony:
Both points addressed in d27e380.

**1. Throttle leaking into the shared repair path.** The gate lived in `_fts_needs_rebuild()`, which `/lcm doctor repair apply` reaches via `repair_external_content_fts()` (command.py:720-721). Made the throttle opt-in: `repair_external_content_fts()` and `_fts_needs_rebuild()` now default to `throttle=False` (forced deep check); only the startup path (`ensure_external_content_fts`) passes `throttle=True`. Doctor repair calls `repair_external_content_fts()` directly, so it is now unthrottled with **no change to command.py**.

I reproduced your same-row-count case locally before fixing (structural pass, integrity `fail` -> checksum mismatch) and added it as a regression: explicit repair rebuilds and restores integrity despite a fresh marker, while `throttle=True` still skips the deep check.

**2. Non-finite interval.** `nan`/`inf` are now rejected via `math.isfinite()` and fall back to the 24h default, with a test covering `nan`/`inf`/`-inf`/`Infinity`.

Focused validation: `pytest tests/test_lcm_core.py tests/test_db_bootstrap_fts.py -q` -> `210 passed`.

stephenschoettler (approved):
Looks good now. The startup path keeps the throttle, explicit repair remains unthrottled, and the same-row-count stale-token regression covers the blocker I reproduced. Non-finite interval parsing is handled too.

Validated locally: `pytest tests/test_db_bootstrap_fts.py tests/test_lcm_core.py -q` passed (210 tests), `pytest -q` passed (895 tests), `compileall` and `git diff --check` clean. CI is green after approving the fork run.

ma-pony:
Validation update: I now have the Hermes runtime locally and ran the full suite against it (Hermes venv, Python 3.11).

`pytest -q` -> **892 passed, 1 skipped, 2 failed**.

The 2 failures are `test_path_containment::test_path_containment_within_allowed_base` and `test_path_security::test_configured_externalization_path_inside_allowed_base_accepted`. Both are pre-existing and macOS-only: they assert a resolved path starts with the test tmp dir, but macOS symlinks `/tmp` -> `/private/tmp`. They reproduce identically on clean `main`, and they exercise `externalize.py` path logic — not this change (diff is `db_bootstrap.py` + its tests + one README line). They should pass on Linux CI, matching the 892 you saw.

PR #242, opened 2026-06-09
fix: prefer bound session on sibling chain when host has zero DAG

emiltsoi:
## Summary

Adds a sibling-chain fallback in `_continue_compression_boundary()`: when the stale host `old_session_id` has zero DAG nodes AND the bound session belongs to a different `conversation_id` (lifecycle fork) but shares the same parent, prefer the bound session as the carry-over source instead of setting `source_session_id=""` and orphaning the chain.

## Why

The #224 fix (`3d9c876`) handles stale host sessions by checking `_has_summary_nodes(old_session_id)`. When a session has zero DAG nodes — which happens naturally when the gateway tracks a very old session (8+ days stale) — the fix falls through to `source_session_id = old_session_id` directly. The bound session that actually crossed the compression boundary is never considered.

Additionally, when `_gateway_session_key` is `None` at init, two sessions on the same channel can receive different `conversation_id`s — creating a lifecycle fork. The `bound_conversation_matches` check is `False` even though both sessions share the same parent and the bound session has valid DAG. Without the fallback, the bound session's entire DAG chain is permanently orphaned — invisible to the bridge, unrecoverable.

The fix is narrow: one new condition in the existing `else` path, gated on three guards (shared parent, zero-DAG host, bound has DAG). Falls through to existing behavior if any guard fails.

## Change

28-line addition to `engine.py:_continue_compression_boundary()` — one new condition block.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -1899,8 +1899,8 @@ class LCMEngine(ContextEngine):
                         source_state = None
 
         conversation_id = (
-            kwargs.get("conversation_id")
-            or (source_state.conversation_id if source_state else None)
+            (source_state.conversation_id if source_state else None)
+            or kwargs.get("conversation_id")
             or self._conversation_id
             or source_session_id
             or old_session_id
--- a/engine.py
+++ b/engine.py
@@ -1865,8 +1865,38 @@ class LCMEngine(ContextEngine):
                         old_session_id,
                     )
                 else:
-                    source_session_id = ""
-                    source_state = None
+                    # Fallback: sibling chain with zero-DAG parent.
+                    # When stale old_session_id has no DAG nodes AND the
+                    # bound session belongs to a different conversation_id
+                    # but shares the same last_finalized_session_id
+                    # (parent) — prefer the bound session despite the
+                    # conversation_id mismatch. This handles the lifecycle
+                    # fork case where two sessions on the same channel
+                    # received different conversation_ids.
+                    bound_shares_parent_with_host = bool(
+                        bound_state
+                        and bound_state.last_finalized_session_id == old_session_id
+                    )
+                    host_has_no_dag = not bool(
+                        self._dag.get_session_nodes(old_session_id)
+                    )
+                    if (
+                        bound_shares_parent_with_host
+                        and host_has_no_dag
+                        and (bound_is_active_source or bound_is_finalized_source)
+                        and bound_has_summary_nodes
+                    ):
+                        source_session_id = previous_session_id
+                        source_state = bound_state
+                        logger.warning(
+                            "LCM compression boundary using bound session %s on sibling chain as carry-over source; host old_session_id=%s has zero DAG, parent=%s matches",
+                            previous_session_id,
+                            old_session_id,
+                            bound_state.last_finalized_session_id,
+                        )
+                    else:
+                        source_session_id = ""
+                        source_state = None
 
         conversation_id = (
             kwargs.get("conversation_id")
--- a/tests/test_lcm_engine.py
+++ b/tests/test_lcm_engine.py
@@ -7644,6 +7644,408 @@ class TestSessionRollover:
         assert conversation_b is not None
         assert conversation_b.current_session_id == "new-hermes-session"
 
+    # ── Sibling-chain fallback tests (PR #242, zero-DAG host) ──
+
+    def test_compression_boundary_sibling_chain_zero_dag_host_positive(
+        self, engine,
+    ):
+        """Active bound sibling with zero-DAG host — fallback activates."""
+        engine.on_session_start(
+            "lcm-source",
+            platform="telegram",
+            context_length=200000,
+            conversation_id="conversation-a",
+        )
+        source_store_id = engine._store.append(
+            "lcm-source",
+            {"role": "user", "content": "sibling chain context must move"},
+            token_estimate=17,
+            source="telegram",
+        )
+        stale_host_store_id = engine._store.append(
+            "old-hermes-session",
+            {"role": "user", "content": "stale host context must stay"},
+            token_estimate=11,
+            source="telegram",
+        )
+        source_node_id = engine._dag.add_node(SummaryNode(
+            session_id="lcm-source",
+            depth=0,
+            summary="sibling chain summary",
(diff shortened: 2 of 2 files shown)


stephenschoettler:

Please add:
- Why

emiltsoi:
## Test validation

Test suite requires `hermes-agent` core on PYTHONPATH:

```bash
PYTHONPATH=/path/to/hermes-agent pip install -r test-requirements.txt
PYTHONPATH=/path/to/hermes-agent pytest tests/test_lcm_engine.py -q
```

Results:
- **main (clean)**: 401 passed, 73 warnings
- **fix/sibling-chain-zero-dag-parent**: **401 passed, 73 warnings**
- **Zero regressions** — change adds a conditional fallback, never alters existing code paths

The sibling-chain condition is narrow: only activates when:
1. `host_source_session_id` is empty (existing path falls to `else`)
2. `bound_conversation_matches` is False (existing path hits the `else` inside the `else` — where `source_session_id=""` was unconditional)
3. Additional guards: shared parent, host has zero DAG, bound has DAG

If any of those three guards fail, behavior is identical to before (`source_session_id=""`).

emiltsoi:
Updated — added `## Summary`, `## Why`, and `## Change` sections to match the repo template. Let me know if anything else is needed before review.

stephenschoettler (changes requested):
Thanks. I’m going to hold this for the lifecycle edge cases.

The new fallback bypasses the normal `bound_conversation_matches` guard, but the later lifecycle binding still chooses `kwargs.get("conversation_id")` before `source_state.conversation_id`. That means an explicit mismatching `conversation_id` can move messages/DAG nodes from the bound source conversation while finalizing/binding under a different conversation. In my repro, conversation A kept `current_session_id="bound-session"` even after its rows/nodes had been reassigned into the new session under conversation B.

Please either reject this sibling-chain fallback when a requested `conversation_id` is present and differs from `bound_state.conversation_id`, or force the carry-over lifecycle finalization/bind to use the selected source conversation.

Also, this needs tests. At minimum:
- positive regression for the zero-DAG stale host + same-parent bound sibling case
- explicit `conversation_id` mismatch negative case
- bound source has no DAG
- parent mismatch
- host `old_session_id` has DAG and remains host-authoritative
- finalized sibling case, if that path is intended

The branch is also behind main, so please rebase after the fix and rerun CI.

emiltsoi:
Fixed — the `conversation_id` binding now respects the sibling-chain source selection.

**Change:** One-line priority swap in `_continue_compression_boundary()` L1901-1902:
- `source_state.conversation_id` is now position 1 (was position 2)
- `kwargs.get("conversation_id")` is now position 2 (was position 1)

**Why this works:** When `source.conversation_id == kwargs["conversation_id"]` (normal path), the OR chain produces the same value — zero behavior change. The only delta is the sibling-chain fork case where `conversation_id` differs, which is the bug being fixed.

**Test suite:** 401 passed, 73 warnings — zero regressions against main.

**Blast radius analysis by case:**
- Normal (no fork): source == kwargs → same value → no change
- Sibling-chain fork (the bug): source wins (correct) instead of kwargs (wrong)
- source_state is None: kwargs wins → no change
- source_state has None conversation_id: kwargs wins → no change
- kwargs has no conversation_id: source wins (was pos 2, now pos 1) → no change

**Next:** Writing the 7 test cases you requested (positive regression, conversation_id mismatch negative, no-DAG guard, parent mismatch, host-authoritative, finalized sibling, plus the specific regression case). Will push those separately. The branch is now rebased on main, CI should pick it up.

emiltsoi:
All review feedback addressed.

**Fix:** One-line priority swap — `source_state.conversation_id` now position 1 (was 2), `kwargs.get("conversation_id")` now position 2 (was 1).

**Test cases (7):**
- ✅ Positive: zero-DAG stale host + same-parent active bound sibling (the bug case)
- ✅ Negative: bound source has no DAG → fallback deactivated
- ✅ Negative: parent mismatch → fallback deactivated
- ✅ Negative: host has DAG → stays host-authoritative
- ✅ Positive: active sibling with different conversation_id → fallback activates
- ✅ Edge: source_state=None → kwargs fallback
- ✅ Regression: conversation_id from bound session, not kwargs (the specific bug)

**Full test suite:** 408 passed, 73 warnings — zero regressions (baseline 401 + 7 new).

Branch is rebased on main, ready for re-review.

stephenschoettler:
Update: your latest push with the test cases landed while I was writing the previous comment.

I re-checked the current head locally. The code blocker looks cleared and the new regression coverage is in place.

Validation on `dba4fecf9f3b6383d6fdd0e8ed8631d346f4b9bd`:
- `pytest -q tests/test_lcm_engine.py -k 'sibling_chain or compression_boundary'` -> 29 passed
- `pytest -q tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py` -> 624 passed
- same validation after a local merge with current `main` -> 29 passed / 628 passed
- `git diff --check origin/main...HEAD` -> passed

Remaining process item: GitHub still marks the branch behind current `main`, and the fork CI run is still `action_required` with no checks. Please rebase onto current `main`; after that I’ll approve the fresh fork CI run and re-check.

emiltsoi:
Rebased onto current `main` — clean.

- `pytest -q tests/test_lcm_engine.py -k "sibling_chain"` → 7 passed
- `pytest -q tests/test_lcm_engine.py` → 408 passed, 73 warnings

Ready for CI approval and re-review.

stephenschoettler (approved):
Looks good now. The lifecycle binding uses the selected source conversation, the zero-DAG sibling fallback is guarded, and the requested regression coverage is in place.

Validated locally on `ef25563a1eabdc273944dd1946508d6f9f58781a`:
- `pytest -q tests/test_lcm_engine.py -k 'sibling_chain or compression_boundary'` -> 29 passed
- `pytest -q tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py` -> 628 passed
- `git diff --check origin/main...HEAD` -> passed

Fork CI is green across 3.11, 3.12, 3.13, 3.14, and workflow-lint.

PR #244, opened 2026-06-09
Fix conversation-scoped runtime state for shared LCM engines

### LINKED ISSUE #243: Shared LCM engine can write compression output to the wrong conversation
## Summary

Hermes gateway can run multiple Discord/forum thread sessions concurrently in one Python process while `hermes-lcm` is exposed as a single shared context-engine instance. Runtime session state on that shared `LCMEngine` can be overwritten by another conversation before `compress()` persists raw messages or summary DAG nodes.

This can cause one conversation's raw messages, summaries, and lifecycle frontier to be written under another conversation/session.

## Root cause

`LCMEngine` keeps mutable runtime fields process-wide, including:

- `_session_id`
- `_conversation_id`
- `_ingest_cursor`
- `_last_compacted_store_id`
- `_foreground_session_id`
- compression counters/status fields

The host currently calls compression as:

```python
agent.context_compressor.compress(messages, current_tokens=..., focus_topic=...)
```

so `compress()` relies on whatever `_session_id` / `_conversation_id` is currently bound on the shared engine. In a gateway process with cached agents and concurrent executor turns, a different thread/session can rebind those fields before another conversation calls `compress()`.

## Minimal reproduction shape

1. Shared engine starts session A / conversation A.
2. Shared engine starts session B / conversation B, overwriting process-wide binding.
3. Cached agent for A calls `compress(A_messages)` without a fresh `on_session_start(A)`.
4. A raw rows and summary nodes are persisted under B's session/conversation.

## Expected behavior

Compression should bind runtime cursors/counters/session identity to the active logical conversation before writing raw rows, summary nodes, or lifecycle frontier state.

## Fix approach

A PR is coming that scopes mutable runtime state by logical conversation and resolves the active scope from explicit kwargs or Hermes gateway's task-local `HERMES_SESSION_KEY` context value, preserving backward compatibility with the current host call signature.

tulysa:
## Summary

Fixes #243.

This PR prevents a shared `LCMEngine` instance from writing compression output to the wrong session/conversation when Hermes gateway serves multiple logical conversations concurrently in one process.

## Why

Hermes gateway can reuse one process-wide context-engine instance across multiple logical conversations. The current host integration calls:

```python
agent.context_compressor.compress(messages, current_tokens=..., focus_topic=...)
```

without passing the current session/conversation into `compress()`. That means `compress()` relies on mutable engine fields such as `_session_id`, `_conversation_id`, `_ingest_cursor`, and `_last_compacted_store_id`.

In concurrent gateway traffic, another conversation can call `on_session_start()` and overwrite those fields before a cached agent calls `compress()`. When that happens, the wrong session/conversation can receive:

- raw message rows,
- summary DAG nodes,
- lifecycle frontier updates.

This is particularly risky because the bad write happens before compression-boundary recovery has a chance to correct session drift.

## What changed

- Add per-conversation runtime state snapshots for session/cursor/compression/lifecycle bookkeeping fields.
- Bind the shared engine to the active conversation for `compress()` using:
  1. explicit `compress(..., session_id=..., conversation_id=...)` kwargs,
  2. Hermes gateway task-local `HERMES_SESSION_KEY` via dynamic import of `gateway.session_context`,
  3. known session-to-conversation mapping,
  4. current bound conversation as fallback.
- Keep `compress(messages, current_tokens=..., focus_topic=...)` backward compatible.
- Keep `on_session_start()` behavior compatible with existing callers: after session start the engine remains bound to that session.
- Refresh scoped runtime state after `update_from_response()`.
- Add regression coverage for cached-agent/shared-engine contamination and compression-boundary continuation.

## Validation

Passed:

```bash
python3 -m py_compile engine.py tests/test_context_scoped_runtime.py
python3 -m pytest tests/test_context_scoped_runtime.py tests/test_lcm_engine.py tests/test_lcm_command.py tests/test_lcm_core.py -q
# 650 passed, 1 warning
```

Full suite note:

```bash
python3 -m pytest -q
# 909 passed, 2 failed, 1 warning
```

The two failures also reproduce when run alone and appear orthogonal to this PR:

- `tests/test_ingest_protection.py::test_import_lossless_claw_externalizes_legacy_data_uri_content`
- `tests/test_ingest_protection.py::test_import_lossless_claw_respects_externalization_path_env`

They involve lossless-claw importer externalized-payload path/content expectations, not conversation-scoped runtime state.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -386,6 +386,14 @@ class LCMEngine(ContextEngine):
         self._auxiliary_session_ids: set[str] = set()
         self._auxiliary_lineage_session_ids: set[str] = set()
         self._auxiliary_session_lock = threading.RLock()
+        # Hermes gateways can run many conversations concurrently in one
+        # process while the plugin manager exposes one shared context-engine
+        # instance. Keep mutable compression/session cursors scoped by logical
+        # conversation so one thread cannot steal another thread's _session_id
+        # before compress() persists raw rows or DAG nodes.
+        self._runtime_state_lock = threading.RLock()
+        self._runtime_state_by_conversation: dict[str, dict[str, Any]] = {}
+        self._runtime_conversation_by_session: dict[str, str] = {}
 
     def _resolve_db_path(self, hermes_home: str = "") -> Path:
         """Resolve the SQLite path for the active Hermes profile/home."""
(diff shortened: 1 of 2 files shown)


stephenschoettler:

Please add:
- Why
- Validation

tulysa:
Updated the PR description to include the requested `Why` and `Validation` sections. Code is unchanged.

stephenschoettler (changes requested):
Thanks for the update. I’m going to hold this one because the fix still leaves the shared engine corruptible outside `compress()`.

I reproduced this sequence on the PR branch:

1. `on_session_start("A-session", conversation_id="conv-A", platform="alpha")`
2. `on_session_start("B-session", conversation_id="conv-B", platform="beta")`
3. `on_session_end("A-session", A messages)`

Actual result: A’s rows are written as `B-session` / `beta`, and the B lifecycle is finalized with `A-session` while A remains active.

So the PR scopes `on_session_start()` and `compress()`, but not the other mutable entry points. Please add a regression for the exact `A start -> B start -> A on_session_end` case, then either scope every mutable ContextEngine entry point (`on_session_end`, preflight/should_compress, `update_from_response`, tool/status/reset/rollover paths, token/status accounting) or switch to the clone-per-agent approach so one conversation cannot inherit another conversation’s bound runtime state.

stephenschoettler:
Closing this as superseded by #247.

The bug report and implementation work here were useful, but I decided not to merge the shared-runtime-state approach because it still leaves too many mutable ContextEngine entry points needing scoped rebinding.

#247 landed the safer clone-per-agent direction instead. That keeps per-agent session/lifecycle/ingest state isolated while sharing the configured store.

Leaving #243 open because the full live fix still depends on the Hermes Agent host calling `clone_for_agent()` before per-agent session/model binding.

PR #247, opened 2026-06-09
fix: isolate LCM engine state per agent

the3asic:
## Summary

- Add `LCMEngine.clone_for_agent()` so Hermes can create one runtime LCM engine per `AIAgent`.
- The clone shares the configured SQLite database path/configuration but gets isolated session binding, lifecycle, and ingest cursor state.
- Add a regression proving two cloned engines can bind different conversations and finalizing one does not write/finalize the other.
- Fix the test package shim so monkeypatch paths like `hermes_lcm.db_bootstrap...` resolve through the parent package.

## Why

`LCMEngine` stores mutable runtime state on the engine object (`_session_id`, `_conversation_id`, ingest cursor/frontier, foreground markers). Hermes gateway can keep multiple cached agents alive at once. If all agents reuse the one plugin-registered engine object, one conversation can rebind another conversation's LCM ingest/finalization state.

This is the preferred LCM-side shape for #243 because it isolates the full runtime engine instead of trying to scope selected mutable fields on a shared object.

This pairs with NousResearch/hermes-agent#42683, which adds the host-side optional clone hook. Without that host change, this method is harmless but not called by Hermes core.

## Validation

Current PR head: `6c00c13f382147f6b42e1e23ebfbb5347ecb6fe0`
Current base used for refresh: `66cb6c78fe82b0f18f0d94734acd6d307b846276`

- [x] Focused validation: `python -m pytest -q tests/test_lcm_core.py::TestLCMEngineCloning::test_clone_for_agent_isolates_session_binding_while_sharing_store` -> `1 passed, 21 warnings`
- [x] Focused/default slice: `python -m pytest -q tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py tests/test_host_capability.py` -> `639 passed, 46 warnings`
- [x] Default validation: `python -m pytest -q` -> `921 passed, 60 warnings`
- [x] Low-fd default validation: `prlimit --nofile=1024:1024 -- python -m pytest -q` -> `921 passed, 60 warnings`
- [x] `git diff --check origin/main...HEAD` -> passed

## Notes

- Original implementation and regression are by @the3asic in this PR.
- Maintainer refresh merged current `main` into the contributor branch so authorship/history is preserved.
- This should use `Refs #243`, not a closing keyword, until the host-side clone hook in NousResearch/hermes-agent#42683 lands and is verified.
- Follow-up review should verify Hermes Agent calls `clone_for_agent()` before per-agent `update_model()` / `on_session_start()`, and that slash/status command routing does not keep pointing at the unused prototype engine.

Refs #243

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -4,6 +4,7 @@ Implements the ContextEngine ABC. Replaces the built-in ContextCompressor
 with a DAG-based summarization system that preserves every message.
 """
 
+import copy
 import inspect
 import json
 import logging
@@ -386,6 +387,25 @@ class LCMEngine(ContextEngine):
         self._auxiliary_session_ids: set[str] = set()
         self._auxiliary_lineage_session_ids: set[str] = set()
         self._auxiliary_session_lock = threading.RLock()
+
+    def clone_for_agent(self) -> "LCMEngine":
+        """Return a fresh runtime engine for one AIAgent instance.
+
+        Hermes registers plugin context engines process-wide, while gateway
+        runtimes may keep multiple cached AIAgent instances alive at once
+        (different platforms, chats, cron jobs, etc.).  LCM stores mutable
+        session binding and ingest cursor state on the engine object itself, so
+        sharing one registered instance across agents can let one conversation
+        rebind another conversation's raw-message ingest and lifecycle state.
+
+        The clone shares the same durable SQLite database path/configuration,
+        but gets independent session/cursor/lifecycle runtime state. Hermes core
+        calls update_model() on the returned engine before on_session_start().
+        """
+        return type(self)(
+            config=copy.deepcopy(self._config),
+            hermes_home=self._hermes_home,
+        )
 
     def _resolve_db_path(self, hermes_home: str = "") -> Path:
         """Resolve the SQLite path for the active Hermes profile/home."""
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -43,6 +43,7 @@ if pkg_name not in sys.modules:
             sub_mod = importlib.util.module_from_spec(sub_spec)
             sub_mod.__package__ = pkg_name
             sys.modules[sub_name] = sub_mod
+            setattr(mod, py_file.stem, sub_mod)
             try:
                 sub_spec.loader.exec_module(sub_mod)
             except Exception:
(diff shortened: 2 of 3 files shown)


stephenschoettler:

Please add:
- Validation

the3asic:
Updated the PR description with a `Validation` section. No code changes.

stephenschoettler (commented):
This is the direction I prefer for the shared-engine bug: isolate the full engine object per agent instead of swapping selected runtime fields on one shared instance.

I’m leaving this as a comment for now because this LCM-side hook is only effective once the Hermes Agent companion change actually calls `clone_for_agent()` before per-agent `update_model()` / `on_session_start()`. Please keep the companion PR linked and include validation that the host is invoking the clone path.

One more thing to verify before this is treated as the complete fix: slash/status command routing should not keep pointing at the unused prototype engine once agents receive clones.

stephenschoettler:

stephenschoettler:
This still looks useful, but the branch is stale: it is behind main and has no visible checks on the latest head.

Safe next step: salvage the useful patch onto current main with attribution if replay and validation pass. I do not want to merge this stale branch directly.

stephenschoettler:
Refreshed this onto current `main` while preserving @the3asic's original commit history.

Current head: `6c00c13f382147f6b42e1e23ebfbb5347ecb6fe0`
Base: `66cb6c78fe82b0f18f0d94734acd6d307b846276`

Local validation passed on the refreshed head:
- focused clone regression: `1 passed`
- focused/default slice: `639 passed`
- full suite: `921 passed`
- low-fd full suite: `921 passed`
- compileall, import script py_compile, install/update bash syntax, and diff check passed

GitHub CI is green on the refreshed head. I updated the PR body to keep this as `Refs #243` because the LCM hook still depends on the host-side clone call in NousResearch/hermes-agent#42683 before it completes the live shared-engine fix.

I requested @Tosko4 for the core session-boundary review.

stephenschoettler:
Thanks. CI is green on `6c00c13`, and I've asked @Tosko4 for review. If @Tosko4 approves and checks stay green, I'll merge this.

Tosko4 (approved):
Reviewed the current head against latest main. The LCM-side shape looks right to me: cloned engines get isolated session, lifecycle, and ingest cursor state while still sharing the configured store, and the PR keeps the host dependency explicit instead of claiming #243 is fully fixed here.

I reran the focused clone regression, the core/engine/packaging/host slice, the full suite, compile checks, script syntax checks, and diff check locally. All passed.

Approving from my side.

Tosko4:
@stephenschoettler

PR #256, opened 2026-06-11
fix(lifecycle): gc empty session rows to prevent unbounded accumulation

flooryyyy:
## Summary

Adds `prune_empty_sessions()` to `LifecycleStateStore` to garbage-collect lifecycle rows for sessions that never ingested any messages or summary nodes. Exposes 3 new config knobs (`LCM_EMPTY_LIFECYCLE_GC_ENABLED`, `LCM_EMPTY_LIFECYCLE_GC_THRESHOLD`, `LCM_EMPTY_LIFECYCLE_GC_MAX_AGE_HOURS`) and integrates the cleanup into `_bind_lifecycle_state()` to run automatically when the table exceeds the threshold. Also adds `/lcm doctor clean lifecycle` (dry-run) and `/lcm doctor clean lifecycle apply` operator commands.

## Why

Every gateway restart, cron tick, or ephemeral TUI/cron session creates a lifecycle row via `bind_session()`. If the session never ingests a single message, the row persists forever. LCM had no cleanup mechanism for zero-data sessions. On a 30-day-old live instance, 573 of 1393 lifecycle rows (41%) had zero stored data — pure orphans living in `lcm_lifecycle_state` with no corresponding messages or `summary_nodes`.

## Validation

- [x] Focused validation: `pytest tests/test_lcm_core.py::TestLifecycleStateStore::test_prune_empty_sessions_deletes_row_with_zero_data -v` -> PASSED
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q`
  - [x] `pytest -q`
  - [x] `bash -lc 'ulimit -n 1024 && pytest -q'`
  - [x] `git diff --check`

## Notes

- Protected sessions (the actively-bound `_session_id`) are never pruned.
- GC is wrapped in `try/except` so a failure never breaks session binding.
- Performance: ~5ms on a 1400-row table.


stephenschoettler:

Please add:
- Summary
- Why
- Validation

flooryyyy:
my agent randomly brought this up and got told to find a high-quality solution and submit a caveman-pr, aimed to be minimalistic. i was unaware of the contribution guidelines at the time of opening the PR. i've updated it now, let me know if it is okay :)

stephenschoettler (changes requested):
Thanks for updating the PR body. The feature direction is good and CI is green, but I need two data-safety issues fixed before this merges.

1. `/lcm doctor clean lifecycle apply` needs to be backup-first.

This is a destructive cleanup path, but it currently calls `engine._lifecycle.prune_empty_sessions(...)` directly and returns success without creating or reporting a backup. Existing `/lcm doctor clean apply` is backup-first, and this should follow the same operator-safety contract: create the backup, abort if backup creation fails, and include the backup path/size in the output.

2. `prune_empty_sessions()` needs to be atomic/race-safe.

The lifecycle connection runs in autocommit mode. The new method snapshots `sessions_with_data`, then loops and deletes lifecycle rows one by one. A failure mid-loop can partially delete rows, and a concurrent writer can add message/node data after the snapshot but before deletion. Please wrap candidate selection and deletion in an explicit transaction, preferably `BEGIN IMMEDIATE`, and recheck emptiness in the delete path so a concurrent message/node write cannot land after the snapshot and then lose its lifecycle row.

Also please handle the small surface cleanup while you are there:

- add command-surface tests for the new lifecycle clean scan/apply output, including denied apply and backup behavior
- document `/lcm doctor clean lifecycle`, `/lcm doctor clean lifecycle apply`, and the new `LCM_EMPTY_LIFECYCLE_GC_*` env vars in README/help
- either remove the plugin version bump from this feature PR or update all user-facing version references consistently
- fix or remove the trailing empty `Refs #` in the PR body

Local validation on the current head passed:

- `python -m pytest tests/test_lcm_core.py::TestLifecycleStateStore -k prune_empty -q`
- `python -m pytest tests/test_lcm_engine.py -k test_bind_lifecycle_gc -q`
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q`
- `python -m compileall -q .`
- `python -m py_compile scripts/import_lossless_claw.py && bash -n scripts/install.sh scripts/update.sh`

Once the backup/atomicity fixes are in and CI is green again, this should be in good shape for technical review.

stephenschoettler (changes requested):
The data-safety blockers from my last review are fixed on this head. I verified the lifecycle apply path now creates a backup before deletion, and a forced mid-delete failure rolls back instead of partially pruning rows. CI is also green.

One cleanup pass remains before I can clear the changes-requested review:

1. Please document the new slash commands in the operator command list and `/lcm help` output:
   - `/lcm doctor clean lifecycle`
   - `/lcm doctor clean lifecycle apply`

   The env vars are documented now, but the command list still jumps from `doctor clean apply` to `doctor repair`, and `_help_text()` does not mention the lifecycle variants.

2. The version bump is still present. `plugin.yaml` and runtime identity tests now say `0.16.3`, but the README still shows `hermes-lcm v0.16.2`. Either revert the feature PR version bump and leave it for the release PR, or update every user-facing version reference consistently.

Local validation on current head `3739665`:

- `python -m pytest tests/test_lcm_command.py::test_lcm_doctor_clean_lifecycle_reports_empty_candidates tests/test_lcm_command.py::test_lcm_doctor_clean_lifecycle_apply_is_backup_first_and_deletes_safe_candidates tests/test_lcm_command.py::test_lcm_doctor_clean_lifecycle_apply_denied_by_default tests/test_lcm_command.py::test_lcm_doctor_clean_rejects_unknown_extra_args -q` -> 4 passed
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> 641 passed
- `python -m pytest -q` -> 936 passed
- `prlimit --nofile=1024:1024 -- python -m pytest -q` -> 936 passed
- `python -m compileall -q . && python -m py_compile scripts/import_lossless_claw.py && bash -n scripts/install.sh scripts/update.sh && git diff --check origin/main...HEAD` -> pass

After those two small surface fixes, this should be ready for Tosko4 review because it changes lifecycle/core cleanup behavior.

stephenschoettler (approved):
Final cleanup is now on the PR branch.

Verified:
- command help and README now document `/lcm doctor clean lifecycle` and `/lcm doctor clean lifecycle apply`
- feature PR version bump is reverted to v0.16.2
- local targeted tests: 11 passed
- local full suite: `python -m pytest -q` -> 937 passed
- local low-fd full suite: `prlimit --nofile=1024:1024 -- python -m pytest -q` -> 937 passed
- static validation: compileall, py_compile, bash -n, and diff checks passed
- GitHub CI is green on `18a1a28`

Approving from maintainer side. I requested Tosko4 review before merge because this touches lifecycle cleanup and destructive operator behavior.

Tosko4 (changes requested):
Found one blocker.

`engine.py:1183` enables automatic lifecycle GC by default once the table exceeds the threshold, and `empty_lifecycle_gc_max_age_hours` defaults to `None`. At that point `prune_empty_sessions()` deletes any lifecycle row whose referenced sessions have no messages or summary nodes, except for this engine's single `_session_id`.

That can delete another live session from the same shared LCM database. A session that has called `on_session_start()` but has not ingested its first message yet has exactly the empty shape this GC targets. If a second engine/session starts while the table is over the threshold, the second engine can prune the first engine's active lifecycle row. The later message ingest still writes raw messages, but the durable lifecycle/conversation row is gone, so frontier persistence and lifecycle recovery for that conversation are lost.

I reproduced it locally with two `LCMEngine` instances sharing one DB and `empty_lifecycle_gc_threshold=1`:

```text
after e1 start row_count= 1 active-other_exists= True
after e2 start row_count_e1= 1
active-other_exists_e1= False
active-other_exists_e2= False
messages_active_other= 1
lifecycle_after_ingest_exists= False
```

The focused tests and full suite pass (`937 passed, 1 warning`), but they do not cover this cross-engine active-empty-session case.

I think the automatic path needs a safer staleness guard before this is mergeable. Good options would be to default a nonzero age guard, always preserve recent `current_bound_at` rows, or limit automatic GC to rows that are provably stale rather than merely data-empty. The backup-first manual doctor cleanup can stay more explicit, but the enabled-by-default startup GC should not be able to delete another currently active session.

stephenschoettler:
Tosko4 is right. I reproduced the cross-engine active-empty-session failure locally on this head (`18a1a28`).

```text
after e1 start row_count= 1 active-other_exists= True
after e2 start row_count_e1= 1
active-other_exists_e1= False
active-other_exists_e2= False
messages_active_other= 1
lifecycle_after_ingest_exists= False
```

So my maintainer approval should not be treated as merge-ready until this is fixed. The enabled-by-default startup GC must not be able to delete another live session that has bound lifecycle state but has not ingested its first message yet.

Please patch the automatic GC path with a real staleness/liveness guard and add a regression test with two `LCMEngine` instances sharing one DB:

1. engine A starts a session and creates a lifecycle row with no messages yet
2. engine B starts and triggers empty lifecycle GC
3. engine A's lifecycle row still exists
4. engine A ingests its first message
5. engine A's lifecycle row still exists

I also noticed the current "recheck emptiness right before deletion" comment is misleading: it checks the same `sessions_with_data` snapshot, not a fresh per-candidate query. Either make that recheck real inside the transaction or remove the claim.

stephenschoettler:
@Tosko4 I pushed a maintainer fix at `f29ec0e` for the blocker you found.

Changes:
- automatic GC now defaults to a 24h age guard, so startup GC preserves recently bound empty sessions
- stale empty rows are still pruned once old enough
- pre-delete emptiness recheck now performs fresh table queries inside `BEGIN IMMEDIATE`
- added a two-engine regression covering the active-empty session case

Validation:
- focused lifecycle/doctor slice: 14 passed
- `python -m pytest -q`: 938 passed
- low-fd full suite: 938 passed
- compileall, py_compile, bash -n, and git diff --check: pass
- GitHub CI green on `f29ec0e`

Could you re-review when you have a chance?

Tosko4 (approved):
The new head fixes my blocker.

I re-ran the cross-engine repro against `f29ec0e`: with the new default 24h age guard, a second engine starting in the same shared DB no longer prunes another recently-bound empty active session, and the first session can ingest its first message while keeping its lifecycle row.

Verified locally:

```text
after e1 start row_count= 1 active-other_exists= True
after e2 start row_count_e1= 2
active-other_exists_e1= True
active-other_exists_e2= True
messages_active_other= 1
lifecycle_after_ingest_exists= True
```

Also ran:

```text
23 focused lifecycle/doctor tests passed
938 passed, 1 warning
python -m compileall -q .
```

Project shape looks good now: the automatic cleanup is bounded by a conservative staleness guard, while immediate deletion stays explicit/operator-driven through the doctor apply path.

stephenschoettler:

flooryyyy:
only just ended up seeing progress on this PR as i was busy. thank you for finishing the PR :)

PR #260, opened 2026-06-16
fix: per-turn ingest for WebUI sessions + batch timestamp dedup

kaishi00:
## Summary

Two fixes for LCM message persistence:

1. **Per-turn ingest (`__init__.py` + `engine.py`):** Register a `post_llm_call` plugin hook that calls `engine.ingest(conversation_history)` after every completed turn.  Messages land in the durable store regardless of whether compression triggers.  The existing `_ingest_messages` cursor prevents duplicates if `compress()` runs later the same turn.

2. **Batch timestamp dedup (`store.py`):** Move `time.time()` inside the for-loop in `append_batch()` so each message gets its own unique timestamp.

### 1. WebUI messages never reach LCM

LCM only ingests messages when (a) compression triggers, (b) `on_session_end` fires, or (c) an LCM tool is called.  Telegram sessions expire → cleanup → `on_session_end()` → ingest.  But WebUI sessions stay alive forever in the browser, and short conversations never hit the compression threshold.  Result: **WebUI messages from short conversations are invisible to LCM forever** — journal crons, cross-session search, and memory providers all miss them.

### 2. Batch timestamp collision

`append_batch()` calculated `time.time()` **once** before the loop and reused it for ALL messages in the batch.  Result: 50-122+ messages sharing identical timestamps, breaking all date-based queries (journal entries pull wrong days, time-range filtering returns wrong results).

## Validation

- [x] Focused validation: `python3 -m compileall -q .` → clean
- [x] Focused validation: `git diff --check` → clean
- [x] Default validation: `pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q` → 630 passed, 1 warning (deprecation), 0 failed (1 pre-existing timeout unrelated to this change)
- [x] `python -m compileall -q .` → clean

## Regressions

Five deterministic regression tests added:

- `test_append_batch_timestamps_are_unique_per_row` — 50-message batch, all timestamps distinct and non-decreasing
- `test_below_threshold_turn_persists_without_compression` — 3-message conversation persisted, compression count stays 0
- `test_ingest_then_compress_no_duplicate_rows` — ingest then compress on a long conversation, row count unchanged
- `test_ignored_session_ingest_persists_nothing` — `cron:*` pattern, ingest is a no-op
- `test_stateless_session_ingest_persists_nothing` — `telegram:*` pattern, ingest is a no-op

## Architecture

The `post_llm_call` hook is registered from the LCM plugin's `register()` function via `get_plugin_manager()._hooks`.  It fires once per turn after the tool-calling loop completes and receives `conversation_history` including the assistant response.  This is a **self-contained LCM-local fix** — no companion hermes-agent PR is required.

`_ingest_messages()` is called from:
1. `compress()` — only when threshold exceeded
2. `on_session_end()` — only on session expiry
3. `handle_tool_call()` — only when LCM tools invoked
4. **`post_llm_call` hook** (new) — unconditionally every turn

The existing cursor-based dedup means if `compress()` runs later the same turn, already-ingested messages are skipped.

## Notes

- Zero extra LLM calls — `_ingest_messages()` is pure SQLite writes.  LLM calls only happen inside `compress()` for summarization.
- The companion hermes-agent PR (#47243) that added `ingest()` calls from `turn_context.py` is **no longer needed** — the hook approach is self-contained.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -137,4 +137,30 @@ def register(ctx):
     else:
         logger.info("LCM slash command registration unavailable on this Hermes host; continuing without /lcm")
 
+    # Register a post_llm_call hook so every completed turn is persisted to
+    # the durable store, regardless of whether compression triggers.  Without
+    # this, short WebUI conversations (which never expire and may never hit
+    # the compression threshold) are invisible to LCM forever.
+    #
+    # The hook fires once per turn after the tool-calling loop completes and
+    # receives conversation_history including the assistant response.  The
+    # existing _ingest_messages cursor prevents duplicates if compress() runs
+    # later the same turn.
+    try:
+        from hermes_cli.plugins import get_plugin_manager as _get_pm
+        _mgr = _get_pm()
+
+        def _on_post_llm_call(**kwargs):
+            history = kwargs.get("conversation_history")
+            if history:
+                try:
+                    engine.ingest(history)
+                except Exception as exc:
+                    logger.debug("LCM post_llm_call ingest error: %s", exc)
+
+        _mgr._hooks.setdefault("post_llm_call", []).append(_on_post_llm_call)
+        logger.debug("LCM registered post_llm_call hook for per-turn ingest")
+    except Exception as exc:
+        logger.debug("LCM could not register post_llm_call hook: %s", exc)
+
     logger.info("LCM plugin loaded — lossless context management active")
--- a/engine.py
+++ b/engine.py
@@ -651,7 +651,7 @@ class LCMEngine(ContextEngine):
     def ingest(self, messages: List[Dict[str, Any]]) -> None:
         """Persist messages to the durable store every turn.
 
-        Called by the host (turn_context) so messages land in LCM
+        Called by the post_llm_call plugin hook so messages land in LCM
         regardless of whether compression triggers — short WebUI
         conversations never hit the compression threshold and never
         expire like Telegram sessions do, so without this they'd never
--- a/engine.py
+++ b/engine.py
@@ -648,6 +648,31 @@ class LCMEngine(ContextEngine):
         self._last_boundary_skip_time = 0
         return False
 
+    def ingest(self, messages: List[Dict[str, Any]]) -> None:
+        """Persist messages to the durable store every turn.
+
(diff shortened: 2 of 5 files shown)


stephenschoettler:

Please add:
- Summary
- Why
- Validation

kaishi00:
Heads up — the `store.py` timestamp fix breaks one existing test:

`test_search_recency_same_timestamp_pool_is_limit_stable` in `tests/test_lcm_core.py`

The test was asserting the broken behavior — that all messages in a batch share an identical timestamp:

```python
timestamp = store.get(ids[0])["timestamp"]
assert [result["timestamp"] for result in short_results] == [timestamp] * len(short_results)
```

With the fix, each message gets its own `time.time()` call, so timestamps are no longer identical. I updated the test to check the real invariant (search result ordering is deterministic — top-N from a limited search matches first N of an unlimited search) without assuming identical timestamps.

The change is included in this PR. Let me know if you'd prefer the test fix as a separate commit or handled differently.

stephenschoettler (changes requested):
Thanks — both problems are real, but I can’t merge this as-is yet.

The timestamp fix is directionally good. Please remove the inline arrow comment and add/keep a deterministic regression that proves `append_batch()` timestamps are generated per row.

The per-turn ingest fix currently does not actually run on current Hermes Agent. This PR adds `LCMEngine.ingest()`, but Hermes Agent does not call `context_engine.ingest()` from `turn_context` or anywhere else, so below-threshold WebUI sessions still won’t be persisted.

Preferred LCM-local fix: register a `post_llm_call` hook from the plugin and call `engine.ingest(conversation_history)` there. That hook sees the completed turn, including the assistant response. The existing ingest cursor should prevent duplicates when preflight/compression already ingested part of the same message list.

Please add regressions for:
- below-threshold completed turn persists without compression
- preflight/compression plus `post_llm_call` ingest does not duplicate rows
- ignored/stateless sessions remain gated
- batch timestamps are generated per row

Happy to review again after that.

stephenschoettler (approved):
Reviewed current head. Requested changes are addressed and CI is green.

PR #280, opened 2026-06-25
fix: pin summary role to user after system anchor to prevent HTTP 400

fesalfayed:
## Summary

Fix an HTTP 400 from Anthropic after the second context compaction. `_assemble_context` assigned the DAG summary `role="assistant"`. After two compactions the fresh tail and earlier turns have been compacted away, so the summary becomes the first message reaching the provider with only the system prompt ahead of it. The Anthropic adapter extracts the system prompt into its separate `system` field, leaving an `assistant` turn as `messages[0]` — which Anthropic rejects with a misleading error:

```
messages.2: `tool_use` ids were found without `tool_result` blocks
immediately after: toolu_.... Each `tool_use` block must have a
corresponding `tool_result` block in the next message.
```

The error names `tool_use`, but the structural violation is the leading `assistant` turn; Anthropic requires `messages[0]` to be `role="user"`. (Verified against a saved request dump: all tool_use/tool_result pairs were adjacent and matched — zero orphans — the only defect was the leading assistant role.)

## Why

`_assemble_context` builds `[leading anchor (system)] + [summary] + [fresh tail]`. The role picker was:

```python
last_role = result[-1].get("role", "system") if result else "system"
summary_role = "assistant" if last_role != "assistant" else "user"
```

When the system prompt is the only leading anchor, `last_role == "system"` → `summary_role = "assistant"` → leading-assistant after system extraction → 400. This mirrors the same class of bug fixed in the built-in compressor upstream (NousResearch/hermes-agent #52160): the summary that lands right after the system head must be `role="user"`.

The fix pins `summary_role = "user"` only when `result` ends in a `system` message (the system-anchor case). The no-system-prompt path (gateway sessions that start with a user message) is left untouched, preserving the existing reassembly/round-trip behavior.

## Validation

- [x] Focused: `pytest tests/test_lcm_engine.py::TestEngineCompress::test_assemble_context_summary_role_is_user_after_system_anchor -q` → **1 passed** (RED on `main`: `- user / + assistant`; GREEN with fix)
- [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` → all green except one pre-existing flaky timing test (`test_on_session_end_returns_quickly_under_real_sqlite_writer_lock` — passes in isolation and on stock `main`) and two pre-existing macOS `/var`↔`/private/var` realpath failures in `test_path_*` (fail identically on unmodified `main`).
- [x] `git diff --check`

## Notes

Adds `test_assemble_context_summary_role_is_user_after_system_anchor` (the regression guard — fails on `main`, passes with the fix). Three existing assertions that snapshotted the old `assistant` role were updated; their actual intent (LCM-note injection, budget selection, and durable-row dedup on replay) is unaffected — the preserved-objective scaffold is recognized by its content prefix (`_is_replayed_context_scaffold_message`) regardless of role, so the round-trip/no-duplicate-durable-rows invariant still holds (verified directly: `get_session_count` stays constant across replay after the role flip).

Refs NousResearch/hermes-agent#52160

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -4285,7 +4285,15 @@ class LCMEngine(ContextEngine):
         # Collect DAG summaries — highest depth first for context hierarchy
         summary_parts: list[str] = []
         last_role = result[-1].get("role", "system") if result else "system"
-        summary_role = "assistant" if last_role != "assistant" else "user"
+        if result and result[-1].get("role") == "system":
+            # The system prompt is the only leading anchor, so the summary
+            # becomes the first message reaching the provider. Anthropic
+            # extracts the system prompt into a separate field and requires
+            # messages[0] to be role "user"; an assistant summary here is
+            # rejected with HTTP 400 after the second compaction.
+            summary_role = "user"
+        else:
+            summary_role = "assistant" if last_role != "assistant" else "user"
         if anchor_part is not None:
             anchor_msg = {"role": summary_role, "content": anchor_part}
             if summary_budget is None or count_message_tokens(anchor_msg) <= summary_budget:
--- a/engine.py
+++ b/engine.py
@@ -4285,12 +4285,13 @@ class LCMEngine(ContextEngine):
         # Collect DAG summaries — highest depth first for context hierarchy
         summary_parts: list[str] = []
         last_role = result[-1].get("role", "system") if result else "system"
-        if result and result[-1].get("role") == "system":
-            # The system prompt is the only leading anchor, so the summary
-            # becomes the first message reaching the provider. Anthropic
-            # extracts the system prompt into a separate field and requires
-            # messages[0] to be role "user"; an assistant summary here is
-            # rejected with HTTP 400 after the second compaction.
+        if not result or result[-1].get("role") == "system":
+            # The summary becomes the first provider-visible message: either no
+            # leading anchor exists (gateway-style assembly) or the system
+            # prompt is the only anchor, which Anthropic extracts into a
+            # separate field. Either way messages[0] must be role "user"; an
+            # assistant summary here is rejected with HTTP 400 after the second
+            # compaction.
             summary_role = "user"
         else:
             summary_role = "assistant" if last_role != "assistant" else "user"
--- a/tests/test_lcm_core.py
+++ b/tests/test_lcm_core.py
@@ -3737,8 +3737,14 @@ class TestAssemblyBudgetSelection:
         assert "KEEP_USER_DECISION" in contents
         assert "Latest compact status" in contents
         assert "oversized assistant tool chatter" not in contents
+        # The preserved objective is replayed as scaffolding (identified by its
+        # content prefix, not its role), so it must never appear as a raw
+        # user-role turn that could be re-ingested as a durable row.
         assert not any(
-            msg.get("role") == "user" and "KEEP_USER_DECISION" in str(msg.get("content", ""))
+            msg.get("role") == "user"
+            and "[Current user objective preserved from compacted history]"
+            not in str(msg.get("content", ""))
+            and "KEEP_USER_DECISION" in str(msg.get("content", ""))
             for msg in assembled
         )
 
@@ -3802,8 +3808,14 @@ class TestAssemblyBudgetSelection:
             and "KEEP_USER_DECISION" in str(msg.get("content", ""))
             for msg in assembled
         )
+        # The preserved objective is replayed as scaffolding (identified by its
+        # content prefix, not its role), so it must never appear as a raw
+        # user-role turn that could be re-ingested as a durable row.
         assert not any(
-            msg.get("role") == "user" and "KEEP_USER_DECISION" in str(msg.get("content", ""))
+            msg.get("role") == "user"
+            and "[Current user objective preserved from compacted history]"
+            not in str(msg.get("content", ""))
+            and "KEEP_USER_DECISION" in str(msg.get("content", ""))
             for msg in assembled
         )
 
--- a/tests/test_lcm_engine.py
+++ b/tests/test_lcm_engine.py
@@ -4076,7 +4076,7 @@ class TestEngineCompress:
             "[Expand for details: backed prior]"
         )
         messages = [
-            {"role": "assistant", "content": active_summary_marker},
+            {"role": "user", "content": active_summary_marker},
             {"role": "user", "content": "fresh tail question"},
             {"role": "assistant", "content": "fresh tail answer"},
         ]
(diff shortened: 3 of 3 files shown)


stephenschoettler:

fesalfayed:
For the record — this PR's fix is confirmed against a real captured failure, not just a constructed test.

A `claude-opus-4-8` session crashed on the second compaction with the exact HTTP 400 this PR fixes. The runtime auto-saved the failing request body on the non-retryable error; decoding it programmatically over all messages:

```
reason          : non_retryable_client_error
total messages  : 58
msg[0].role     : assistant      # blocks: [text, text, tool_use]
msg[1].role     : user           # blocks: [text, tool_result]  -> matches msg[0]'s tool_use
orphaned pairs  : 0              # every tool_use has its tool_result in the immediately-following message
error           : messages.2: `tool_use` ids were found without `tool_result`
                  blocks immediately after: toolu_01WzS6...
```

So with **zero** non-adjacent tool_use/tool_result pairs, the error wording is a red herring — the only structural defect is `messages[0].role == "assistant"`. That assistant turn is the DAG summary emitted by `_assemble_context`: after the system prompt is extracted into Anthropic's separate `system` field, the summary becomes the literal first `messages` entry, which Anthropic rejects unconditionally.

The trigger in the logs lines up exactly with this code path:

```
hermes_plugins.hermes_lcm.engine: LCM compaction #2: 506 messages → 65 ...
agent.conversation_loop: Non-retryable client error: Error code: 400 ...
  messages.2: `tool_use` ids were found without `tool_result` blocks
  immediately after: toolu_01WzS6...
```

Pinning `summary_role="user"` when the system prompt is the only leading anchor (this PR) makes `messages[0]` a `user` turn and the request is accepted. Same class of fix as the built-in compressor in NousResearch/hermes-agent#52167 / issue #52160.

fesalfayed:
Rebased onto current `main` (now on top of #281 `0de8264`) — clean, no conflicts (#281 touched `ingest_protection.py`, disjoint from `engine.py`). Force-pushed `6b77c7e`.

Focused validation rerun on the rebased branch:

- **Regression test** `TestEngineCompress::test_assemble_context_summary_role_is_user_after_system_anchor` — RED on the rebased base (reverting just `engine.py` to the parent → `AssertionError: - user / + assistant`), GREEN with the fix.
- `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` → **672 passed, 1 skipped**, plus one flaky timing test (`TestSessionRollover::test_on_session_end_returns_quickly_under_real_sqlite_writer_lock`) that passes in isolation and on stock `main` — it's load-dependent under the full parallel run, unrelated to this change.
- `python -m compileall -q .` ✓
- `python -m py_compile scripts/import_lossless_claw.py` ✓
- `bash -n scripts/install.sh scripts/update.sh` ✓
- `git diff --check` ✓

Ready for review on the current branch.

stephenschoettler:
Thanks for rebasing and for the captured-failure evidence. I agree the direction is right, but I think this needs one more fix before approval.

Blocker: the patch only handles the system-anchor path. `_assemble_context()` can also be called without a system anchor, and in that case this branch still emits the DAG summary as the first provider-visible message with role `assistant`, which is the same structural class this PR is trying to eliminate.

I reproduced that on current PR head `6b77c7e`:

```text
system + user tail roles= ['user', 'user'] first= user adjacent_same= True
no system + user tail roles= ['assistant', 'user'] first= assistant adjacent_same= False
system + assistant tail roles= ['user', 'assistant'] first= user adjacent_same= False
```

So the system-anchor captured failure is fixed, but the no-system/gateway-style assembly path still has a leading assistant summary.

Requested change:
- Make the role rule general: if the summary block would be the first non-system/provider-visible message, assign it role `user`; otherwise preserve alternation against the preceding visible message.
- Add a regression for `_assemble_context(None, tail)` with an existing DAG summary asserting the first non-system message is `user`.

Validation I ran on this PR head:

```text
python -m pytest -q tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py
674 passed, 46 warnings in 44.62s

python -m pytest -q
1032 passed, 60 warnings in 56.85s

python -m compileall -q .
all passed
```

Also noted: GitHub currently reports no checks on this branch, so after the update I would want fresh CI before merge.

fesalfayed:
Good catch — fixed at `0ace007`.

Made the role rule general in `_assemble_context()`: the summary block is pinned to `user` whenever it would be the first provider-visible message — both the no-leading-anchor (gateway-style) case and the system-prompt-only case. Otherwise it preserves alternation against the preceding visible message, unchanged.

```python
if not result or result[-1].get("role") == "system":
    summary_role = "user"
else:
    summary_role = "assistant" if last_role != "assistant" else "user"
```

Re-running your reproduction on the new head:

```text
system + user tail      roles= ['user', 'user']       first= user adjacent_same= True
no system + user tail   roles= ['user', 'user']       first= user adjacent_same= True
system + assistant tail roles= ['user', 'assistant']  first= user adjacent_same= False
```

The no-system path now leads with `user`.

Regression added — `test_assemble_context_summary_role_is_user_without_system_anchor` asserts `_assemble_context(None, tail)` with an existing DAG summary yields a first non-system message of role `user` (fails before this commit: `assert 'assistant' == 'user'`, passes after). Also updated `test_compress_reassembles_backed_active_summary_marker_on_noop`, whose input marker hard-coded the old leading-`assistant` role; its leading summary marker now reflects the canonical `user` role.

Validation:

```text
python -m pytest -q tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py
673 passed, 1 failed, 1 skipped

python -m compileall -q .          -> ok
bash -n scripts/install.sh scripts/update.sh  -> ok
```

The one failure is `test_on_session_end_returns_quickly_under_real_sqlite_writer_lock` — a pre-existing timing flake (`elapsed < 0.3` against a 750ms busy_timeout); it edge-passes on the first run and fails on reruns on stock head too, independent of this change. (My local full-suite run also shows two `test_path_*` failures, but those are an artifact of running from a `/tmp -> /private/tmp` symlinked worktree on macOS and also fail on stock head — not present on your clean-machine run.)

Agreed on wanting fresh CI before merge.

stephenschoettler:
Thanks, this addresses the blocker I raised.

I verified the merge result locally against current `origin/main` and the role rule now covers both provider-visible leading-summary cases:

```text
focused regressions: 2 passed
python -m pytest -q tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py
677 passed, 47 warnings

python -m pytest -q
1041 passed, 61 warnings

python -m compileall -q .
git diff --check HEAD
all passed
```

Remaining gate before merge: GitHub still shows this branch as behind current `main` and there are currently no checks reported on the PR. Please rebase once more onto current `main` and let CI run fresh. Assuming CI is green after that, I do not see another blocker.

fesalfayed:
Rebased onto current `main` — head is now `485370a`, `0` commits behind, conflict-free (the two commits that had landed, #279 doctor-paths and #285 deepcopy clone-safe, replayed cleanly under the change).

Re-verified after the rebase:

```text
focused regressions (summary_role_is_user_*, marker_on_noop): 3 passed
python -m pytest -q tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py
675 passed, 1 skipped
```

(The one local failure is `test_on_session_end_returns_quickly_under_real_sqlite_writer_lock` — the pre-existing `elapsed < 0.3` vs 750ms busy_timeout timing flake; it's green on your clean-machine run, which is why your count is 677.)

Branch is `MERGEABLE`. Over to fresh CI.

stephenschoettler (approved):
Approved. I rechecked the rebased head against current main locally and fresh CI is green. The previous leading-assistant blocker is resolved for both system-anchor and no-system assembly paths.

PR #298, opened 2026-07-04
Data-safety fixes: GC tombstone, ingest visibility, ignore-drop count, recall order

ai-ag2026:
## Summary
- Transcript GC only takes the fast externalized-ref branch when the whole row IS the placeholder (`is_externalized_placeholder`); an embedded ref now falls through to the content-equality lookup instead of tombstoning the entire row.
- Per-turn `ingest()` and `should_compress_preflight()` record persistence failures (count, last error) instead of swallowing them at debug; `get_status()` and `lcm_doctor` surface an `ingest_health` check, and failures escalate to error when consecutive.
- Preflight fails closed on ingest failure (does not compact against a store missing the latest turn) but still honors emergency overflow recovery.
- `ignore_message_patterns` drops are counted, logged at info, and surfaced in status/doctor (`ignore_pattern_drops`).
- The relevance/hybrid LIKE search fallback scans deterministically recent-first up to the candidate cap instead of scoring an arbitrary storage-order slice with no `ORDER BY`.

## Why
- The core guarantee is that nothing is ever lost, but a tool row that merely quotes an externalized placeholder (e.g. output of the LCM recall tools) had its surrounding, never-externalized text discarded by GC.
- A failing store (disk full, lock beyond busy-timeout, corruption) previously stopped persisting turns silently, with only a debug log; the loss must be operator-visible and must not let the prompt overflow the provider.
- An over-broad `ignore_message_patterns` silently discarded substantive turns from the "lossless" store; drops should at least be visible.
- Without an `ORDER BY`, relevance/hybrid recall on CJK/degraded-FTS queries never scored the best matches beyond the first storage-order page.

## Validation
- Resolved current-main merge conflicts on branch head `f66c290`.
- Addressed Codex P2 on LIKE fallback ordering: SQL candidate cap now orders by relevance/hybrid primary score before exact whole-query tie-breaks, avoiding single-term exact rows crowding out stronger multi-term matches.
- Focused local validation with host stub: CJK LIKE fallback relevance tests, exact whole-query/multi-term regression tests, sensitive-ingest search regression, and ingest-failure status tests passed.
- CI green on head `f66c290`: workflow-lint and Python 3.11, 3.12, 3.13, 3.14.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -34,6 +34,7 @@ from .externalize import (
     build_transcript_gc_placeholder,
     extract_externalized_ref,
     find_externalized_payload_for_message,
+    is_externalized_placeholder,
     load_externalized_payload,
     maybe_externalize_tool_output,
 )
@@ -465,6 +466,13 @@ class LCMEngine(ContextEngine):
             self._config.ignore_message_patterns
         )
         self._ignored_message_count: int = 0
+        # Raw messages permanently dropped because they matched
+        # ignore_message_patterns. These are NOT persisted anywhere, so an
+        # over-broad operator pattern silently discards substantive turns from
+        # the "lossless" store. Count + log them so the loss is at least
+        # visible; full lossless retention (store with ignored=1) is a larger
+        # follow-up that touches cursor reconciliation and FTS.
+        self._ignore_pattern_dropped_count: int = 0
 
         # Track which store_ids have been ingested into the DAG
         self._last_compacted_store_id: int = 0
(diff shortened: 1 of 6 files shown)


stephenschoettler:

Please add:
- Summary
- Why
- Validation

stephenschoettler:
@codex review

Current head: `7035dd9a5962`.

ai-ag2026:
@codex review Current head: 64ecdcc4a68c.

Tosko4:
@ai-ag2026 This looks aligned with the LCM direction and the core fixes are good.

One small thing before this should be considered ready: `ignore_pattern_dropped_count` is exposed through `engine.get_status()` and `lcm_doctor`, but it does not appear in the public `lcm_status` payload. Since the PR says these drops are surfaced in status/doctor, please also include the count in `lcm_status` (for example under `session_filters`) and add/adjust a small test for that surface.

Also note: the latest real Codex review I can see was on `7035dd9a59`; current head is `64ecdcc4a68c`. After the status-surface tweak, it would be good to get a real current-head Codex pass rather than relying on the older review.

stephenschoettler:

Tosko4:
@ai-ag2026 I saw the Codex connector reply that it is not configured for your account, so I am triggering the current-head review from my side.

Tosko4:
@codex review

ai-ag2026:
@codex review

Tosko4:
@codex review

stephenschoettler (approved):

PR #299, opened 2026-07-04
Perf: memoize token counts and skip unchanged metadata commits

ai-ag2026:
## Summary
- Add an LRU cache to `count_tokens`, keyed on the string; non-string/unhashable inputs (e.g. `tool_call` arguments passed as a dict) are counted uncached to preserve the existing tolerance, and strings over 32KB bypass the cache to bound its memory.
- Make the char-based token fallback (used when `tiktoken` is absent) non-ASCII-aware so CJK/other dense scripts are not undercounted ~3-4x.
- Read-before-write the generated-ignored placeholder count/ordinal metadata: skip the UPSERT (and its fsync commit under `synchronous=FULL`) when the stored value already matches.

## Why
- Assembly and preflight re-count the same message content many times per turn, and the assembler re-measures candidates while packing to budget (~O(n^2) on tool-heavy transcripts); the encoder is decided once per process, so a content-keyed cache is stable and behavior-preserving.
- An undercounted CJK context makes preflight under-trigger and assembly overflow the real budget.
- The placeholder metadata was UPSERT+committed unconditionally on every ingest, i.e. up to two fsync commits per turn even when nothing changed.

## Validation
- CI green after fix commit `bb0e016` across workflow-lint and Python 3.11-3.14 matrix

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -3575,7 +3575,16 @@ class LCMEngine(ContextEngine):
                 return
             serialized = json.dumps(payload, sort_keys=True)
             with self._store._write_lock:
+                wrote = False
                 for key in count_keys:
+                    # Skip the write (and its fsync commit under synchronous=FULL)
+                    # when the stored value already matches. This runs on every
+                    # ingest and was previously an unconditional UPSERT+commit.
+                    existing = conn.execute(
+                        "SELECT value FROM metadata WHERE key = ?", (key,)
+                    ).fetchone()
+                    if existing is not None and existing[0] == serialized:
+                        continue
                     conn.execute(
                         """
                         INSERT INTO metadata(key, value)
@@ -3584,7 +3593,9 @@ class LCMEngine(ContextEngine):
                         """,
                         (key, serialized),
                     )
-                conn.commit()
+                    wrote = True
+                if wrote:
+                    conn.commit()
         except Exception:
             logger.debug("LCM ignored placeholder count metadata write failed", exc_info=True)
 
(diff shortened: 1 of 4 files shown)


stephenschoettler:

Please add:
- Summary
- Why
- Validation

stephenschoettler:
@codex review

Current head: `04c042d5c744`.

ai-ag2026:
@codex review Current head: bb0e016195e1.

Tosko4:
@stephenschoettler for context: I think this PR is useful, but I would frame it narrowly as a token-budget/performance patch, not as a deeper LCM “understands token cutting” change.

The direction that fits LCM for me is:

- token counting must be accurate enough that preflight/assembly do not under-trigger or overflow the provider window, especially when `tiktoken` is unavailable;
- repeated token measurement on the hot path can be cached, but the cache has to stay small and incidental. It should not become an uninspectable in-memory store of old message content;
- metadata write avoidance is fine as runtime hygiene, as long as it does not change the raw-message/DAG/recovery contract.

The part I would not overclaim: this does not make LCM semantically understand clipped content. That still belongs in the core LCM contract: token-aware chunking, stable source lineage, offsets/provenance/reason metadata where we split, and recoverability through expand/search back to the original raw material.

One concrete blocker before merge: the current cache cap looks accidental:

```py
_MAX_CACHEABLE_TOKEN_TEXT_CHARS=4096_768
```

That is about 4MB per string, not 32KB. With a 2048-entry LRU that can retain roughly 7.8GiB of string payloads in the worst case. The new oversized-string test derives the payload size from the constant, so it will not catch the intended 32KB boundary regressing.

I’d fix this by pinning the cap explicitly, e.g. `32 * 1024` / `32_768`, and adding a literal boundary test for 32KB vs 32KB+1. After that, I think the PR is directionally fine as a bounded budget/perf improvement.

Local validation I ran on current head `bb0e016195e1`:

- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py -q -o addopts=` → 784 passed
- `python -m pytest -q -o addopts=` → 1185 passed
- `git diff --check origin/main...HEAD` → clean

stephenschoettler:

Tosko4:
@codex review

ai-ag2026:
@codex review

Tosko4:
@codex review

stephenschoettler (approved):
Approved for merge under Stephen approval: reviewer gate t_9eb57f4e passed final head; CI is green and no unresolved current material threads.

PR #301, opened 2026-07-04
Harden ingest protection: ReDoS-safe redaction, wrapped base64, path containment

ai-ag2026:
## Summary
- Route only the `private_key` sensitive pattern (lazy `.*?` under DOTALL) through the optional `regex` engine with a match timeout, or a length bound when `regex` is absent; the other patterns are character-class-bounded and linear, so they always run via stdlib and never fail open.
- Detect and externalize line-wrapped base64 (MIME 76 / PEM 64 chars per line), and compute the base64 ratio over whitespace-stripped content so canonical CRLF line endings do not sink it below threshold.
- Warn (once) when a configured externalized-payload path escapes the `hermes_home` base and `LCM_HERMES_BASE_DIR` is unset; the strict raise still applies when the base is set.

## Why
- The `private_key` pattern rescans to end-of-string for every unmatched `BEGIN` header, so a multi-MB payload with many headers and no `END` blocked ingest for seconds-to-minutes; restricting fail-open to that one pattern avoids a silent secret-redaction bypass on the linear patterns under load.
- Line-wrapped base64 never forms a single 4096-char contiguous run, so it slipped past the externalizer and landed inline in SQLite/FTS/WAL/backups; CRLF (the canonical MIME case) additionally failed the ratio gate.
- Default deployments had no path containment at all; a warning surfaces an escaped path without breaking legitimate custom volumes.

## Validation
- `PYTHONPATH=/tmp/hermes-lcm-test-stubs python -m pytest tests/test_ingest_protection.py -q` (128 passed)
- `python -m py_compile ingest_protection.py tests/test_ingest_protection.py`
- `git diff --check`

### DIFF
--- a/externalize.py
+++ b/externalize.py
@@ -68,6 +68,22 @@ def _missing_directory_components(path: Path) -> list[Path]:
     return list(reversed(missing))
 
 
+_WARNED_EXTERNALIZATION_PATHS: set[str] = set()
+
+
+def _warn_externalization_path_outside_base(path: Path, allowed_base: Path) -> None:
+    key = str(path)
+    if key in _WARNED_EXTERNALIZATION_PATHS:
+        return
+    _WARNED_EXTERNALIZATION_PATHS.add(key)
+    logger.warning(
+        "LCM externalized-payload path %s is outside the hermes_home base %s; "
+        "set LCM_HERMES_BASE_DIR to enforce strict containment",
+        path,
+        allowed_base,
+    )
+
+
 def get_large_output_storage_dir(config, hermes_home: str = "", *, create: bool) -> Path:
     configured = getattr(config, "large_output_externalization_path", "") or ""
     if configured:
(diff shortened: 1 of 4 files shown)


stephenschoettler:

Please add:
- Summary
- Why
- Validation

stephenschoettler:
@codex review

Current head: `265bf5e760bd`.

stephenschoettler:
@codex review

Current head: `52bfb3dcbb0e`.

ai-ag2026:
@codex review Current head: df094e10dc75.

Tosko4:
@codex review

Tosko4:
@stephenschoettler FYI, I did a local maintainer-style pass on the current head (`df094e10dc75`).

My read: the direction fits Hermes-LCM well. This is storage-boundary hardening in the right layer: keep large/base64 payloads out of SQLite/FTS/WAL/backups while preserving recoverability, make sensitive redaction safe against pathological private-key inputs, and surface escaped externalized-payload paths without breaking existing custom-volume deployments.

I also checked the previous Codex findings against the current diff. The current code appears to address the important cases: wrapped-base64 scanning is linear, large PEM-containing inputs still redact without the optional `regex` package, timeout fallback keeps private-key redaction, and SHA/hex inventories are not treated as wrapped base64.

Local validation I ran:
- `PYTHONPATH=/tmp/hermes-lcm-test-stubs python -m pytest tests/test_ingest_protection.py tests/test_path_containment.py -q -o addopts=` — 136 passed
- `PYTHONPATH=/tmp/hermes-lcm-test-stubs python -m pytest -q -o addopts=` — 1190 passed
- `git diff --check`
- `python -m py_compile ingest_protection.py externalize.py tests/test_ingest_protection.py tests/test_path_containment.py`
- Small manual probes for long single-line base64-like text, wrapped base64, and large PEM redaction without `regex`

I am not treating this as a final ready/merge recommendation yet because I just re-triggered Codex on the current head and the branch is currently reported as behind the base. But purely on architecture and code quality, this looks aligned with the Hermes-LCM model to me.

stephenschoettler:

ai-ag2026:
@codex review

stephenschoettler (approved):
Verified head d2c33837f4f3 through the internal review gate. Approving for merge.

stephenschoettler (approved):
Verified head 9fe86d77da1c through the internal review gate and post-update CI. Approving for merge.

PR #302, opened 2026-07-04
Add a summary spend guard and a schema-too-new open guard

ai-ag2026:
## Summary
- Add `SummarySpendGuard`: a per-scope sliding-window rate limit with backoff for summarizer calls. When the budget is exhausted, escalation falls back to deterministic L3 truncation; the guard is checked per-route so a mid-chain trip stops the remaining fallbacks.
- The guard is deliberately NOT cleared on the automatic `force_overflow` path, and is wired into both the leaf and condensation summarization call sites. Config via optional `getattr` fields; `summary_spend_max_calls=0` disables it.
- Add `SchemaVersionTooNewError`: `run_versioned_migrations` refuses to open a database whose stored `schema_version` exceeds this build's `SCHEMA_VERSION`.

## Why
- The existing circuit breaker reacts only to summarizer failures; a pathologically looping compaction that keeps succeeding can burn auxiliary-model spend without bound. `force_overflow` is the sustained-over-cap state such a loop produces, so clearing the guard there would defeat it; a tripped guard still converges the emergency via L3 (no spend).
- Opening a v-newer database with older code and silently migrating/degrading it risks corrupting data written under semantics this build does not understand; refusing is safer than degrading after a downgrade/rollback.

## Validation
- `python -m pytest tests/test_db_bootstrap_fts.py -q` (16 passed)
- `python -m py_compile db_bootstrap.py tests/test_db_bootstrap_fts.py`
- `git diff --check`


stephenschoettler:

Please add:
- Summary
- Why
- Validation

stephenschoettler:
@codex review

Current head: `5968eaa8c234`.

stephenschoettler:
@codex review

Current head: `6ef368552c59`.

ai-ag2026:
@codex review Current head: d0eb853210c3.

Tosko4:
@stephenschoettler for visibility: I took a maintainer pass over this PR against current head `d0eb853210c38699a81254083aa214381b1b9fa3`.

The direction fits LCM well. Bounding runaway summarizer spend is a good complement to the existing failure circuit breaker, and refusing future DB schemas is the right default for a storage layer that should preserve recoverability instead of silently degrading after a rollback/downgrade.

Local validation on the exact head:
- `python3 -m pytest tests/test_db_bootstrap_fts.py tests/test_lcm_core.py -q -o addopts=` → 243 passed
- `python3 -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=` → 806 passed
- `python3 -m pytest -q -o addopts=` → 1187 passed
- `git diff --check` → clean
- added-line static scan for obvious secrets/shell/eval/pickle/SQL-formatting issues → clean

A few things I would fix or clarify before calling this merge-ready:

1. The PR body says `summary_spend_max_calls=0` disables the guard, but `LCMConfig` does not define these fields and `from_env()` does not parse `LCM_SUMMARY_SPEND_*`. I verified that setting `LCM_SUMMARY_SPEND_MAX_CALLS=0` still leaves the guard at the default `24` calls. Either expose real config fields/env parsing, or remove the configurability claim from the body/comments.

2. `SummarySpendGuard`'s docstring says forced/manual compaction calls `clear()`, but I do not see any `_summary_spend_guard.clear()` call in the codebase. Either implement/test the operator/manual bypass, or adjust the docstring so it does not promise behavior that is not wired.

3. The future-schema guard now prevents startup DDL/table creation, which is good. One small caveat: the startup paths still call `configure_connection()` before the refusal check, so a future-schema DB can still have PRAGMAs applied first, including `journal_mode=WAL`. That may be acceptable if the guarantee is specifically “no schema/DDL writes”, but I would make that boundary explicit.

Also, current GitHub state shows CI green, but the latest current-head `@codex review` did not produce a real Codex review; it returned the connector setup message instead. So I would not treat this head as Codex-clean yet.

Net: good idea, good fit, tests are strong, but I would clean up the config/operator contract and the stale docstring before merge.

stephenschoettler:

Tosko4:
@codex review

ai-ag2026:
@codex review

stephenschoettler (approved):
Verified head e24c54059d72 through the internal review gate and post-update CI. Approving for merge.

PR #303, opened 2026-07-04
Add a steady-state per-turn hot-path benchmark

ai-ag2026:
## Summary
- Add `benchmarking/steady_state.py` and `scripts/lcm_steady_state_bench.py`: replay turns WITHOUT compacting (very high threshold + context length) and report per-turn `ingest`/`preflight` p50/p95 and row-writes-per-turn at a range of history sizes, under baseline / `ignore_message_patterns` / `sensitive_patterns` configs.
- Purely additive: three new files, no existing code touched. Adds 3 tests.

## Why
- The existing replay/stress benchmarks time `compress()`. Nothing measured the path that runs on every turn regardless of compaction: the post-LLM `ingest()` hook plus `should_compress_preflight()`, which re-scans the active history each turn.
- Intended as a regression guard so a per-turn cost that scales with conversation length is visible. On the current code the ignore-patterns and sensitive-patterns configs already show clear growth with history size, which a future ingest-path optimization should flatten.

## Validation
- CI green across workflow-lint and Python 3.11-3.14 matrix

### DIFF
--- a/benchmarking/steady_state.py
+++ b/benchmarking/steady_state.py
@@ -169,51 +169,53 @@ def _measure_case(
     history_sizes: tuple[int, ...],
     iterations: int,
 ) -> list[SteadyStateSample]:
-    engine = _build_engine(case, run_dir)
-    history: list[dict[str, Any]] = []
-    next_index = 0
     samples: list[SteadyStateSample] = []
 
     for target in sorted(history_sizes):
-        # Grow the history to the target size (ingesting as we go so the store
-        # tracks it, mirroring how turns accumulate in production).
-        while len(history) < target:
-            history.extend(_synthetic_turn(next_index))
-            next_index += 1
-            engine.ingest(history)
+        target_run_dir = run_dir / f"{case.name}-{target}"
+        engine = _build_engine(case, target_run_dir)
+        history: list[dict[str, Any]] = []
+        next_index = 0
+        try:
+            # Build an independent history for each target so a nearby smaller
+            # target cannot pre-populate store/identity maps for the next one.
+            while len(history) < target:
+                history.extend(_synthetic_turn(next_index))
+                next_index += 1
+                engine.ingest(history)
 
-        ingest_ms: list[float] = []
-        preflight_ms: list[float] = []
-        writes: list[int] = []
-        for _ in range(iterations):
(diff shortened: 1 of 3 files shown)


stephenschoettler:

Please add:
- Summary
- Why
- Validation

ai-ag2026:
@codex review Current head: bd8e9e27307f.

Tosko4:
Directionally this fits the project well: measuring the per-turn ingest + preflight hot path is useful, and the change is nicely additive.

I would fix a few things before merging:

1. The `ignore_patterns` case can false-green in the default/minimal environment. The repo treats `regex` as optional, CI only installs `pytest`, and without `regex` message-level ignore matching is disabled with a warning. In that setup this benchmark labels the case as `ignore_patterns` while not actually exercising active ignore-pattern matching. Please either make the benchmark/test environment enable `regex`, or detect that the feature is inactive and label/skip the case.

2. The reported `history_size` is post-measurement, not the requested/sample-start size. For example a target of 20 with 2 iterations reports 24. Also, the first measured turns after growing to a new target can include warm-up artifacts in row writes. Please report the target/sample-start size explicitly and consider a short warm-up before collecting samples.

3. `test_steady_state_does_not_compact` does not currently assert that compaction never ran. It would be stronger to assert `compression_count == 0` or spy that `compress()` is not called.

4. Minor output nit: the text report prints values like `0.210m`; these are milliseconds, so `ms` would be clearer.

Local validation on PR head `bd8e9e27307f813d069d4b35a47648105266ec8e`:
- `git diff --check` passed
- static added-line scan found no obvious secrets/shell/eval/pickle/SQL issues
- `python3 -m pytest tests/test_benchmarking_steady_state.py tests/test_benchmarking_cli.py -q -o addopts=` -> 14 passed
- `python3 -m pytest -q -o addopts=` -> 1182 passed

@stephenschoettler FYI — good direction, but I would hold merge until the benchmark cases are actually measuring the paths they claim to measure.

stephenschoettler:

Tosko4:
@codex review

ai-ag2026:
@codex review

stephenschoettler (approved):
Approved for merge under Stephen approval: reviewer gate t_9eb57f4e passed final head; CI is green and no unresolved current material threads.

PR #313, opened 2026-07-05
Fix strict persisted-output replay identity

Tosko4:
## Summary
- Adds strict persisted-output replay identity for Hermes tool-output markers.
- Requires live-file/stat-backed proof before a recovered persisted-output marker can satisfy durable replay identity.
- Treats missing or unrecoverable raw persisted-output markers as append-worthy instead of replay-proof, even when path/preview/length or inline-looking metadata matches.
- Preserves safe file-generation metadata for recovered persisted-output payloads while avoiding raw content hashes and lossy preview digest leaks.
- Expands regression coverage for forged inline metadata, stale durable payloads, externalization-disabled fallback, legacy preview-prefix handling, and lossy redaction retries.

## Why
- Persisted-output markers are only trustworthy while the backing Hermes temp file can be safely recovered and matched.
- Inline marker text is untrusted preview content; treating it as proof after the file is gone can silently drop legitimate retried tool results.
- The safer replay contract is: dedupe only with real proof, otherwise append rather than lose data.

## Validation
- `python -m pytest tests/test_lcm_core.py::TestIngestExternalization -q -o addopts=` → `49 passed`
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=` → `927 passed`
- `python -m pytest -q -o addopts=` → `1400 passed, 12 xfailed`
- `python -m compileall -q engine.py externalize.py ingest_protection.py tests/test_lcm_core.py && git diff --check origin/main...HEAD && git diff --check` → clean
- Static scan of `engine.py`, `externalize.py`, and `ingest_protection.py` for `persisted_output_content_sha256` / `content_sha256` → `STATIC_SCAN_CLEAN`
- Local Codex CLI review with `model_reasoning_effort="xhigh"` on the final branch diff → `No blockers found.`

## Risk / impact
- Replay behavior is intentionally stricter for persisted-output markers whose backing file is missing or unrecoverable: these markers append instead of deduping, to avoid silent data loss.
- This can create duplicates in ambiguous replay cases, but only where the branch cannot prove identity safely.
- Recovered persisted-output payload metadata now carries file generation fields (`size`, `mtime_ns`, `ctime_ns`) for non-content identity checks.
- No schema migration or config change is required.
- The implementation avoids raw full-content hashes and avoids preview-derived digests for lossy redaction cases.

## Scope boundaries
- Does not change Hermes host-side tool result persistence.
- Does not introduce provenance-signed inline trailers; inline marker text remains untrusted.
- Does not try to dedupe missing-file persisted-output markers without live-file proof.
- Does not change unrelated compaction, summarization, or tool APIs.

## Screenshots / before-after
- Not applicable (non-UI change).

## Rollout / operator notes
- No migration steps required.
- Existing externalized payloads remain compatible.
- Legacy preview-prefix metadata is still handled where safe, but lossy redaction cases avoid deriving replay identity from raw legacy preview text.
- Operators should expect safer append behavior for ambiguous missing-file persisted-output replays.

## Reviewer focus
- Please focus on the replay gates around persisted-output markers in `engine.py`.
- Check that missing/unrecoverable temp files cannot become replay-proof through forged inline marker identity/generation text.
- Check that live-file proof, file-generation metadata, redaction handling, and legacy preview-prefix compatibility stay aligned across `ingest_protection.py`, `externalize.py`, and the regression tests.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -34,6 +34,7 @@ from .escalation import (
 )
 from .externalize import (
     build_transcript_gc_placeholder,
+    externalized_tool_result_has_persisted_output_marker,
     extract_externalized_ref,
     find_externalized_payload_for_message,
     find_externalized_tool_result_content_for_call,
@@ -48,9 +49,16 @@ from .extraction import (
     strip_injected_context_blocks,
 )
 from .ingest_protection import (
+    _add_inline_persisted_output_generation_metadata,
+    _add_inline_persisted_output_identity_metadata,
     _expected_persisted_output_chars,
+    _has_inline_persisted_output_generation_metadata,
+    _has_lossy_sensitive_redaction,
+    _inline_persisted_output_generation_metadata,
     _is_hermes_persisted_output_marker,
     _json_has_duplicate_object_keys,
+    _persisted_output_inline_preview_sha256,
+    _persisted_output_marker_identity_digest,
     _persisted_output_preview_prefix_digest,
     _persisted_output_saved_path,
     assistant_output_quarantine_reason,
@@ -60,6 +68,7 @@ from .ingest_protection import (
     protect_messages_for_ingest,
     quarantine_suspicious_assistant_messages,
     recover_hermes_persisted_output,
+    recover_hermes_persisted_output_with_file_stat,
     redact_sensitive_text,
     redact_sensitive_value,
     restore_ingest_payload_placeholders,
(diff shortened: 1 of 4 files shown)


Tosko4:
@codex review

Tosko4:
@stephenschoettler please review

stephenschoettler (approved):

PR #316, opened 2026-07-05
Move metadata JSON read/write onto MessageStore

ai-ag2026:
## Summary
- Add two `MessageStore` methods that own the metadata table's key/value JSON access:
  - `read_metadata_json(key)` -> JSON-decoded value, or `None` for a closed connection / missing / empty row. Decoding is intentionally left unwrapped so a malformed value still raises.
  - `write_metadata_json(keys, serialized, *, skip_unchanged=False)` -> UPSERT the pre-serialized payload to each key under the write lock with a single commit; `skip_unchanged=True` keeps the hot-path "skip write+commit when unchanged" optimization.
- Route all eight engine metadata helpers (`_load_hash_list_for_metadata_keys`, `_remember_hash_for_metadata_keys`, and the placeholder count/ordinal + dependent-reply load/write pairs) through them.

## Why
- Those eight helpers each re-implemented the same SQLite boilerplate: fetch `self._store._conn`, guard it, `SELECT value FROM metadata`, `json.loads`; and on writes acquire `self._store._write_lock`, UPSERT per key, commit (with the count/ordinal writers additionally skipping the write when unchanged, per #299).
- That duplicated ~110 lines and reached into the store's **private** connection and lock from the engine. Consolidating the boilerplate on `MessageStore` removes the duplication and the layering violation — the engine no longer touches `self._store._conn` / `self._store._write_lock` at all.

## Validation
- [x] Focused validation:
  - [x] Direct helper behaviour checks (roundtrip; `skip_unchanged` False->wrote / True->no-op; missing key -> `None`; empty value -> `None`; malformed JSON propagates) -> all pass
  - [x] `PYTHONPATH="<stub>:." python3 -m pytest tests -q -k "metadata or placeholder or ignored or dependent or ordinal"` -> `182 passed`
- [x] Default validation:
  - [x] `PYTHONPATH="<stub>:." python3 -m pytest tests -q` -> `1385 passed, 12 xfailed` (identical to pre-change)
  - [x] `python3 -m compileall -q .`
  - [x] `python3 -m py_compile scripts/import_lossless_claw.py`
  - [x] `git diff --check origin/main...HEAD` -> clean
- [x] Workflow validation: not applicable; no workflows changed.
- [x] Added-line private/secret hygiene scan -> `0` hits.

## Notes
- Behaviour-preserving: signatures, return values, per-caller error scoping, and the unchanged-skip commit semantics are all unchanged. `read_metadata_json` deliberately does not swallow `json` errors so each loader keeps its historical abort-vs-skip behaviour (the hash-list loader discards a partial result on error; the count/ordinal loaders keep what they accumulated).
- Validation used a minimal local `agent.context_engine` stub because this plugin repository is tested standalone outside the Hermes host package.
- Net `engine.py` is 111 lines lighter; `store.py` gains the two documented helpers.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -4714,19 +4714,10 @@ class LCMEngine(ContextEngine):
         if not keys:
             return []
         try:
-            conn = self._store._conn
-            if conn is None:
-                return []
             ordered: list[str] = []
             seen: set[str] = set()
             for key in keys:
-                row = conn.execute(
-                    "SELECT value FROM metadata WHERE key = ?",
-                    (key,),
-                ).fetchone()
-                if not row or not row[0]:
-                    continue
-                data = json.loads(str(row[0]))
+                data = self._store.read_metadata_json(key)
                 if isinstance(data, list):
                     for item in data:
                         digest = str(item)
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):

PR #323, opened 2026-07-06
Encapsulate MessageStore connection behind a public accessor

ai-ag2026:
## Summary
- Add a public `MessageStore.connection` property (read-oriented) plus `MessageStore.commit()` and `MessageStore.backup(dest)`.
- Route every remaining `engine._store._conn` reach-in in `command.py` and `tools.py` through the new public surface.
- Finishes the WS5 encapsulation started in #316 (metadata JSON read/write) for the store's raw connection.

## Why
After #316 moved the metadata JSON read/write onto `MessageStore`, the `/lcm`
command handlers and the inspect/doctor tools were the last callers still
poking the store's private `_conn`: ad-hoc diagnostic reads (integrity /
quick / journal-mode PRAGMAs, FTS sync counts, schema health, payload-risk
scans, inspect totals) and the backup path's flush + snapshot. That coupled
diagnostics and the CLI to the store's private field and its name. Giving the
store a documented public seam removes that coupling and keeps the None/closed
handling in one supported place, mirroring the already-public `db_path`.

## Validation
- `ruff check store.py command.py tools.py` → All checks passed
- `python -m pytest -q -o addopts=` (full suite) → 1545 passed, 12 xfailed
- `python -m pytest tests/test_lcm_command.py tests/test_lcm_inspect.py tests/test_tool_contracts.py tests/test_lcm_core.py -q -o addopts=` → 385 passed

## Notes
- Behaviour-preserving: `connection` returns the same object callers used
  before, so their existing `if conn is None` guards are unchanged; `commit`
  and `backup` are thin wrappers that keep the live-connection requirement (a
  closed store still raises, matching direct `_conn.commit()` / `_conn.backup()`
  use).
- Scope is deliberately the `MessageStore` connection only. The `_dag` and
  `_lifecycle` connections are separate classes with their own `_conn`; the
  backup flush in `_flush_engine_connections` still commits those directly.
- Trade-off, stated plainly: `connection` is a *read* seam for diagnostics, not
  full method-level encapsulation, so it does technically expose a handle a
  caller could write through outside `_write_lock`. That is enforced by
  convention (docstring) rather than the type system. It is not a new capability
  — callers already reached the same `_conn` directly — and writes in-tree still
  go through the store's locked methods. Happy to add per-check read methods
  instead if you'd prefer the connection stay private.
- `engine.py`'s two `_store._append_protected_batch(...)` calls are a private
  *method* on the append path, a different encapsulation question, and are left
  as-is.
- The one hand-rolled `FakeStore` test double gains a `connection` property to
  mirror the real store's public API; other tests set/read the real store's
  `_conn` field directly, which the property returns unchanged.

## Refs
- Follows #316 (metadata JSON on MessageStore); same WS5 encapsulation line.

### DIFF
--- a/command.py
+++ b/command.py
@@ -221,7 +221,7 @@ def _status_text(engine) -> str:
 
 
 def _scan_clean_candidates(engine) -> dict[str, Any]:
-    conn = engine._store._conn
+    conn = engine._store.connection
     try:
         rows = conn.execute(
             """
@@ -306,7 +306,7 @@ def _scan_clean_candidates(engine) -> dict[str, Any]:
 
 
 def _scan_retention_candidates(engine) -> dict[str, Any]:
-    conn = engine._store._conn
+    conn = engine._store.connection
     now = datetime.now().timestamp()
     # SQL is scoped to the foreground session so /lcm doctor retention
     # reports the operator's real conversation rather than whatever side
@@ -467,7 +467,7 @@ def _flush_engine_connections(engine) -> None:
     ``_rotate_backup_database`` (rolling backup) so the connection-flush
     contract stays in one place.
     """
-    engine._store._conn.commit()
+    engine._store.commit()
     engine._dag._conn.commit()
     lifecycle_conn = getattr(getattr(engine, "_lifecycle", None), "_conn", None)
     if lifecycle_conn is not None:
@@ -493,7 +493,7 @@ def _backup_database(engine) -> dict[str, Any]:
 
         dest = sqlite3.connect(str(backup_path))
         try:
-            engine._store._conn.backup(dest)
+            engine._store.backup(dest)
         finally:
             dest.close()
     except (OSError, sqlite3.Error) as exc:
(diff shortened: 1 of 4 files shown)


Tosko4 (approved):
Reviewed at `2823d86423470227efc37993845bd98bef9c3baf`.

No blockers from me.

What I checked:
- The diff is narrow: it adds `MessageStore.connection`, `MessageStore.commit()`, and `MessageStore.backup(dest)`, then routes the `command.py` / `tools.py` store-connection reach-ins through that public surface.
- Local exact-head validation: `git diff --check origin/main...HEAD`, `ruff check store.py command.py tools.py`, focused suite (`tests/test_lcm_command.py tests/test_lcm_inspect.py tests/test_tool_contracts.py tests/test_lcm_core.py`) -> `385 passed`, full suite -> `1545 passed, 12 xfailed`.
- The PR is behind current `main`, so I also tested a local no-commit merge with `origin/main`; it merged cleanly and the focused suite still passed (`385 passed`).

I’m comfortable with this as a small WS5 connection-seam cleanup. One follow-up I would keep separate, not a blocker here: `scripts/import_lossless_claw.py` still uses `store._conn` directly, so if the end state is no direct `MessageStore` connection reach-ins outside the store, that script should get its own pass. This PR’s stated `command.py` / `tools.py` scope is fine.

@stephenschoettler heads-up: from my side this looks valid after the branch is updated/rebased as needed by branch protection.

PR #325, opened 2026-07-06
Extract active-context sanitizers into sanitize.py (WS5 seam 2)

ai-ag2026:
## Summary
- Add `sanitize.py` and move LCMEngine's pure active-context sanitization helpers into it.
- `engine.py` imports the three helpers with external callers and calls them as module functions.

## Why
Seam 2 of the WS5 `LCMEngine` decomposition. The active-context sanitizers are
pure functions with no engine state — they strip internal/reasoning content
from assistant messages, decide whether a message still has visible content,
and detect sensitive-redaction markers. Pulling them into `sanitize.py` removes
~160 lines from the god-class and gives them a small, dependency-light home.

Moved (verbatim): `_contains_sensitive_redaction`, `_structured_part_text`,
`_structured_part_has_visible_assistant_content`,
`_assistant_message_has_visible_content`, `_strip_structured_text_part`,
`_sanitize_active_assistant_content`, `_clean_active_assistant_message`,
`_should_drop_active_assistant_message`, and the `_VISIBLE_TEXT_PART_TYPES` /
`_INTERNAL_ASSISTANT_PART_TYPES` constants.

## Validation
- `ruff check engine.py sanitize.py` → All checks passed
- `python -m pytest -q -o addopts=` (full suite) → 1545 passed, 12 xfailed (identical to base)

## Notes
- Behaviour-preserving: bodies move verbatim. The only edits are dropping the
  `@staticmethod`/`@classmethod` decorators and `cls`/`self` params, and
  rewriting the cluster's own `cls._x(...)` cross-calls to bare module calls.
  Confirmed character-for-character against `origin/main` by an independent diff.
- No import cycle: `sanitize.py` depends only on
  `escalation._strip_reasoning_blocks` and itself; `escalation` imports neither
  `engine` nor `sanitize`. The now-unused `_strip_reasoning_blocks` import was
  dropped from `engine.py`.
- Deliberately left on the engine (state/config/orchestration coupling, for
  later seams): the `_sanitize_active_context_messages` orchestrator,
  `_sanitize_tool_pairs` (couples to the broadly-used module-level
  `_tool_call_id`, and is the only sanitizer with direct white-box tests),
  the preserved-objective helpers (need config/session_id/hermes_home), and
  `_redact_active_replay_messages` (mutates a per-turn message-id set).
- No test references any moved symbol (only `_sanitize_tool_pairs`, which
  stays), so no test changes were needed.

## Refs
- Design: WS5 engine decomposition, seam 2 (Sanitization → `sanitize.py`).
- Independent of #323/#324 (those touch store/command/tools/maintenance);
  this touches only `engine.py` + new `sanitize.py`.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -29,7 +29,6 @@ from .diagnostics import _enforce_state_db_containment
 from .escalation import (
     SummaryCircuitBreaker,
     SummarySpendGuard,
-    _strip_reasoning_blocks,
     summarize_with_escalation,
 )
 from .externalize import (
@@ -81,6 +80,11 @@ from .schemas import (
     LCM_INSPECT,
     LCM_LOAD_SESSION,
     LCM_STATUS,
+)
+from .sanitize import (
+    _clean_active_assistant_message,
+    _contains_sensitive_redaction,
+    _should_drop_active_assistant_message,
 )
 from .session_patterns import (
     build_session_match_keys,
@@ -250,17 +254,6 @@ def _is_codex_gpt55_route(model: str | None, provider: str | None) -> bool:
 _AUTO_FOCUS_MAX_TURNS = 3
 _AUTO_FOCUS_TURN_MAX_CHARS = 260
 _AUTO_FOCUS_MAX_CHARS = 700
-_VISIBLE_TEXT_PART_TYPES = {"text", "input_text", "output_text"}
-_INTERNAL_ASSISTANT_PART_TYPES = {
-    "analysis",
-    "chain_of_thought",
-    "internal",
-    "reasoning",
-    "redacted_thinking",
-    "scratchpad",
-    "thought",
-    "thinking",
-}
 
 
 def _strip_metadata_scalar(value: str) -> str:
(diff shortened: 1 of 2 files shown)


Tosko4 (approved):
I reviewed this against the current PR head (`71d353e`) and I do not see a blocker.

What I checked:
- The diff is limited to `engine.py` and the new `sanitize.py`.
- The moved sanitizer helpers match the previous `LCMEngine` implementations after the expected `staticmethod` / `classmethod` to module-function rewrites.
- `ruff check engine.py sanitize.py` passes.
- Full local suite passes: `1545 passed, 12 xfailed`.
- The PR is currently behind `main`, but a no-commit merge of current `origin/main` applied cleanly, and `tests/test_lcm_engine.py` still passed there (`696 passed`).
- No existing review threads/comments to reconcile.

My read: this is a valid, small decomposition PR. It does not change the active-context sanitizer behavior; it just moves the pure helper cluster out of the engine. The only operational note is that it should be updated onto current `main` before merge so CI runs on the final integration head.

@stephenschoettler FYI: this one looks safe to me from the maintainer side after the checks above. I would still keep the final merge call with you, especially given the surrounding WS5 PR stack.

PR #322, opened 2026-07-06
fix: harden auxiliary lifecycle edge cases

Tosko4:
## Summary
- harden auxiliary/thread-stateless lifecycle handling after PR #312
- preserve live auxiliary generations and native fallback state across stale/no-frame callbacks
- add regressions for reused IDs, retired generations, guarded handoffs, inferred side-channel children, and fallback reset edge cases

## Why
PR #312 is merged, but local adversarial review found more same-session-ID lifecycle edges where stale auxiliary callbacks could still clobber live usage/fallback state or drop normal suffix handling. This follow-up keeps the no-LCM-write contract for auxiliary/stateless side channels while making the host fallback path safer.

## Validation
- `python -m ruff check engine.py tests/test_lcm_engine.py`
- `python -m pytest tests/test_lcm_engine.py tests/test_lcm_core.py -q -k "thread_context_stateless or auxiliary or bypass or stateless or ignored or host_fallback or native_fallback or session_end or compression_boundary or reused or stack_marked or state_db or foreground or zero" -o addopts=` → 289 passed, 690 deselected
- `python -m pytest -q -o addopts=` → 1570 passed, 12 xfailed
- `scripts/validate_release.sh --full --keep-going --output /tmp/hermes-lcm-release-validation-aux-lifecycle-followups5-20260706052133` → PASS
- local `codex review --base origin/main -c model_reasoning_effort='"'"'xhigh'"'"'` → no discrete introduced bug found

## Risk / impact
- Risk is limited to auxiliary/stateless lifecycle and host fallback paths.
- Normal LCM storage paths should remain unchanged.
- The new regressions are intentionally defensive around stale callback ordering and reused session IDs.

PR #326, opened 2026-07-06
Drive scalar env config from a declarative field spec (WS5.6)

ai-ag2026:
## Summary
- Introduce a declarative `ENV_FIELD_SPECS` registry `(name, env_key, py_type)` in `config.py` for the scalar `LCM_*` env overrides.
- Replace the ~35 repetitive per-field assignments in `LCMConfig.from_env` with one uniform loop over the registry.
- Let `presets.py` derive its preset-field env/parse lookups from the same registry instead of two hand-maintained parallel dicts.

## Why
WS5.6 of the WS5 workstream. `from_env` was ~160 lines of
`c.X = _int/_float/_bool/_str("LCM_X", c.X)` boilerplate, and `presets.py`
kept a second copy of the field→env and field→type mapping (`_FIELD_ENV` /
`_FIELD_PARSERS`). This gives one source of truth for the scalar env overrides,
consumed by both config parsing and preset dry-run.

## Validation
- `ruff check config.py presets.py` → All checks passed
- **Equivalence harness** (byte-identical proof): snapshots `from_env()` — full
  `asdict` + `config_sources` + `config_source_warnings` — plus preset override
  discovery across **600 scenarios** (every `LCM_*` key absent / valid /
  invalid). Result: **0 differing scenarios**, preset maps identical.
- `python -m pytest -q -o addopts=` (full suite) → 1545 passed, 12 xfailed

## Notes
- Behaviour-preserving. Source-tracked fields (provenance recording and/or a
  computed default: `fresh_tail_count`, `leaf_chunk_tokens`, `context_threshold`,
  the `summary_spend_*` and `summary_timeout_ms` fields, codex autoraise) stay
  explicit and are skipped by the loop via `_SOURCE_TRACKED_ENV_FIELDS`.
  Pattern-list overrides (with their `_source` sidecar) and the inline
  `gc_max_age_hours` float parse also stay explicit.
- `presets.py` still owns *which* fields are preset-overridable (`_PRESET_FIELDS`,
  unchanged set of 5) and only sources the env var + parse type from the shared
  spec — so `explicit_operator_overrides` / `invalid_operator_overrides` keep
  scanning exactly those 5 fields, not the whole registry.
- No import cycle: `presets` imports `config`; `config` imports neither.
- Independent of #322/#323/#324/#325 (touches only `config.py` + `presets.py`).

## Refs
- Design: WS5 engine decomposition, companion workstream WS5.6 (declarative config).

### DIFF
--- a/config.py
+++ b/config.py
@@ -46,6 +46,10 @@ def _parse_bool_env(key: str, default: bool) -> bool:
     return default
 
 
+def _parse_str_env(key: str, default):
+    return os.environ.get(key, default)
+
+
 def _parse_int_env_with_source(
     key: str,
     default: int,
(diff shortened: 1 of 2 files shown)


Tosko4 (approved):
@stephenschoettler I reviewed this at `3bb954463ab65e9b252fd42e5fb6957675a2daea`. This looks valid to me.

What I checked:
- Diff is limited to `config.py` and `presets.py`; no unrelated churn or secrets.
- `ENV_FIELD_SPECS` covers the scalar env overrides and keeps the source-tracked fields explicit, so `config_sources` / warnings are still preserved for the special cases.
- Pattern-list envs and `gc_max_age_hours` stay explicit, which is the right boundary.
- `presets.py` derives env key/type from the shared registry but still only considers the existing five preset override fields, so this does not widen operator override scope.
- CI is green on the PR head.

Local validation I ran:
- `git diff --check origin/main...HEAD` → pass
- `ruff check config.py presets.py` → pass
- `python -m pytest tests/test_lcm_core.py::TestConfig tests/test_lcm_preset_command.py tests/test_lcm_command.py::test_lcm_status_json_reports_effective_config_sources -q -o addopts=` → 37 passed
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=` → 979 passed
- Custom registry/preset smoke over all 40 `ENV_FIELD_SPECS` entries and all 5 preset override fields → pass
- Import smoke for `hermes_lcm.config`, `presets`, `command`, and `tools` → pass
- `python -m pytest -q -o addopts=` → 1545 passed, 12 xfailed
- Local merge probe into current `origin/main` → clean merge; targeted config/preset tests still 37 passed
- Local Codex xhigh review → no blockers

One non-blocking cleanup note: `config.py` defines `_PRESET_ENV_FIELDS`, while `presets.py` keeps its own `_PRESET_FIELDS` tuple. That is not a behavior issue and I would not block this PR on it, but if the goal is also to centralize the preset field set itself, that can be tidied later.

Approving from the config/preset parity side. The branch is currently marked behind `main`, but the local merge probe was clean.

PR #324, opened 2026-07-06
Move backup/rotate maintenance ops into maintenance.py (WS5 seam 3)

ai-ag2026:
## Summary
- Add `maintenance.py` holding the LCM store's backup/rotate data operations.
- Move `flush_engine_connections`, `backup_database`, and `rotate_backup_database` out of `command.py` into it (renamed to public module functions).
- `command.py` imports the two entry points and keeps only the `/lcm` text formatting.

## Why
This is seam 3 of the WS5 `LCMEngine`/command decomposition: pull the
store/dag/lifecycle connection-flush and SQLite snapshot logic into one module,
separate from `/lcm` command presentation. `command.py` shrinks by ~113 lines
and the maintenance primitives (`/lcm backup`, `/lcm rotate`, and the
backup-first path shared by doctor/clean/repair/source-apply) now live behind a
small, testable module API instead of inline command helpers.

## Validation
- `ruff check maintenance.py command.py` → All checks passed
- `python -m pytest tests/test_lcm_rotate.py tests/test_lcm_command.py -q -o addopts=` → 95 passed
- `python -m pytest -q -o addopts=` (full suite) → 1545 passed, 12 xfailed (identical to base)
- `scripts/validate_release.sh --full --keep-going --output <dir>` → PASS

## Notes
- Behaviour-preserving: the three functions move unchanged — same signatures,
  same return dicts, same error handling. `flush_engine_connections` is used
  only by the two backup writers, so it stays internal to `maintenance.py` and
  is not imported by `command.py`. The text-formatting helpers (`_backup_text`,
  `_rotate_text`, `_rotate_apply_text`) and the doctor/clean/repair/source-apply
  callers stay in `command.py` and call the imported functions.
- No test references these helpers by name — only via `handle_lcm_command` — so
  no test imports changed.
- The engine-side `backup_dir` / `rotate_backup_path` / `rotate_active_session`
  methods are intentionally left on `LCMEngine` for this PR (they are called
  widely across the code and tests); `maintenance.py` calls them via the passed
  engine. They can move in a later seam if desired.

## Stacking
Stacked on #323 (WS5 store-connection encapsulation) — this branch builds on it,
so the moved functions carry #323's `store.commit()` / `store.backup()` calls.
Please review/merge #323 first; GitHub will then narrow this PR's diff to just
the `maintenance.py` extraction.

## Refs
- Design: WS5 engine decomposition, seam 3 (Backup/Rotate/Maintenance).
[...]

PR #327, opened 2026-07-06
Lift /lcm doctor session-stat scans into MessageStore (WS5.5)

ai-ag2026:
## Summary
- Add `MessageStore.scan_session_cleanup_stats()` and `scan_session_retention_stats(session_id)`.
- Move the two `/lcm doctor` session-stat SQL queries out of `command.py` into those methods.

## Why
First slice of WS5.5 (layer discipline). The `/lcm doctor` clean and retention
scanners ran their session-stat SQL directly against the store connection from
the command layer. This lifts those raw queries behind named store methods, so
the store owns its SQL and connection and the command layer keeps only the
pattern-match / staleness / protection policy.

## Validation
- `ruff check store.py command.py` → All checks passed
- `python -m pytest tests/test_lcm_command.py -q -o addopts= -k "clean or retention or doctor or scan"` → 52 passed
- `python -m pytest -q -o addopts=` (full suite) → 1545 passed, 12 xfailed (identical to base)

## Notes
- Behaviour-preserving: the SQL and its parameters are unchanged, and callers'
  error handling is unchanged — each method raises and the caller's existing
  `except` still catches it and returns the same error dict.
- Scope: just these two doctor scanners (a coherent unit). The broader WS5.5
  sweep (other raw command/tools queries → store/dag methods, a read-only engine
  facade for the `engine._*` accesses) is intentionally left for follow-ups so
  each PR stays small and reviewable.

## Stacking
Stacked on #323 (store-connection encapsulation) — these scanners previously
reached `engine._store.connection`, which #323 introduced. Please review/merge
#323 first; GitHub will then narrow this PR's diff to just the query lift.
Independent of #324 (touches different command.py functions).

## Refs
- Design: WS5 companion workstream WS5.5 (layer discipline).

### DIFF
--- a/command.py
+++ b/command.py
@@ -221,37 +221,8 @@ def _status_text(engine) -> str:
 
 
 def _scan_clean_candidates(engine) -> dict[str, Any]:
-    conn = engine._store.connection
     try:
-        rows = conn.execute(
-            """
-            WITH session_ids AS (
-                SELECT session_id FROM messages
-                UNION
-                SELECT session_id FROM summary_nodes
-            ),
-            message_stats AS (
-                SELECT session_id,
-                       COUNT(*) AS message_count,
-                       COALESCE(SUM(token_estimate), 0) AS token_total
-                FROM messages
-                GROUP BY session_id
-            ),
-            node_stats AS (
-                SELECT session_id, COUNT(*) AS node_count
-                FROM summary_nodes
-                GROUP BY session_id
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):
Maintainer gate checked: live head/CI/review threads clean, local merge-train validation passed. Approved for this bounded queue landing.

PR #328, opened 2026-07-06
Extract Codex OAuth route detection into codex_routing.py (WS5 seam)

ai-ag2026:
## Summary
- Move the four Codex-route helpers and their context-cap table out of `engine.py` into a new `codex_routing.py`: `_bare_model_slug`, `_is_openai_codex_route`, `_codex_oauth_context_cap`, `_is_codex_gpt55_route`, and `_CODEX_OAUTH_CONTEXT_CAPS`.
- `engine.py` imports the two helpers it actually calls (`_codex_oauth_context_cap`, `_is_codex_gpt55_route`); the two internal helpers and the caps table travel with them.
- Keep `_CODEX_GPT55_COMPACTION_THRESHOLD` in `engine.py` — it is engine-side compaction policy, not route detection.

## Why
WS5 engine decomposition. `engine.py` is ~9k lines; the Codex OAuth route logic (which model slugs are on the ChatGPT Codex OAuth route and what effective context window that route enforces) is a cohesive, pure, self-contained concern with no engine state. Isolating it into its own module — alongside the existing `model_routing.py` split — makes the route policy discoverable in one place and shrinks the monolith, with zero behaviour change.

## Validation
- `ruff check codex_routing.py engine.py` → All checks passed
- **Equivalence harness** (byte-identical proof): reconstructs the four helpers + caps dict from `git show main:engine.py` and compares against `codex_routing.py` across an exhaustive `(model, provider)` grid — 11 provider variants × 134 model variants (bare/prefixed slugs, whitespace, case, near-miss versions like `gpt-5.55`/`gpt-5.5.1`). Result: **1474 scenarios, 0 differing**, caps dict identical.

## Notes
- Behaviour-preserving; the moved code is byte-identical to its pre-move form.
- Import surface unchanged for callers: `engine.py` keeps calling `_codex_oauth_context_cap` / `_is_codex_gpt55_route` at the same call sites (`_effective_context_length`, `_runtime_context_threshold`). The two remaining helpers (`_bare_model_slug`, `_is_openai_codex_route`) are only used internally by the moved functions, so they are not re-imported into `engine.py` (no unused import).
- No import cycle: `codex_routing` imports nothing from the package; `engine` imports `codex_routing`.
- Existing Codex-cap tests in `tests/test_lcm_engine.py` continue to exercise the moved code through the engine methods.
- Independent of the other WS5 seams (touches only `engine.py` + new `codex_routing.py`).

## Refs
- Design: WS5 engine decomposition, companion to `model_routing.py` / `sanitize.py` / `maintenance.py` seams (#323–#327).

### DIFF
--- a/codex_routing.py
+++ b/codex_routing.py
@@ -1,0 +1,71 @@
+"""Codex OAuth route detection and effective context-window caps.
+
+Isolated from ``engine.py`` (WS5 seam) so the Codex-specific routing policy —
+which model slugs are on the ChatGPT Codex OAuth route and what effective
+context window that route enforces — lives in one cohesive place. These are
+pure helpers with no engine state; ``engine.py`` imports them and keeps its own
+policy constants (for example the gpt-5.5 compaction threshold).
+"""
+
+from __future__ import annotations
+
+# ChatGPT Codex OAuth exposes provider-enforced context windows that can be
+# materially lower than the same model slug on direct OpenAI/OpenRouter routes.
+# Hermes Agent resolves these from chatgpt.com/backend-api/codex/models, with
+# this table as its fallback. LCM sees only the host-advertised context_length;
+# when that value was explicitly overridden above the real Codex OAuth window,
+# we still have to budget against the effective provider window or compaction
+# fires too late and provider requests can overflow.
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):

PR #329, opened 2026-07-06
Extract SQLite lock-contention helpers into sqlite_util.py (WS5 seam)

ai-ag2026:
## Summary
- Move the SQLite busy/lock utilities out of `engine.py` into a new `sqlite_util.py`: `_is_sqlite_locked_error` (exception-chain lock detection), `_sqlite_busy_timeout_ms` (PRAGMA probe), and the `_temporary_sqlite_busy_timeout` context manager.
- `engine.py` imports the two entry points it calls (`_is_sqlite_locked_error`, `_temporary_sqlite_busy_timeout`); the busy-timeout probe is internal and travels with them.
- Drop the now-unused `contextlib.contextmanager` and `typing.Iterator` imports from `engine.py`.

## Why
WS5 engine decomposition. `engine.py` is ~9k lines; detecting SQLite lock-contention error chains and temporarily bounding `busy_timeout` on gateway-critical paths are a cohesive, pure SQLite concern with no engine state. Isolating them into their own module shrinks the monolith with zero behaviour change, alongside the earlier `sanitize.py` / `maintenance.py` / `codex_routing.py` splits.

## Validation
- `ruff check sqlite_util.py engine.py` → All checks passed
- **Equivalence harness**: reconstructs the three helpers from `git show main:engine.py` (AST slice) and proves (1) the moved source is **byte-identical** (3/3 functions), and (2) behavioural parity old-vs-new across 38 scenarios — exception chains (nested `__cause__`/`__context__`, non-sqlite, non-locked, `IntegrityError` carrying "locked", a cyclic `__context__` loop) and a `busy_timeout` base×timeout grid (set/restore across multiple connections plus `None`). Result: **0 differing**.
- `python -m pytest -q -o addopts=` (full suite) → **1570 passed, 12 xfailed**
- `scripts/validate_release.sh --full --keep-going --output <dir>` → **PASS**

## Notes
- Behaviour-preserving; the moved code is byte-identical to its pre-move form.
- `_SESSION_END_BUSY_TIMEOUT_MS` stays in `engine.py` — it is the caller-side timeout budget passed into `_temporary_sqlite_busy_timeout`, not a SQLite primitive.
- Call sites unchanged: `_is_sqlite_locked_error` (3 sites in the SQLite retry paths) and `_temporary_sqlite_busy_timeout` (session-end path) call the same names. `_sqlite_busy_timeout_ms` is only used internally by the context manager, so it is not re-imported into `engine.py` (no unused import).
[...]

PR #330, opened 2026-07-06
Extract plugin/git runtime-identity helpers into runtime_identity.py (WS5 seam)

ai-ag2026:
## Summary
- Move the plugin- and git-identity provenance helpers out of `engine.py` into a new `runtime_identity.py`: `_strip_metadata_scalar`, `_plugin_metadata` (with its `_PLUGIN_METADATA` manifest cache and `_PLUGIN_ROOT`), and `_git_runtime_identity`.
- `engine.py` imports the three symbols it references (`_plugin_metadata`, `_git_runtime_identity`, `_PLUGIN_ROOT`); `_strip_metadata_scalar` and the cache stay internal to the module.
- The move drops `engine.py`'s now-unused `subprocess` import.

## Why
WS5 engine decomposition. `engine.py` is ~9k lines; resolving the plugin's own name/version from its manifest and probing best-effort git identity for source checkouts are a cohesive provenance concern that feeds the `plugin_*` fields of the runtime status/doctor payload. Isolating them into their own module shrinks the monolith with no behaviour change, alongside the earlier `sanitize.py` / `maintenance.py` / `codex_routing.py` / `sqlite_util.py` splits.

## Validation
- `ruff check runtime_identity.py engine.py tests/...` → All checks passed
- **Equivalence harness**: reconstructs the three helpers from `git show main:engine.py` (AST slice) and proves (1) the moved source is **byte-identical** (3/3 functions), and (2) behavioural parity old-vs-new across 38 scenarios — `_strip_metadata_scalar` string grid; `_git_runtime_identity` with a faked `subprocess.run` (no `.git`, `OSError`, `SubprocessError`, non-zero return, status-probe-only failure, clean/dirty trees); `_plugin_metadata` over 9 manifest contents including a missing manifest. Result: **0 differing**.

## Notes
- Behaviour-preserving; the moved functions are byte-identical to their pre-move form. The one intentional non-identity is the module-local `logger` name (`hermes_lcm.engine` → `hermes_lcm.runtime_identity`) on the two best-effort `logger.debug` provenance lines — consistent with the existing `model_routing.py` split.
- `_PLUGIN_ROOT` stays value-identical: both `engine.py` and `runtime_identity.py` live in the plugin root, so `Path(__file__).resolve().parent` resolves to the same directory.
- **Test retargeting** (not a pure move — the manifest cache is genuinely module-scoped): four tests that reached these symbols through the engine module now target `runtime_identity`, with assertions unchanged:
  - `tests/test_lcm_engine.py`: the two `test_plugin_metadata_*` tests patch/read `runtime_identity._PLUGIN_METADATA` (the `global` binds to the module that owns it; patching `engine` would be a no-op and the missing-manifest test would then observe a stale cached value instead of the `"unknown"` default).
  - `tests/test_packaging_install.py`: the two `test_git_runtime_identity_*` tests patch `runtime_identity.subprocess` and resolve the submodule via `sys.modules[f"{module_name}.runtime_identity"]`, since `engine` no longer imports `subprocess`.
- No import cycle: `runtime_identity` imports only the stdlib; `engine` imports `runtime_identity`.
- Independent of the other WS5 seams (touches only `engine.py`, the new module, and the four retargeted tests).

## Refs
- Design: WS5 engine decomposition, companion to the `sanitize.py` / `maintenance.py` / `codex_routing.py` / `sqlite_util.py` seams (#323–#329).


stephenschoettler (approved):
Approved after maintainer conflict refresh onto current main. Local focused tests and GitHub CI are green; conflict resolution was mechanical (kept current registry/sqlite seams, moved runtime identity helpers to the new module).

PR #331, opened 2026-07-06
Extract message tool-call/noise analysis into message_analysis.py (WS5 seam)

ai-ag2026:
## Summary
- Move the pure message-list inspection helpers out of `engine.py` into a new `message_analysis.py`: `_tool_call_id`, `_assistant_tool_call_ids`, `_matched_tool_call_ids`, and `_is_synthetic_assistant_noise` (with its `_SYNTHETIC_ASSISTANT_NOISE` vocabulary).
- `engine.py` imports the four it calls; the synthetic-noise vocabulary stays internal to the module.

## Why
WS5 engine decomposition. `engine.py` is ~9k lines; extracting/pairing assistant↔tool call ids across a message list and detecting synthetic "noise" turns (acks/heartbeats) are stateless inspection helpers with no engine state. Isolating them shrinks the monolith with no behaviour change, alongside the earlier `sanitize.py` / `maintenance.py` / `codex_routing.py` / `sqlite_util.py` / `runtime_identity.py` splits.

## Validation
- `ruff check message_analysis.py engine.py` → All checks passed
- **Equivalence harness**: reconstructs the four helpers from `git show main:engine.py` (AST slice) and proves (1) the moved source is **byte-identical** (4/4 functions, plus an identical `_SYNTHETIC_ASSISTANT_NOISE` set), and (2) behavioural parity old-vs-new across 60 scenarios — `_tool_call_id` over non-dict / missing-id / falsy-id / whitespace inputs; `_assistant_tool_call_ids` and `_matched_tool_call_ids` over 12 message-list shapes (orphan tool results, unmatched assistant calls, whitespace ids, duplicates, `tool_calls=None`); `_is_synthetic_assistant_noise` over 31 content strings (case, wrapping brackets/backticks/markdown, near-misses like "acking"). Result: **0 differing**.
- `python -m pytest -q -o addopts=` (full suite) → **1570 passed, 12 xfailed**
- `scripts/validate_release.sh --full --keep-going --output <dir>` → **PASS**

## Notes
- Behaviour-preserving; the moved code is byte-identical to its pre-move form.
- Two engine methods unpack a tuple element into a local also named `_tool_call_id` (`role, content, _tool_call_id, tool_calls = identity`). That is a function-local rebinding that shadowed the module-level function before this change and still does after it — unchanged behaviour.
- No import cycle: `message_analysis` imports only `re` + `typing`; `engine` imports `message_analysis`.
[...]

PR #332, opened 2026-07-06
Extract active-engine registry into engine_registry.py (WS5 seam)

ai-ag2026:
## Summary
- Move the process-wide active-LCM-clone registry out of `engine.py` into a new `engine_registry.py`: the registry `RLock`, the two `WeakValueDictionary` registries (by session id / conversation id), the resolver/matcher helpers, the removal helper, and the public `resolve_active_lcm_engine`.
- `engine.py` imports the shared lock, the two registries, `_remove_registry_entries_for_engine`, and `resolve_active_lcm_engine`. The matcher helpers (`_is_usable_lcm_engine`, `_engine_matches_session_binding`, `_engine_matches_conversation_binding`) stay internal to the module.

## Why
WS5 engine decomposition. `engine.py` is ~9k lines; the process-wide registry that lets post-turn ingest follow the active per-agent LCM clone (instead of the plugin singleton) is a cohesive concern with its own lock and weak maps. Isolating it shrinks the monolith with no behaviour change, alongside the earlier `sanitize.py` / `maintenance.py` / `codex_routing.py` / `sqlite_util.py` / `runtime_identity.py` / `message_analysis.py` splits.

## Validation
- `ruff check engine_registry.py engine.py` → All checks passed
- **Equivalence harness**: reconstructs the five helpers from `git show main:engine.py` (AST slice) and proves (1) the moved source is **byte-identical** (5/5 functions), and (2) behavioural parity old-vs-new by driving two independent registries through an 18-step register/resolve/remove sequence (session/conversation collisions, non-usable engines, missing-ingest engines, session-id reuse across engines) and comparing both the resolved engine and the full registry key→engine structure at every step. Result: **0 differing**.

## Notes
- Behaviour-preserving; the moved code is byte-identical to its pre-move form.
- The `LCMEngine` binding methods (`_register_active_engine_binding` / `_unregister_active_engine_binding`) keep their bodies verbatim: the lock and the two registries are shared mutable objects, so mutating them through the imported names (`_ACTIVE_ENGINES_BY_SESSION_ID[sid] = self`, `with _ACTIVE_ENGINE_REGISTRY_LOCK:`) hits the same objects the resolver reads.
- `resolve_active_lcm_engine` is re-exported from `engine.py` (`# noqa: F401`) so any host importing it from `hermes_lcm.engine` keeps working; it is also importable from `hermes_lcm.engine_registry`.
- No import cycle: `engine_registry` imports only the stdlib; `engine` imports `engine_registry`.
- Independent of the other WS5 seams (touches only `engine.py` + the new module).

## Refs
- Design: WS5 engine decomposition, companion to the earlier seams (#323–#331).

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -26,6 +26,13 @@ from agent.context_engine import ContextEngine
 from .config import LCMConfig
 from .dag import SummaryDAG, SummaryNode
 from .diagnostics import _enforce_state_db_containment
+from .engine_registry import (
+    _ACTIVE_ENGINE_REGISTRY_LOCK,
+    _ACTIVE_ENGINES_BY_CONVERSATION_ID,
+    _ACTIVE_ENGINES_BY_SESSION_ID,
+    _remove_registry_entries_for_engine,
+    resolve_active_lcm_engine,  # noqa: F401  (re-exported: hosts import it from .engine)
+)
 from .escalation import (
     SummaryCircuitBreaker,
     SummarySpendGuard,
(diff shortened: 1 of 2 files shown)


stephenschoettler (approved):

PR #333, opened 2026-07-06
Extract placeholder-ledger bookkeeping into a mixin (WS5 Seam 1)

ai-ag2026:
## Summary
- Move the 35 ignored-active-replay placeholder / dependent-reply bookkeeping methods out of `LCMEngine` into a `PlaceholderLedgerMixin` in the new `placeholder_ledger.py`, and mix it in: `class LCMEngine(PlaceholderLedgerMixin, ContextEngine)`.
- The methods are relocated **verbatim** and keep running bound to the engine instance, so nothing else changes: no call sites, no tests.

## Why
WS5 engine decomposition, Seam 1. `engine.py` was ~9k lines. The placeholder ledger — session-scoped metadata-key builders, generated-placeholder hash/count/ordinal load+persist, active-replay digest budgets, dependent-reply fingerprints/records, and `_apply_ignored_active_replay_placeholders` — is a large, cohesive concern (~660 lines). A mixin lifts it into its own file while preserving `self` semantics: because `self` is still the `LCMEngine`, the methods read/write the engine's shared runtime state and call engine helpers through normal MRO attribute lookup.

## Validation
- `ruff check placeholder_ledger.py engine.py` → All checks passed
- **Byte-identity proof**: AST-slices all 35 methods from `git show main:engine.py` (in `LCMEngine`) and from the new `placeholder_ledger.py` (in `PlaceholderLedgerMixin`) and asserts the source text is identical → **35/35 byte-identical**, and **35/35 removed** from `engine.py`'s `LCMEngine`.
- Focused placeholder/ignored/dependent/digest/ordinal/hash tests → **105 passed**
- `python -m pytest -q -o addopts=` (full suite) → **1570 passed, 12 xfailed** (the ~130 white-box tests in `tests/test_lcm_engine.py` that call these private methods directly are unchanged and green)
- `engine.py`: 9048 → 8391 lines (−657)

## Notes
- Behaviour-preserving; method bodies are byte-identical to their pre-move form (only the module-local `logger` name changes, as in the earlier `runtime_identity.py` split).
- **State ownership is unchanged.** The two shared placeholder-id sets (`_generated_ignored_active_replay_placeholder_hashes` / `_message_ids`) and the per-turn `_current_compress_store_ids_by_message_id` cache stay owned/initialised by `LCMEngine` — they are also touched by sanitize/reconcile methods that remain on the engine — so the mixin is a pure relocation, not a state-ownership change. It accesses them, and callbacks (`_get_store_id_map_for_messages`, `_matches_ignore_message_patterns`, `_stable_tool_calls_identity`, `_message_replay_identity`, `_copy_active_replay_messages_preserving_generated_ids`), via `self`.
- No import cycle: `placeholder_ledger` imports only stdlib + `.message_content`; `engine` imports `placeholder_ledger`.
- MRO: `LCMEngine(ContextEngine, PlaceholderLedgerMixin)` — the moved names live only on the mixin (removed from `LCMEngine`), so they resolve there; no collisions with `ContextEngine`.

## Refs
- Design: WS5 engine decomposition, Seam 1. Companion to the earlier module seams (#323–#332).


stephenschoettler (approved):

PR #334, opened 2026-07-06
Extract the leaf-compaction pipeline into a mixin (WS5 Seam 6)

ai-ag2026:
## Summary
- Move the compaction gate + leaf pipeline out of `LCMEngine` into a `CompactionMixin` in the new `compaction.py`, and mix it in ahead of `ContextEngine`: `class LCMEngine(CompactionMixin, ContextEngine)`.
- Moved verbatim: `should_compress`, `should_compress_preflight` (public), `_replay_diff_requests_ingest_cleanup`, `_has_ignored_backlog_outside_fresh_tail`, `_leaf_compaction_candidate_status`, `_working_leaf_chunk_tokens`, `_select_oldest_leaf_chunk`, and the main `compress` entry point.

## Why
WS5 engine decomposition, Seam 6. The compaction gate + leaf-selection pipeline is a large, cohesive concern (~750 lines). A mixin lifts it into its own file while preserving `self` semantics: `self` is still the `LCMEngine`, so `_ingest_cursor`, `_store`, `_dag`, `_lifecycle`, the per-turn caches and the ~60 sibling helpers stay on the engine and resolve via normal MRO attribute lookup.

## Validation
- `ruff check compaction.py engine.py` → All checks passed
- **Byte-identity proof**: AST-slices all 8 methods from `git show main:engine.py` (in `LCMEngine`) and from `compaction.py` (in `CompactionMixin`) → **8/8 byte-identical**, and **8/8 removed** from `engine.py`.
- Focused compaction tests (compress/leaf/compact/preflight/rescue/chunk) → **153 passed**
- `engine.py`: 9048 → 8296 lines (−752)

## Notes
- Behaviour-preserving; method bodies are byte-identical to their pre-move form (only the module-local `logger` name changes).
- **MRO:** the mixin is placed BEFORE `ContextEngine` so its `compress` / `should_compress` / `should_compress_preflight` override the `ContextEngine` protocol defaults (which also define those names). The private helpers have no `ContextEngine` counterpart.
- **Summarize step stays on the engine.** `_summarize_leaf_chunk_with_rescue` (and its `_next_leaf_rescue_chunk` / `_is_retry_worthy_leaf_summary_error` helpers) is deliberately left on `LCMEngine`, because a large number of tests patch the module-level `summarize_with_escalation` on `hermes_lcm.engine` to intercept the LLM call. `compress()` reaches it via `self` as a callback-out, so behaviour is unchanged and no test is touched.
- Drops `engine.py`'s now-exclusive `_contains_sensitive_redaction` import (moved to `compaction.py`).
- No import cycle: `compaction` imports stdlib + `.dag` / `.message_content` / `.sanitize` / `.tokens`; `engine` imports `compaction`.

## Refs
- Design: WS5 engine decomposition, Seam 6. Companion to Seam 1 (#333) and the module seams (#328–#332).


stephenschoettler (approved):
Approved after maintainer conflict refresh onto current main. Local focused tests/ruff/diff-checks and GitHub CI are green; no active review-thread debt. Conflict resolution was mechanical (kept current WS5 helper seams/reconcile mixin and added CompactionMixin to the current engine mixin stack).

PR #335, opened 2026-07-06
Extract ingest-cursor reconciliation into a mixin (WS5 Seam 4)

ai-ag2026:
## Summary
- Move the ingest-cursor reconciliation machinery and the stable replay-identity primitives it relies on out of `LCMEngine` into a `ReconcileMixin` in the new `reconcile.py`, and mix it in: `class LCMEngine(ReconcileMixin, ContextEngine)`.
- 20 methods: tool-call/message replay-identity primitives, persisted-output durable-replay evidence helpers, and the reconciliation orchestration (`_find_reconciled_cursor_for_store_tail`, `_reconcile_ingest_cursor_from_store`, `_get_store_id_map_for_messages`, …).

## Why
WS5 engine decomposition, Seam 4. The post-restart reconciliation of the persisted store tail against the active message list — plus the replay-identity primitives that make it deterministic — is the single largest cohesive concern left in `engine.py` (~1.1k lines). A mixin lifts it into its own file while preserving `self` semantics: `_store`, `_session_id`, `_config` and the safety-critical `_ingest_cursor` stay on the engine and resolve via `self` (`_ingest_cursor` is still written by `_ingest_messages` from the reconciler's return value).

## Validation
- `ruff check reconcile.py engine.py` → All checks passed
- **Byte-identity proof**: AST-slices all 20 methods from `git show main:engine.py` vs `reconcile.py` → **18/20 byte-identical**; the two differing methods differ *only* by the deliberate class-qualifier rename below (verified: reversing `ReconcileMixin.` → `LCMEngine.` makes them identical). **20/20 removed** from `engine.py`.
- Focused reconcile/replay-identity/store-id-map/persisted-output tests → **63 passed**
- `engine.py`: 9048 → 7943 lines (−1105)

## Notes
- Behaviour-preserving. Method bodies are byte-identical except two mechanical, behaviour-preserving adjustments needed to break an `engine` ↔ `reconcile` import cycle:
  - The two tool-call-identity **staticmethods** call the canonicalizer via the mixin class (`ReconcileMixin._canonicalize_tool_call_identity_value`) instead of `LCMEngine`. `LCMEngine` inherits the staticmethod from the mixin, so resolution is identical.
  - `_PRESERVED_OBJECTIVE_CONTEXT_PREFIX` moves to `reconcile.py` (used by the reconciliation scan) and is **re-imported into `engine.py`** for its other readers, so no code that reads it changes.
- Drops six now-exclusive `ingest_protection`/`externalize` imports from `engine.py` (moved to `reconcile.py`).
- No call sites and no tests change; the white-box tests that bind `_message_replay_identity` / `_get_store_id_map_for_messages` call them on an `LCMEngine` instance and resolve via the mixin.
- The reconcile methods are all private, so mixing after `ContextEngine` is fine (no protocol-name collision).

## Refs
- Design: WS5 engine decomposition, Seam 4. Companion to Seam 1 (#333), Seam 6 (#334), and the module seams (#328–#332).


stephenschoettler (approved):
Approved after maintainer conflict refresh onto current main. Local focused tests/ruff/diff-checks and GitHub CI are green; no active review-thread debt. Conflict resolution was mechanical (kept current WS5 helper seams and added ReconcileMixin to the current engine mixin stack).

PR #338, opened 2026-07-06
Prefer the host's explicit subagent signal over the call-stack frame walk (WS5.7)

ai-ag2026:
## What this changes
LCM now identifies a subagent/auxiliary child session — and resolves its parent
session — from the **host's own explicit signal** when the host provides it,
instead of walking the call stack and reading private agent attributes. The old
mechanism stays as a fallback, so nothing regresses on hosts that don't emit the
signal.

## Why
When `on_session_start` fires for a session, LCM needs to know *is this a
subagent, and who is its parent?* to attribute lineage correctly (session
ownership, ignore/stateless handling). The `ContextEngine.on_session_start`
callback carries **no** subagent/parent argument, so until now LCM's only option
was `_in_process_parent_session_id` walking up ~32 `inspect.currentframe()`
frames and sniffing private attributes on the calling agent object
(`_subagent_id`, `_parent_subagent_id`, `_delegate_depth`, `log_prefix`,
`enabled_toolsets`, `_memory_write_context`, …).

That heuristic works but is fragile:
- it reads **undocumented host internals** — a host rename silently breaks
  detection (no error, just wrong lineage);
- it depends on stack shape (a 32-frame budget), so refactors on either side can
  move the target frame out of range.

The host already publishes the same information **explicitly and publicly**: the
plugin hook bus fires `subagent_start` / `subagent_stop` with
`child_session_id`, `parent_session_id`, `child_subagent_id`,
`parent_subagent_id`, and `child_role` when a delegate/subagent is spawned. (The
shipped `observability/nemo_relay` plugin already consumes these.) Using that
signal is both more robust and keeps LCM off the host's private surface.

## How
- **`aux_session.py`** — a small, process-wide, lock-guarded, memory-bounded
  registry `_SUBAGENT_LINEAGE_BY_SESSION_ID` keyed by child session id, with
  `record_subagent_start` / `record_subagent_stop` / `explicit_subagent_lineage`
  helpers. `_in_process_parent_session_id` consults this explicit registry
  **before** the frame walk.
- **`__init__.py`** — `register()` subscribes to `subagent_start` /
  `subagent_stop` when the host exposes `register_hook` (guarded with
  `getattr`/`callable`, like the existing `register_tool` gating). The
  `subagent_start` event fires before the child's `on_session_start`, so the
  linkage is present when LCM needs it.

## Precedence (unchanged where the signal is absent)
1. an explicit `parent_session_id` kwarg (highest);
2. **new:** the recorded `subagent_start` lineage for this session id;
3. the legacy `inspect.currentframe()` frame walk (fallback).

On a host with no plugin hook bus, step 2 is simply empty and behaviour is
byte-for-byte the old path.

## Validation
- `ruff check` → clean.
- New `tests/test_subagent_lineage.py` (5 tests): registry record/query/stop
  roundtrip, missing-child-id ignored, **explicit parent takes precedence over
  the frame walk** (and reverts to the frame walk after `subagent_stop`), hook
  subscription via `register()`, and the no-hook-bus no-op fallback.
- `python -m pytest -q -o addopts=` (full suite) → **1575 passed, 12 xfailed**
  (1570 prior + 5 new; the existing frame-walk tests are unchanged and green,
  confirming the fallback is untouched).
- `scripts/validate_release.sh --full --keep-going` → **PASS**.

## Notes
- Additive and backward-compatible: no host change is required. (A future
  one-line host addition — passing `parent_session_id=` into the
  `on_session_start` call — could let LCM drop the frame walk entirely, but this
  PR does not depend on it.)
- Builds on the WS5 Seam 5 extraction (`aux_session.py`, #337).

## Refs
- WS5.7 (frame-walk isolation) from the WS5 engine-decomposition plan.


stephenschoettler (approved):
Reviewed via the maintainer gate; CI is green and the reviewer lane passed on head 24238bda55d923f8400f6f4ad392855fd13e426c.

PR #339, opened 2026-07-06
Encapsulate DAG/lifecycle connections behind a public accessor (WS5.5)

ai-ag2026:
## What this changes
`command.py` and `tools.py` reached directly into the **private** SQLite
connections of the summary DAG (`engine._dag._conn`) and the lifecycle store
(`engine._lifecycle._conn`) to run read-only doctor/inspect diagnostics — FTS
sync counts, integrity checks, latest-node lookups, and the maintenance-debt
scan. This adds a public read-only `connection` accessor to both stores and
routes those reach-ins through it.

## Why
Reaching into a helper's `_conn` from another module bypasses the store's
encapsulation and its write-lock contract, and is exactly the kind of layer
violation the `MessageStore.connection` accessor (#323) was introduced to fix for
the message store. The DAG and lifecycle stores were still accessed raw. Making
the read-only diagnostic access explicit keeps the SQLite ownership in one place
and lets those stores evolve their internals without breaking `command.py` /
`tools.py`.

## How
- **`dag.py`** — add a `SummaryDAG.connection` read-only property (live
  connection or `None`), documented as read-only; writes still go through the
  DAG's own methods under `_db_lock`.
- **`lifecycle_state.py`** — add the same `LifecycleStateStore.connection`
  property.
- **`command.py` / `tools.py`** — route the seven cross-module reach-ins
  (`engine._dag._conn` ×5, `engine._lifecycle._conn` ×2, including the
  `getattr(..., "_conn", None)` doctor form) through the new accessors.

## Validation
- `ruff check dag.py lifecycle_state.py command.py tools.py` → clean.
- `python -m pytest -q -o addopts=` (full suite) → **1570 passed, 12 xfailed**;
  the 199 doctor/inspect/fts/integrity/lifecycle tests exercise the rerouted
  accessors.
- `scripts/validate_release.sh --full --keep-going` → **PASS**.

## Notes
- Behaviour-preserving: each accessor returns the same connection object the
  reach-in previously read (and tolerates `None`). Writes are unchanged.
- Completes the WS5.5 "layer discipline" cleanup begun in #323 (store connection)
  and #327 (doctor session-stat scans) for the remaining two SQLite-backed
  helpers.

## Refs
- WS5.5 (layer discipline) from the WS5 plan; companion to #323, #327.


stephenschoettler (approved):
Maintainer review: narrow accessor refactor looks behavior-preserving. CI is green; focused local validation passed on the merge result.

PR #346, opened 2026-07-07
Make schema-column migrations idempotent under concurrent startup

ai-ag2026:
## Summary
- Add `add_column_if_missing(conn, existing_columns, column, alter_sql)` in `db_bootstrap.py` and route all column migrations through it. It keeps the existing "skip if already present" check and additionally swallows exactly the `duplicate column name` `OperationalError` (any other error still propagates), making each `ALTER TABLE ... ADD COLUMN` idempotent under concurrency.
- Converted call sites: `ensure_lifecycle_state_columns` / `ensure_message_origin_columns` (db_bootstrap.py), `MessageStore._ensure_source_column` / `_ensure_conversation_id_column` (store.py), `SummaryDAG._ensure_source_window_columns` (dag.py).

## Why
In the multi-agent deployment (gateway + CLI sessions + sub-agents) every process opens its own connection to the same `lcm.db` and runs the startup migrations concurrently. Each column migration did check-`PRAGMA table_info`-then-`ALTER` in autocommit with no guard, so two processes could both observe a column as absent and both issue the ALTER. The loser raised `sqlite3.OperationalError: duplicate column name`, which propagated out of `_init_db` and **crashed store construction**. This bites precisely at an upgrade boundary, when many processes restart together.

Behaviour is unchanged on the non-racing path (same check, same ALTERs); only the concurrent loser now skips instead of crashing.

## Validation
- New regression test `TestConcurrentStartupMigration` (tests/test_crash_safe_wal.py) races `ensure_message_origin_columns` from 8 threads, each with its own connection to one seeded pre-v5 DB (FTS-free, to isolate the column DDL). Without the guard 7/8 threads raise `duplicate column name: conversation_id`; with it every thread migrates and the column exists exactly once.
- `ruff check .` -> clean.
- `PYTHONPATH=<hermes-agent> python -m pytest -q -o addopts=` -> `1576 passed, 12 xfailed`.
- `scripts/validate_release.sh --full` -> `PASS: release validation full` (all 12 gates, incl. low-fd).

## Notes
- Out of scope / follow-up: the FTS structural rebuild (`repair_external_content_fts`) has a separate concurrency window on a different trigger (missing shadow tables / corruption, not column upgrades — a column migration does not change row counts, so it does not enter that path). Left for a focused follow-up so this PR stays scoped to the verified crash.
- Rollback: revert the commit; migrations return to the prior unguarded check-then-ALTER.

## Refs
Surfaced by a comparison against the more mature lossless-claw, whose `runLcmMigrations` serializes migrations under `BEGIN EXCLUSIVE`.

### DIFF
--- a/dag.py
+++ b/dag.py
@@ -22,6 +22,7 @@ from typing import Any, Dict, List, Optional
 
 from .db_bootstrap import (
     ExternalContentFtsSpec,
+    add_column_if_missing,
     configure_connection,
     ensure_external_content_fts,
     refuse_schema_version_too_new,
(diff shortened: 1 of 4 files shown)


stephenschoettler (approved):
Approved via Hermes Kanban reviewer gate t_d925191a. Evidence: live CI green 6/6, no unresolved review threads, and local validation passed on the reviewed head.

PR #347, opened 2026-07-07
Discard reasoning-only summaries instead of persisting them (unclosed <think>)

ai-ag2026:
## Summary
- Add `_sanitize_reasoning_summary(text)` in `escalation.py` and route `_call_llm_for_summary`'s return through it, so a reasoning-only summarizer response is discarded (returns `""`) instead of being persisted as the summary.
- Add `_REASONING_START_RE`, which matches an *unclosed* leading reasoning marker — including non-tag shapes: `<|think|>` / `<|start_of_think|>`, `[think]`, and prose headers `Thinking Process:` / `Chain of thought:`.

## Why
`_strip_reasoning_blocks` only removes *closed* `<think>...</think>` pairs (backreferenced regex). A reasoning model (Qwen/QwQ/DeepSeek-R1, GLM, MiniMax) that runs into `max_tokens` before emitting the closing tag leaves an **unclosed** block the regex cannot match. The leftover raw reasoning — which, per the existing module comment, often quotes the summarizer system prompt verbatim — was then accepted as the summary purely because `count_tokens(result) < source_tokens`, persisted as a `SummaryNode`, and replayed into the model as context (and re-fed by `lcm_expand_query`). That is both a summary-quality bug and a system-prompt leak.

With the fix, a reasoning-only response is falsy, so the existing escalation path advances to the next model / L2 / deterministic L3 truncation of the real source text. Behaviour is unchanged for normal summaries and already-handled closed blocks.

## Validation
- `_sanitize_reasoning_summary` unit test: closed->clean, unclosed->"", pipe/bracket/prose shapes->"", empty->"".
- `_call_llm_for_summary` integration: an unclosed `<think>` response returns `""` (without the fix it returns the raw reasoning blob — proven by temporarily reverting the wire-in).
- Escalation test: a reasoning-only primary escalates to the fallback model instead of accepting the empty result.
- `ruff check .` -> clean; full suite `1578 passed, 12 xfailed`; `scripts/validate_release.sh --full` -> `PASS: release validation full`.

## Notes
- The same closed-tag-only strip is used in `extraction.py`, `tools.py`, and `sanitize.py`; those non-summary paths are left for a focused follow-up so this PR stays scoped to the persisted-summary leak.
- Rollback: revert the commit; `_call_llm_for_summary` returns to `_strip_reasoning_blocks(content).strip()`.

## Refs
Mirrors lossless-claw `sanitizeReasoningSummaryText` (src/summarize.ts), which forces the empty-summary path on `reasoningOnly` output.

### DIFF
--- a/escalation.py
+++ b/escalation.py
@@ -35,6 +35,24 @@ _THINK_BLOCK_RE = re.compile(
     re.IGNORECASE | re.DOTALL,
 )
 
+# Matches the *start* of a reasoning block with no required close. Applied to
+# text after closed <think>...</think> pairs have been stripped: if what
+# remains still begins with a reasoning marker, the model emitted an *unclosed*
+# block (typically because it ran into max_tokens before the closing tag), and
+# the leftover raw reasoning must not be persisted as the summary. Covers the
+# angle-tag family plus pipe-delimited (<|think|>), bracket ([think]), and
+# prose-header (``Thinking Process:`` / ``Chain of thought:``) shapes.
+_REASONING_START_RE = re.compile(
+    r"^\s*(?:"
+    r"<\s*(?:think|thinking|reasoning|thought|REASONING_SCRATCHPAD)(?:\s[^>]*)?>"
+    r"|<\|\s*(?:start_of_)?(?:think|thinking|reasoning|thought)\s*\|>"
+    r"|\[\s*(?:think|thinking|reasoning|thought)\s*\]"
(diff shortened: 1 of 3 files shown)


stephenschoettler (approved):
Approved via Hermes Kanban reviewer gate t_5128458c. Evidence: live CI green 6/6, no unresolved review threads, and combined local queue validation passed before merge.

PR #348, opened 2026-07-07
Budget L3 deterministic truncation in tokens, not chars*4 (CJK overshoot)

ai-ag2026:
## Summary
- Add `truncate_text_to_tokens(text, max_tokens, *, from_end=False)` in `tokens.py`: an exact tiktoken-encoder cut when available, with a script-density-aware char fallback (bounded-corrected so it never exceeds the budget) when tiktoken is absent.
- Rewrite `escalation._deterministic_truncate` to budget its head+tail cut in **tokens** (reserving the marker's token cost), instead of a flat `char_budget = max_tokens * 4`.

## Why
HL counts tokens accurately with tiktoken everywhere else, but the Level-3 deterministic-truncation fallback sized its cut with chars*4. For CJK / dense scripts (~1-2 chars/token) that overshoots the real token budget ~2-4x, so an L3 summary aimed at N tokens could emit 2-4N — overflowing the very budget L3 exists to guarantee and defeating overflow recovery on non-English conversations. Latin text is materially unchanged (chars*4 was already about right there).

## Validation
- `truncate_text_to_tokens` respects the budget (head/tail, short text, zero budget).
- `_deterministic_truncate` on dense CJK yields far fewer tokens than the reconstructed old chars*4 output and stays within ~1.4x of the budget (vs the old ~2-4x); converges; Latin head/tail preserved.
- `ruff check .` -> clean; full suite `1579 passed, 12 xfailed`; `scripts/validate_release.sh --full` -> `PASS: release validation full`.

## Notes
- Test-environment caveat baked into the test: CI has no tiktoken, so the char fallback over-counts the ASCII marker at the CJK density; the CJK test therefore compares against the reconstructed old chars*4 output (env-robust) rather than an absolute token bound. With tiktoken the result lands within a token or two of the budget.
- Rollback: revert; `_deterministic_truncate` returns to the chars*4 cut.

## Refs
Surfaced by a comparison against lossless-claw, which budgets its deterministic cap with the Unicode-aware estimateTokens (`compaction.ts` `capSummaryText`) that HL's L3 bypassed.

### DIFF
--- a/escalation.py
+++ b/escalation.py
@@ -400,12 +400,30 @@ def _deterministic_truncate(text: str, max_tokens: int) -> str:
         # Budget too small to afford the head/tail marker; single head cut.
         return truncate_text_to_tokens(text, max_tokens)
 
-    body_budget = max_tokens - marker_tokens
-    head_tokens = body_budget // 2
-    tail_tokens = body_budget - head_tokens
-    head = truncate_text_to_tokens(text, head_tokens)
-    tail = truncate_text_to_tokens(text, tail_tokens, from_end=True)
-    return head + _L3_TRUNCATION_MARKER + tail
+    def assemble(body_tokens: int) -> str:
+        head_tokens = body_tokens // 2
+        tail_tokens = body_tokens - head_tokens
+        head = truncate_text_to_tokens(text, head_tokens)
+        tail = truncate_text_to_tokens(text, tail_tokens, from_end=True)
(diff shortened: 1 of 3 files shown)


stephenschoettler (approved):
Approved via Hermes Kanban reviewer gate t_5128458c. Evidence: live CI green 6/6, no unresolved review threads, pair validation with #347 passed, and combined local queue validation passed before merge.

PR #358, opened 2026-07-09
Fix foreground recovery after late auxiliary reclassification

stephenschoettler:
## Summary
- Restore the previous foreground session when a child session is reclassified as auxiliary at first ingest.
- Keep `lcm_status` and default `lcm_grep` scoped to the real foreground session while the auxiliary side channel remains active.
- Add regression coverage for the late background-review marker path and preserve the no-durable-write invariant for auxiliary first-ingest sessions.

## Background
This follows up the Codex P2 from #357: if bind-time auxiliary detection misses a background-review child, first-ingest recovery was marking the child auxiliary but clearing the foreground pointer, which could make current-session tools fall back to the auxiliary child.

## Validation
- Focused auxiliary/session-filtering pytest suite: passed (`20 passed`).
- Ruff on changed files: passed.
- Diff whitespace check: passed.
- Independent reviewer pass completed before publication.

## Risk
Focused hotfix only. The full repository test suite was not rerun; the targeted suite covers the late auxiliary reclassification path, existing auxiliary no-pollution behavior, active-clone packaging paths, and foreground session-filtering behavior.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -47,6 +47,37 @@ def _host_forwards_registered_tool_messages(ctx) -> bool:
     return bool(capability)
 
 
+def _engine_bound_session_id(engine) -> str:
+    """Return the lifecycle/ingest session bound on an LCM engine.
+
+    ``current_session_id`` is an operator-facing foreground view and can differ
+    from the bound ingest session while an auxiliary side channel is active.
+    Post-turn ingest rebinding must use the bound id or it can append a resumed
+    foreground turn to a stale auxiliary child.
+    """
+    return str(
+        getattr(engine, "bound_session_id", "")
+        or getattr(engine, "_session_id", "")
+        or ""
+    )
+
+
+def _ensure_engine_bound_to_session(
+    active_engine,
+    session_id: str,
+    *,
+    platform: str = "",
+    conversation_id: str = "",
+) -> None:
+    session_id = str(session_id or "")
+    if session_id and _engine_bound_session_id(active_engine) != session_id:
+        active_engine.on_session_start(
+            session_id,
+            platform=platform,
+            conversation_id=conversation_id or None,
+        )
+
+
 def register(ctx):
     """Plugin entry point — register the LCM context engine and tools."""
     from .config import LCMConfig
@@ -205,15 +236,12 @@ def register(ctx):
                 # can deliver stale lane metadata alongside the correct active
                 # session id; rebinding a clone on conversation_id mismatch
                 # alone would move it away from the runtime it is serving.
-                if (
-                    session_id
-                    and str(getattr(active_engine, "current_session_id", "") or "") != session_id
-                ):
-                    active_engine.on_session_start(
-                        session_id,
-                        platform=platform,
-                        conversation_id=conversation_id or None,
-                    )
+                _ensure_engine_bound_to_session(
+                    active_engine,
+                    session_id,
+                    platform=platform,
+                    conversation_id=conversation_id,
+                )
                 active_engine.ingest(history)
             except Exception as exc:
                 logger.debug("LCM post_llm_call ingest error: %s", exc)
--- a/aux_session.py
+++ b/aux_session.py
@@ -586,6 +586,7 @@ class AuxiliarySessionMixin:
         kwargs: Dict[str, Any],
         session_id: str = "",
         include_explicit: bool = True,
+        require_auxiliary_frame: bool = True,
     ) -> str:
         explicit = str(kwargs.get("parent_session_id") or "")
         if include_explicit and explicit:
@@ -604,7 +605,7 @@ class AuxiliarySessionMixin:
                 if frame is None:
                     return ""
                 caller_self = frame.f_locals.get("self")
-                if not self._caller_is_auxiliary_agent_frame(caller_self):
+                if require_auxiliary_frame and not self._caller_is_auxiliary_agent_frame(caller_self):
                     frame = frame.f_back
                     continue
                 parent = str(getattr(caller_self, "_parent_session_id", "") or "")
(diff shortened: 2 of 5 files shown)


stephenschoettler:
@codex review

Please review this hotfix specifically for the #357 P2 foreground-preservation concern. Confirm whether late first-ingest auxiliary reclassification now restores/keeps the prior foreground view and does not reintroduce durable writes for auxiliary review sessions.

stephenschoettler:
@codex review

I pushed a follow-up commit for your P2. The post-LLM hook now compares the incoming session id against the engine's bound ingest/lifecycle session, not the operator-facing foreground `current_session_id`, before rebinding. Please re-review the late auxiliary reclassification path for foreground persistence into stale auxiliary sessions.

stephenschoettler:
@codex review

I pushed a follow-up for the stacked late auxiliary-bind case. `_remember_foreground_rebind_candidate()` now carries forward the already-saved original foreground when a second provisional auxiliary session binds before the first reclassification, and the regression test covers two late-marked review children restoring `foreground-session`.

stephenschoettler:
@codex review

I pushed a follow-up for the stale-candidate case. After a provisional foreground bind successfully persists as a real foreground ingest, the saved rebind candidate is cleared; the new regression covers foreground A -> confirmed foreground B -> late-marked review child C, and verifies C restores to B rather than stale A.

stephenschoettler:
@codex review

I pushed a follow-up for the tool-call ingest path. Candidate clearing is now centralized and called after successful post-turn ingest and LCM tool-call ingest; the new regression covers foreground A -> foreground B first persisted by `handle_tool_call(..., messages=...)` -> late-marked review child C, and verifies C restores to B.

stephenschoettler:
@codex review

I pushed a follow-up for the compression-boundary foreground case. Successful `_continue_compression_boundary()` binds now clear the provisional rebind candidate immediately, and the new regression covers foreground A -> compression-boundary foreground B -> late-marked review child C restoring to B.

stephenschoettler:
@codex review

I pushed a follow-up for the fresh foreground-without-ingest case. Late auxiliary restore now prefers the in-process parent session when that parent is a known normal foreground binding, so foreground A -> foreground B with no ingest yet -> late-marked review child C restores to B rather than stale A. The new regression covers that exact path.

stephenschoettler:
@codex review

I pushed a follow-up for the stale-parent restore case. Late auxiliary recovery now only prefers the in-process parent when that parent matches the foreground snapshot displaced by the provisional bind; otherwise it restores the saved active foreground. I also widened the provisional-parent frame check so an unmarked-but-real foreground child still preserves the un-ingested foreground case. New regression covers foreground A -> confirmed foreground B -> late-marked review child from stale parent A, and verifies the operator-facing foreground stays on B.

stephenschoettler:
@codex review

Addressed the latest stale-parent/un-ingested foreground case at `bce4428`. The rebind snapshot now records the candidate's observed in-process parent, so a later stale-parent review child can distinguish a real unpersisted foreground displacement from stacked provisional auxiliary binds. Added regression for foreground A stored -> foreground B bound/un-ingested -> stale review child from A; existing stacked auxiliary regression still passes.

stephenschoettler:
@codex review current head f82437a. Narrow follow-up for your branched-foreground P2: uses the Hermes state.db _branched_from marker as the authoritative signal for a real un-ingested foreground branch, preserves the stacked provisional-aux regression, and adds the failing B(parent A) -> stale review C(parent A) regression. Local validation: python -m pytest tests/test_auxiliary_first_ingest.py tests/test_lcm_engine.py tests/test_lcm_command.py tests/test_packaging_install.py -q; ruff check __init__.py aux_session.py engine.py tests/test_auxiliary_first_ingest.py; git diff --check.

stephenschoettler:
@codex review

stephenschoettler:
@codex review current head 06dddaf. Narrow follow-up for your nested late auxiliary P2: when a second provisional late auxiliary binds with parent equal to the first provisional candidate, LCM now only promotes that first candidate as the restore target if it is a confirmed foreground branch via state.db _branched_from (or has no recorded parent). Added regression for foreground F -> late aux A(parent F) -> nested late aux B(parent A), proving both auxiliary sessions stay empty and the operator-facing foreground restores to F. Local validation: python -m pytest tests/test_auxiliary_first_ingest.py tests/test_lcm_engine.py tests/test_lcm_command.py tests/test_packaging_install.py -q; ruff check engine.py tests/test_auxiliary_first_ingest.py; git diff --check.

stephenschoettler:
@codex review current head c135a40. Narrow follow-up for your preflight-ingest candidate-clearing P2: successful preflight/compress ingest paths now clear the provisional foreground rebind candidate once the bound session is confirmed, including no-new-message and all-dropped-message paths. Added regression for foreground A -> preflight/compress foreground B -> late stale-parent review child C, verifying C stays auxiliary and the operator-facing foreground remains B. Local validation: python -m pytest tests/test_auxiliary_first_ingest.py tests/test_lcm_engine.py tests/test_lcm_command.py tests/test_packaging_install.py -q; ruff check engine.py tests/test_auxiliary_first_ingest.py; git diff --check.

stephenschoettler:
@codex review current head fa4c553.

Narrow follow-up for your compaction writes P2: compaction preflight/compress now run the same late auxiliary reclassification guard before `_bypasses_lcm_context_management()` and before any direct `_ingest_messages()` write, so missed background-review children become side-channel/stateless before compaction writes. Added regressions for preflight-first-write and compress-first-write auxiliary reclassification preserving foreground and avoiding durable `review-session` rows.

Local validation: `python -m pytest tests/test_auxiliary_first_ingest.py tests/test_lcm_engine.py tests/test_lcm_command.py tests/test_packaging_install.py -q` -> 818 passed; `ruff check engine.py compaction.py tests/test_auxiliary_first_ingest.py` -> pass; `git diff --check` -> pass. GitHub CI run 29011964607 is green.

PR #366, opened 2026-07-09
fix: preserve initial user anchor through compaction

### LINKED ISSUE #355: Compaction can remove the only user message → Qwen-family chat templates reject the request (400: 'No user query found in messages')
Found on v0.12.0 driving a local Qwen3.5-family model (Ornith-9B via oMLX/mlx-lm).

**Symptom:** after a leaf compaction on an agentic session (one user query followed by a long assistant/tool chain), the next model call fails `400: Chat template error: No user query found in messages.` — the client (Hermes) retries the deterministic 400 and surfaces 'model provider failed after retries'.

**Root cause:** `LCMEngine._leading_anchor_count` anchors only the system prompt, so the session's single user message is eligible for compaction. Once folded into a DAG summary, no user-role message remains (fresh tail is all assistant/tool), and Qwen-family chat templates hard-raise (`chat_template.jinja`: `raise_exception('No user query found in messages.')` when the trailing scan finds no real user query).

**Fix that works for us** (patched locally, verified end-to-end — compacted shape now renders, control still 400s):
1. Anchor the first user message too:
```python
n = 1 if (messages and messages[0].get('role') == 'system') else 0
for i in range(n, min(len(messages), n + 3)):
    if messages[i].get('role') == 'user':
        return i + 1
return n
```
2. Adapt the overflow-recovery call site so the multi-message anchor isn't dropped: `working_messages[1 if leading_anchor_count else 0:]` for the tail argument (identical under old semantics).

Anchoring the original query also keeps the task grounding in raw context, which the existing docstring already argues for. Happy to PR if useful.

stephenschoettler:
## Summary
- preserve the initial real user message across compaction and replay
- keep summary replacement and overflow recovery aligned with multi-message anchors
- add regressions for regular and overflow compaction paths

## Validation
- 5 focused regression tests passed
- 803-test suite passed
- ruff passed
- `git diff --check` passed

Closes #355

### DIFF
--- a/compaction.py
+++ b/compaction.py
@@ -439,10 +439,14 @@ class CompactionMixin:
                     pressure_messages[candidate_start:fresh_tail_start],
                 )
             )
+            compactable_store_ids_by_message_id = self._get_store_id_map_for_messages(
+                working_messages[leading_anchor_count:]
+            )
             raw_compactable_pairs = [
                 (working_msg, pressure_msg)
                 for working_msg, pressure_msg in compactable_pairs
                 if not self._is_replayed_context_scaffold_message(working_msg)  # type: ignore[attr-defined]
+                or id(working_msg) in compactable_store_ids_by_message_id
             ]
             if len(raw_compactable_pairs) != len(compactable_pairs):
                 dropped_replayed_scaffold_messages = True
--- a/compaction.py
+++ b/compaction.py
@@ -647,7 +647,11 @@ class CompactionMixin:
             )
             self._dag.add_node(node)
             self._maybe_gc_compacted_tool_results(compacted_chunk, source_store_ids)
-            self._last_compacted_store_id = max(consumed_store_ids) if consumed_store_ids else 0
+            if consumed_store_ids:
+                self._last_compacted_store_id = max(
+                    self._last_compacted_store_id,
+                    max(consumed_store_ids),
+                )
             self._persist_frontier_marker()
 
             pressure_remaining_messages = pressure_messages[leading_anchor_count + selected_raw_len:]
--- a/compaction.py
+++ b/compaction.py
@@ -775,6 +775,10 @@ class CompactionMixin:
             self._write_generated_ignored_placeholder_hash_ordinals(
                 self._generated_placeholder_digest_ordinals_for_active_replay(sanitized_messages)
             )
+            if dropped_replayed_scaffold_messages and sanitized_messages != messages:
+                snapshot_writer = getattr(self, "_write_active_replay_snapshot", None)
+                if callable(snapshot_writer):
+                    snapshot_writer(sanitized_messages)
             return sanitized_messages
 
         # Step 6: Check if condensation is needed
@@ -853,5 +857,9 @@ class CompactionMixin:
         self._write_generated_ignored_placeholder_hash_ordinals(
             self._generated_placeholder_digest_ordinals_for_active_replay(compressed)
         )
+        if compressed != messages:
+            snapshot_writer = getattr(self, "_write_active_replay_snapshot", None)
+            if callable(snapshot_writer):
+                snapshot_writer(compressed)
 
         return compressed
--- a/compaction.py
+++ b/compaction.py
@@ -617,7 +617,15 @@ class CompactionMixin:
             source_lineage_chunk = [
                 message for message in source_lookup_chunk if id(message) not in dependent_reply_message_ids
             ]
+            pre_frontier_sources = self._get_formerly_anchored_user_source_map(
+                working_messages[leading_anchor_count:]
+            )
             source_store_ids = self._get_store_ids_for_messages(source_lineage_chunk)
+            source_store_ids.extend(
+                pre_frontier_sources[id(message)]
+                for message in source_lineage_chunk
+                if id(message) in pre_frontier_sources
+            )
             source_store_ids = sorted(dict.fromkeys(source_store_ids))
             consumed_store_ids = self._get_store_ids_for_messages(source_lookup_chunk)
             consumed_store_ids = sorted(dict.fromkeys(consumed_store_ids))
--- a/compaction.py
+++ b/compaction.py
@@ -433,16 +433,29 @@ class CompactionMixin:
                 break
 
             candidate_start = leading_anchor_count
-            while (
-                candidate_start < fresh_tail_start
-                and self._is_replayed_context_scaffold_message(working_messages[candidate_start])
-            ):
-                candidate_start += 1
-            if candidate_start > leading_anchor_count:
+            compactable_pairs = list(
+                zip(
+                    working_messages[candidate_start:fresh_tail_start],
+                    pressure_messages[candidate_start:fresh_tail_start],
+                )
+            )
+            raw_compactable_pairs = [
+                (working_msg, pressure_msg)
+                for working_msg, pressure_msg in compactable_pairs
+                if not self._is_replayed_context_scaffold_message(working_msg)  # type: ignore[attr-defined]
+            ]
+            if len(raw_compactable_pairs) != len(compactable_pairs):
                 dropped_replayed_scaffold_messages = True
-                working_messages = working_messages[:leading_anchor_count] + working_messages[candidate_start:]
-                pressure_messages = pressure_messages[:leading_anchor_count] + pressure_messages[candidate_start:]
-                candidate_start = leading_anchor_count
+                working_messages = (
+                    working_messages[:candidate_start]
+                    + [working_msg for working_msg, _pressure_msg in raw_compactable_pairs]
+                    + working_messages[fresh_tail_start:]
+                )
+                pressure_messages = (
+                    pressure_messages[:candidate_start]
+                    + [pressure_msg for _working_msg, pressure_msg in raw_compactable_pairs]
+                    + pressure_messages[fresh_tail_start:]
+                )
                 n = len(working_messages)
                 fresh_tail_start = max(0, n - self._config.fresh_tail_count)
                 if fresh_tail_start <= leading_anchor_count:
--- a/compaction.py
+++ b/compaction.py
@@ -386,10 +386,23 @@ class CompactionMixin:
         # or provider context after the durable row has been written.
         working_messages = self._ingest_messages(messages)
         ingest_cleanup_changed_active_context = working_messages != messages
+        restored_working_messages = [
+            self._strip_coalesced_generated_summary_scaffold(message)  # type: ignore[attr-defined]
+            for message in working_messages
+        ]
+        restored_coalesced_summary_tail = any(
+            restored is not original
+            for original, restored in zip(working_messages, restored_working_messages)
+        )
+        if restored_coalesced_summary_tail:
+            working_messages = restored_working_messages
+            ingest_cleanup_changed_active_context = True
         anchor_source_messages = list(working_messages)
         pressure_messages = messages if len(messages) == len(working_messages) else working_messages
+        if restored_coalesced_summary_tail:
+            pressure_messages = working_messages
         leaf_compacted_this_turn = False
-        dropped_replayed_scaffold_messages = False
+        dropped_replayed_scaffold_messages = restored_coalesced_summary_tail
         leaf_passes = 0
         critical_budget_pressure = self._critical_budget_pressure_reached(
             observed_tokens=observed_prompt_tokens,
--- a/compaction.py
+++ b/compaction.py
@@ -37,6 +37,22 @@ class CompactionMixin:
         if callable(maybe_reclassify):
             maybe_reclassify()
 
+    @staticmethod
+    def _split_leading_anchor_and_context_tail(
+        messages: List[Dict[str, Any]],
+        leading_anchor_count: int,
+    ) -> tuple[Optional[Dict[str, Any]], List[Dict[str, Any]], List[Dict[str, Any]]]:
+        leading_anchor_count = max(0, min(leading_anchor_count, len(messages)))
+        leading_anchors = messages[:leading_anchor_count]
+        system_msg = (
+            leading_anchors[0]
+            if leading_anchors
+            and isinstance(leading_anchors[0], dict)
+            and leading_anchors[0].get("role") == "system"
+            else None
+        )
+        return system_msg, leading_anchors, messages[leading_anchor_count:]
+
     def should_compress(self, prompt_tokens: int = None) -> bool:
         if self._bypasses_lcm_context_management():
             if self._compression_boundary_cooldown_active():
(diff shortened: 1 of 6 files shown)


Tosko4 (changes requested):
Requesting changes on `105b69e1f373b876cb404c13c80df48ccbef6ddc`.

I inspected the complete three-file diff, issue #355, all current PR comments/reviews, the unresolved GraphQL review thread, every changed-symbol call site, and CI on this exact head. The core direction is sound, but three replay/overflow cases still block approval:

1. Restart replay duplicates durable rows. After normal compaction returns the initial user as a raw leading anchor, `_reconcile_ingest_cursor_from_store()` only advances past the system scaffold. A direct restart + one-follow-up probe grew the session from 8 rows to 15 instead of 9, duplicated the initial user, and persisted derived summary/tail content again. This confirms the existing thread at `compaction.py:773` and also shows synthetic-summary contamination.

2. Forced overflow can still remove the only real user when that user is inside `fresh_tail_count`. With `fresh_tail_count=32`, a 40-token assembly cap, and `[system, user, oversized assistant]`, `_leading_anchor_count()` returned 1 and recovery returned roles `[system, assistant]`. The existing assertion at `tests/test_lcm_engine.py:20601` currently pins this shape. That still leaves Qwen without a user query.

3. Generated user-role LCM summaries are classified as prompt-bearing anchors. A `[Recent Summary ...][Expand ...]` message is correctly recognized by `_is_replayed_context_scaffold_message()`, but `_is_prompt_bearing_user_message()` returns true and `_leading_anchor_count()` returns 2. In a second compaction probe, the old summary appeared twice in active context. Blank user content is classified as prompt-bearing too.

Validation run:

- `python3 -m pytest tests/test_lcm_engine.py -q -o addopts= -k 'single_initial_user or later_user_keeps_initial'` with the CI-compatible ContextEngine stub: `3 passed, 700 deselected`.
- `python3 -m pytest tests/test_lcm_core.py::TestAssemblyBudgetSelection::test_non_contiguous_raw_user_tail_replay_does_not_duplicate_durable_rows tests/test_lcm_core.py::TestAssemblyBudgetSelection::test_non_contiguous_preserved_prompt_replay_does_not_duplicate_durable_rows -q -o addopts=`: `2 passed`.
- `PYTHON=python3 scripts/validate_release.sh --full --keep-going --output <fresh-dir>`: passed every gate; focused pytest `409 passed`, full pytest `1608 passed, 12 xfailed`, low-fd pytest `1608 passed, 12 xfailed`, benchmark smoke, stress smoke, and release stress all passed.
- `uvx --from ruff==0.15.13 ruff check .`: passed.
- GitHub CI is green for lint, workflow-lint, and Python 3.11-3.14 on this SHA.

An independent read-only Codex CLI review on `gpt-5.6-sol` with `model_reasoning_effort="xhigh"` independently confirmed the same three blocker classes.

Limitations: I did not run a live Qwen/oMLX provider call; the provider-facing verdict is based on the exact active-context role shapes. Please add restart regressions for no-delta and one-follow-up replay, cover forced overflow with the sole user inside the fresh tail, and exclude generated/blank scaffolds from real-user anchor classification before the next review.

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

Tosko4 (changes requested):
Requesting changes on `72fd725d498f2f443d2ec2a6cb332381c152b23d`.

I re-reviewed the full branch and grouped the remaining failures by root cause so we do not keep fixing one replay shape at a time.

1. **P1 — restart replay proof is incomplete and depends on the current assembly cap** (`reconcile.py:461`, `reconcile.py:475`). Frontier-zero overflow can return `[system, user, small durable assistant tail]`; after restart, reconciliation accepts only the two-message raw anchor prefix and stores the already-durable tail again (4 rows became 5, with two tail copies). The annotated-system path has a second ambiguity: it uses the current cap as evidence for what the previous process emitted. Increasing or disabling the cap caused a genuinely new assistant message identical to an omitted tail to be skipped, while lowering the cap duplicated a fully replayed tail. Replay proof cannot depend on mutable current config.

2. **P1 — durable prompt-bearing users are still located positionally and classified with active-message semantics** (`reconcile.py:446`, `reconcile.py:459`, `engine.py:1527`, `engine.py:1557`). With durable `[system, generated user scaffold, real user, assistant tail]`, replaying `[annotated system, real user]` advanced only over the system scaffold and duplicated the real user (4 rows became 5). Separately, a durable externalized user payload that matches `ignore_message_patterns` is ignored with `stored_row=True` but `_is_prompt_bearing_user_message()` still classifies it as prompt-bearing. One real user plus that ignored row therefore disables the only real-user anchor. Please compare against the first durable prompt-bearing user and apply stored-row ignore semantics during durable classification.

3. **P1 — the compaction frontier can jump across an omitted raw row with no DAG lineage** (`compaction.py:633`, `compaction.py:663`). Starting from overflow output `[system, user, row 4]`, a later user demoted the initial anchor and the next node recorded source IDs `[2, 4]`. The frontier then advanced to 4, permanently stranding durable row 3 behind the exclusive mapping frontier with no summary source path. Any frontier advance must prove contiguous coverage or keep the frontier below the first gap.

4. **P2 — a replay summary at the fresh-tail boundary is emitted twice** (`compaction.py:435`). The new filter removes generated scaffolds only from the compactable slice before `fresh_tail_start`. If the first compaction returns `[system, user, summary]`, a later real user demotes the initial user while that summary remains at the fresh-tail boundary. The next compaction emits the old summary once from the DAG and once from the raw tail. I reproduced two copies even though the scaffold was correctly absent from the summarizer input.

5. **P1 — the preserved user can still produce an invalid complete role sequence** (`engine.py:4812`, `engine.py:4855`). With one DAG summary and an ordinary assistant fresh tail, assembly returns `[system, user, assistant(summary), assistant(tail)]`. A tool-call tail returns `[system, user, assistant(summary), assistant(tool_calls), tool]`; another boundary probe returned consecutive user roles. The tests currently assert the leading prefix, not the complete provider-visible sequence. Please add full-sequence coverage for plain assistant and tool-call/result tails while preserving tool-pair contiguity.

- Anchor/replay-focused `tests/test_lcm_engine.py` selection: `16 passed, 708 deselected`.
- Full pytest: `1629 passed, 12 xfailed`.
- Seven deterministic boundary probes reproduced the restart duplication/data-loss, durable-user classification, lineage gap, summary duplication, and role-sequence failures above.
- `uvx --from ruff==0.15.13 ruff check .`: passed.
- The earlier exact-head full release validation also passed focused/full/low-fd pytest, benchmark smoke, smoke stress, and release stress.
- GitHub CI is green for workflow lint, lint, and Python 3.11–3.14 on this SHA.

I did not run a live Qwen/oMLX or strict-alternation provider call. The storage, replay, lineage, and context-duplication findings are direct deterministic reproductions; the provider impact of finding 5 is based on the emitted role sequence.

Tosko4 (changes requested):
One additional P1 blocker on `72fd725d498f2f443d2ec2a6cb332381c152b23d` surfaced in the final adversarial pass and is distinct from the durable-user-positioning issue in my earlier review.

**P1 — literal user input that matches rendered-summary syntax is silently discarded** (`engine.py:3623`, `reconcile.py:461`, `compaction.py:442`). Scaffold classification currently relies on role/content shape alone; it does not require persisted generation provenance or even an existing DAG node.

I reproduced this with an on-disk frontier-zero restart. After overflow returned the raw `[system, user]` anchor, the next genuine user delta was:

```text
[Recent Summary (d0, node 999)]
This is literal user input.
[Expand for details: literal]
```

The engine classified it as `scaffold=True` and `prompt_bearing=False`. Reconciliation advanced over all three incoming messages as `replayed durable tail`; durable row count remained `5 → 5`, and the literal user message had zero stored copies. The same content-only classifier is also used by the compaction filter, so this shape can be excluded from summarizer input and active assembly as well.

Generated-scaffold recognition needs provenance. If provenance is absent or ambiguous, preserve the message. Please add regressions proving that an exact user-authored wrapper referencing a nonexistent node persists once across restart and remains prompt-bearing/available for compaction, while a genuinely generated scaffold with durable provenance is still skipped safely.

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

Tosko4 (changes requested):
Requesting changes on `ce09140318148ebea7300aaf527a38818d1ac15e`.

Two provider-facing regressions remain despite the green suite:

1. **P1: preserve a literal summary-shaped latest user prompt.** `engine.py:4801` now routes the reverse scan through `_is_prompt_bearing_user_message()` without allowing a durable literal summary-shaped user. Under assembly pressure, a real newest prompt matching `[Recent Summary ...] ... [Expand for details ...]` is skipped and an older user objective is preserved instead. A direct capped `_assemble_context()` probe returned the old objective and omitted the newest request. The same classification can advance a singleton new literal user delta as a `skipped scaffold-only prefix`, so this needs provenance-based handling rather than shape-only filtering.

2. **P1: keep the complete provider role sequence valid after adding the raw user anchor.** With `system,user` preserved, one DAG summary, and an ordinary assistant fresh tail, `_assemble_context()` returns `system,user,assistant,assistant`. Strict-alternation templates reject that sequence. The repair needs to preserve assistant tool calls and their contiguous tool results as a unit, not fix alternation by dropping or separating tool output.

Validation on this head:
- `python -m pytest -q -o addopts=`: `1647 passed, 12 xfailed`
- `scripts/validate_release.sh --full --keep-going`: PASS
- direct assembly/reconciliation probes reproduced both blockers

Please add regressions through the real assembly path for the literal-summary user case and the complete role sequence, including a tool-call/result tail.

Tosko4 (changes requested):
Follow-up on the same head from the completed restart-reconciliation matrix. Two additional P1 blockers remain:

1. **Fresh system content can be discarded by scaffold shape alone.** `reconcile.py:452-454` accepts any system message containing the two LCM note phrases as a generated system anchor. It does not prove equality with the durable system prompt plus the exact generated annotation. In a positive-frontier restart probe, a brand-new system payload containing those phrases followed by the durable initial user advanced the cursor through both messages. The durable row count stayed unchanged and the new system payload was stored zero times. Generated-anchor recognition needs exact provenance/content proof, not phrase containment.

2. **Restart replay proof depends on the current assembly cap rather than the cap that produced the active snapshot.** `reconcile.py:490-519` evaluates partial/full suffix feasibility using `_effective_assembly_token_cap()` after restart. A partial suffix assembled under the prior cap and replayed after increasing the cap reconciled only through the generated scaffold (`cursor=3/4`), so the already-durable visible suffix would be appended again. The inverse cap change similarly duplicates rows. Replay proof must be historical/cap-independent, with regressions for unchanged, increased, disabled, and decreased caps plus no/new/repeated deltas.

These are separate from the two assembly blockers in my previous review and reproduced on `ce09140318148ebea7300aaf527a38818d1ac15e`.

Tosko4 (changes requested):
Requesting changes on `ce09140318148ebea7300aaf527a38818d1ac15e`.

The anchor-preservation direction is right, but three cases still block approval:

1. `engine.py:4889-4942` can emit invalid provider ordering. With a preserved `[system, user]` anchor, one DAG summary, and an ordinary assistant fresh tail, the complete role sequence is `[system, user, assistant, assistant]`. An assistant tool-call tail produces the same adjacent-assistant boundary before its tool result. The current tests assert the leading pair but not the full emitted sequence.

2. Literal summary-shaped user input can still replace current intent with stale intent. `engine.py:4801` uses the shape-rejecting prompt predicate while selecting the latest omitted user objective, so a genuine newest `[Recent Summary ...][Expand ...]` user prompt is skipped and an older objective is preserved instead. The provenance check at `engine.py:3706-3727` also treats an exact rendering of a real local node as generated even when the user appends a new request, so that authored content remains vulnerable to scaffold filtering.

3. Restart replay proof depends on the assembly cap configured after restart. Starting from a frontier-zero capped replay of `[system, user, durable assistant tail]`, then restarting with the cap disabled and appending one new user produced 7 durable rows instead of 6; the visible assistant tail was stored twice. `reconcile.py:490-523` accepts only the shorter anchor prefix once the current cap no longer proves the historical partial-tail shape.

- Targeted anchor/replay/overflow selection: 26 passed.
- Core/engine/packaging suite: 1,057 passed.
- Full pytest: 1,647 passed, 12 xfailed.
- `PYTHON=python3 scripts/validate_release.sh --full --keep-going --output <fresh-dir>`: PASS.
- GitHub lint, workflow-lint, and Python 3.11-3.14 checks are green.

I did not run a live provider request, but the emitted role sequence and durable-row duplication are deterministic at the engine boundary. Please add regressions for complete role/tool ordering, literal-summary latest-user provenance, and restart replay across unchanged/increased/disabled/decreased caps.

stephenschoettler:
@codex review

stephenschoettler:
@codex review

stephenschoettler:
@codex review

Tosko4 (changes requested):
Requesting changes on `71d505f75e8f15f07995831b0c73ed6a571468a5`.

I reproduced one branch-specific issue that needs fixing before this can land: `_write_active_replay_snapshot()` builds identities through `_message_replay_identity()`, which dereferences externalized payload refs. The serialized snapshot metadata therefore contains the full payload instead of the placeholder. That duplicates potentially large or sensitive externalized content into SQLite metadata and bypasses the externalization cleanup boundary.

Please keep replay snapshot identities bounded and non-dereferencing, or store a stable digest/ref, and add a regression proving the metadata never contains the payload body.

Two other issues I found, cross-conversation source mapping and tool-result stubs pushing the final context above the assembly cap, also reproduce on current `main`. I am not treating those as regressions from this PR; they should be tracked separately.

This branch has grown to 40 commits and roughly 4.1k added lines across replay, lineage, conversation scoping, and context assembly. I recommend fixing only the snapshot metadata issue here, updating the stale PR body and validation notes, and stopping further scope growth. If the current replay changes cannot be isolated cleanly, restarting from `main` in smaller slices will be safer than layering more edge-case fixes onto this branch.

stephenschoettler:
Superseded by #370, rebuilt from fresh main as a single commit limited to issue #355.

PR #370, opened 2026-07-13
fix: preserve the sole real user prompt through compaction

### LINKED ISSUE #355: Compaction can remove the only user message → Qwen-family chat templates reject the request (400: 'No user query found in messages')

stephenschoettler:
## Summary
- Rebuild issue #355 from fresh `main` as one commit that retains the sole real prompt-bearing user anchor through both normal compaction and forced-overflow recovery.
- Keep provider-visible context structurally valid when the fresh tail begins with an assistant turn by choosing a compatible synthetic summary role without splitting assistant/tool pairs.
- Limit the change to `compaction.py`, `engine.py`, and `tests/test_lcm_engine.py`.

## Why
- A single-query agentic session can otherwise compact away its only real user message, causing Qwen-family chat templates to reject the next request with `No user query found in messages`.
- This is a fresh-main replacement for #366, deliberately excluding its replay/snapshot machinery, cross-conversation mapping, and adjacent recovery work.

## Validation
- [x] Focused validation: `pytest tests/test_lcm_engine.py::TestSingleUserPromptAnchor -q` -> `13 passed`
- [ ] Default validation:
  - [ ] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q`
  - [x] `pytest -q` -> `1618 passed, 12 xfailed`
  - [ ] `bash -lc 'ulimit -n 1024 && pytest -q'`
  - [x] `git diff --check`
- [ ] Workflow validation, if workflows changed: `actionlint` (not applicable; no workflow files changed)
- [x] Additional validation: `pytest tests/test_lcm_engine.py -q` -> `713 passed`
- [x] Additional validation: `ruff check engine.py compaction.py tests/test_lcm_engine.py && ruff check .`
- [x] Independent exact-head reviewer gate -> pass

## Notes
- The exact grouped default test command and the `ulimit` variant were not rerun separately; the complete suite passed on this exact head.
- Structural provider-context invariants passed, including retained system/user anchors, valid role alternation, contiguous tool-call/result pairs, and prompt-cache behavior.
- No live provider-template HTTP smoke was run.
- Exactly one commit over fresh `main`; no replay/snapshot machinery was carried forward.

Fixes #355

### DIFF
--- a/compaction.py
+++ b/compaction.py
@@ -408,10 +408,9 @@ class CompactionMixin:
             n = len(working_messages)
             fresh_tail_start = max(0, n - self._config.fresh_tail_count)
 
-            # Keep only a real system prompt anchored. Gateway sessions may
-            # pass only conversation messages, so index 0 can be an old user
-            # turn; that must remain eligible for compaction instead of being
-            # replayed forever as fresh-looking intent.
+            # Keep a real system prompt and, while it is the only real user
+            # turn, its immediately following prompt. Gateway sessions may pass
+            # only conversation messages, so index 0 remains compactable.
             leading_anchor_count = self._leading_anchor_count(working_messages)
             if fresh_tail_start <= leading_anchor_count:
                 noop_reason = "no eligible raw backlog outside fresh tail"
@@ -664,8 +663,9 @@ class CompactionMixin:
                 leading_anchor_count = self._leading_anchor_count(working_messages)
                 compressed = self._assemble_overflow_recovery_context(
                     working_messages[0] if leading_anchor_count else None,
-                    working_messages[leading_anchor_count:],
+                    working_messages[1 if leading_anchor_count else 0:],
                     assembly_cap_override=recovery_assembly_cap,
+                    preserve_leading_user=leading_anchor_count == 2,
                 )
                 return self._finalize_forced_overflow_result(
                     working_messages,
@@ -683,8 +683,9 @@ class CompactionMixin:
                 try:
                     sanitized_messages = self._assemble_context(
                         active_context_messages[0] if leading_anchor_count else None,
-                        active_context_messages[leading_anchor_count:],
+                        active_context_messages[1 if leading_anchor_count else 0:],
                         assembly_cap_override=recovery_assembly_cap,
+                        preserve_leading_user=anchor_leading_count == 2,
                     )
                 finally:
                     self._pending_context_anchor_messages = None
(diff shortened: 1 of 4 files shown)


Tosko4 (changes requested):
I’m requesting changes on `fa6c779` for two blockers.

1. `engine.py:4706-4787` can emit `system, user, user, assistant` when the sole user prompt is retained, a DAG summary exists, and the selected fresh tail begins with an assistant turn. That avoids adjacent assistants by creating adjacent users, so the complete provider-visible sequence is still not structurally valid. The regression test at `tests/test_lcm_engine.py:11562-11595` only rejects adjacent assistant roles; it needs to assert the complete valid sequence while preserving contiguous tool-call/result pairs.

2. `engine.py:1489-1526` classifies prompt-bearing users through content-only summary/objective/todo scaffold checks. I reproduced genuine user-authored literals matching each wrapper being treated as non-real, with `_leading_anchor_count == 1`. That leaves the session’s sole raw prompt eligible for compaction, which defeats this fix for valid literal input. The distinction needs durable generation provenance or an equally unambiguous occurrence identity, plus collision coverage for literal user text.

- `python -m pytest tests/test_lcm_engine.py::TestSingleUserPromptAnchor tests/test_lcm_engine.py::TestAssemblyGuardrails tests/test_lcm_engine.py::TestPostCompactionIngestion -q -o addopts=` — 30 passed
- `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q -o addopts=` — 1028 passed
- `python -m pytest -q -o addopts=` — 1618 passed, 12 xfailed
- `scripts/validate_release.sh --full --keep-going` — passed
- Direct exact-head/base probes reproduced both blockers; GitHub CI is green

I did not run a live Qwen provider HTTP smoke.

PR #371, opened 2026-07-14
fix: never block token counting on tiktoken's network download

zab-01:
## Reworked per review — thank you @Tosko4, you were right

Your point 4 sent me back to instrument the actual stack, and the finding invalidates my original design entirely. This PR is now a different, much smaller change; the async executor, backlog semaphore, drain handle, and pre-warm are all **gone**, and the four post-hook tests pass **unmodified** (the hook is synchronous again).

## The real root cause (profiled, reproduced)

`tiktoken.get_encoding("cl100k_base")` downloads its BPE file over the network on first use when it isn't cached on disk. On a restricted-egress host that request hangs until the transport gives up. `_get_encoder()` sits under `count_tokens()` on the ingest path — which hosts run inside the synchronous `post_llm_call` hook — so the first turn after every process start blocked reply delivery.

cProfile of the first ingest of a ~45K-token history on a restricted-egress host:

```
1  128.108s  engine.py ingest
1  128.107s  engine.py _ingest_messages
2  127.852s  tokens.py count_message_tokens
1  127.791s  tiktoken load_tiktoken_bpe → requests.get(...)
```

Second turn, same process: 0.1s. You were right that `ingest()` does no summarization — my "two 60s summary timeouts" reading of the 127s was a coincidence of numbers; the duration is the network transport's give-up time, which is why it was so repeatable.

## The fix

`tokens.py` only: load the encoder in a daemon thread. The first caller waits up to 2s (the common cached-on-disk case still gets exact counts immediately); after that, callers never wait — they use the existing char-based estimator until the loader finishes, then the encoder is adopted and `_count_tokens_cached` is cleared (counts for identical text change on adoption). A permanently blocked download degrades to the estimator for the life of the process — the same posture as tiktoken being uninstalled, which this module already supports.

Measured on the same blocked-egress host: first-turn hook **128.5s → 2.2s**, fully synchronous.

## Your review points

1–3 (stale rebind, foreground serialization, executor lifecycle): all real, all mooted — that machinery no longer exists in this PR.
4 (prove the root cause): profile and reproduction above.

Validation: full suite 1608 passed (3 new tests: hung-loader non-blocking contract, post-load adoption + cache clear, loader-failure fallback); ruff clean; `compileall` clean.

### DIFF
--- a/tests/test_tokens_encoder_loading.py
+++ b/tests/test_tokens_encoder_loading.py
@@ -21,6 +21,7 @@ def _reset_encoder_state(monkeypatch):
     monkeypatch.setattr(tokens_mod, "_encoder", None)
     monkeypatch.setattr(tokens_mod, "_encoder_ready", False)
     monkeypatch.setattr(tokens_mod, "_encoder_thread", None)
+    monkeypatch.setattr(tokens_mod, "_encoder_generation", 0)
     tokens_mod._count_tokens_cached.cache_clear()
     yield
     tokens_mod._count_tokens_cached.cache_clear()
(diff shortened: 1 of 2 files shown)


Tosko4 (changes requested):
I'm requesting changes because this moves canonical engine mutation onto an unsynchronized background thread and introduces reproducible session drift and message loss.

1. **A queued turn can rebind a live clone to a stale session.** `active_engine` is captured at `__init__.py:310`, queued at `:346`, then `_ensure_engine_bound_to_session()` runs later at `:277`. I reproduced this again on the current head: turn A queued, the clone moved to session B, then queued A ran and changed the live binding back from B to A. Foreground work for B can then run against A's lifecycle state.

2. **The single worker does not serialize against foreground writers.** `compress()`, preflight/tool ingest, session-end flushes, and lifecycle callbacks still mutate the same engine concurrently. `_ingest_messages()` reads the cursor at `engine.py:3778` and later overwrites it at `:4130` without a whole-engine ownership boundary. A deterministic race produced duplicate raw rows, then a late background ingest changed a post-compression cursor from 2 back to 6; the next genuine user/assistant suffix wrote zero rows.

3. **The executor has no production lifecycle.** `ThreadPoolExecutor` workers are not daemon threads and Python joins pending work at interpreter exit. The executor created at `__init__.py:226` is never shut down or unregistered, while the first queued job can wait up to 90 seconds. `hook.drain()` only gives tests/callers a manual wait primitive; no host shutdown path calls it, it does not stop the executor, and it does not prevent stale work from mutating a rebound or closed engine.

4. **The stated root cause needs to be proven before changing this boundary.** On this head, `LCMEngine.ingest()` (`engine.py:1170-1204`) only performs raw-message persistence; it does not call leaf summarization. The summarizer call sites are in compaction/condensation. Please identify the exact synchronous stack that produced the 127-second wait, otherwise this may not address the measured path.

Validation on exact head `6185312ee67ae29739768ceecc136477d3bb10d1`:

- Focused maintainer suite: **409 passed**.
- The four updated post-hook tests pass because they call `hook.drain()`.
- Exact head stale-rebind probe still ends on stale session A after the clone has moved to current session B; exact base remains on B.
- Independent cursor-race probe: duplicate persistence plus a late cursor overwrite caused the next genuine two-message suffix to persist zero rows.

The safe shape should keep fast raw persistence separate from expensive background preparation, with generation/ownership checks and atomic publication for any canonical state change. At minimum this needs deterministic stale-generation, concurrent ingest/compress/tool/lifecycle, duplicate-row, late-cursor, shutdown, and backlog-overflow coverage before it is safe to merge.

zab-01:
@Tosko4 reworked per your review — you were right on all four points, and point 4 changed the diagnosis entirely: the stall was tiktoken's first-use BPE download hanging on a restricted-egress host under `count_tokens()`, not summarization. The PR is now a two-file `tokens.py` fix (bounded background encoder load + estimator fallback + adoption cache-clear); the async executor and all its machinery are gone, and your four post-hook tests pass unmodified. Full writeup with the cProfile trace is in the updated description. Re-requesting review.

Tosko4 (changes requested):
I'm still requesting changes because the current head has a cache-adoption race.

`_encoder_loader()` publishes the encoder and clears `_count_tokens_cached`, but an LRU miss already running the fallback can finish after that clear and repopulate the cache with the estimator result. I reproduced this deterministically on exact head `7c8c3fde8ad5a6ba0fd2a980abd9a42173006346` under CPython 3.11 and 3.14: I paused one count inside `_fallback_token_estimate`, let the loader adopt the encoder and clear the cache, then released the fallback. The first and next identical counts both returned `123` instead of the encoder's `7`, despite `_encoder_ready` being true.

That stale estimate can remain until eviction and affect compaction/truncation budgets. Please make the cache generation-aware, or otherwise ensure pre-adoption estimates cannot populate the post-adoption cache. Moving `cache_clear()` under `_encoder_lock` alone would not close an in-flight LRU insertion. Add an event-driven regression for this ordering; concurrent-caller coverage for one loader, bounded first-caller waiting, prompt later callers, and exact post-adoption counts would also pin the intended contract.

Validation on the exact head:
- `python3 -m pytest tests/test_tokens_encoder_loading.py -q -o addopts=`: 3 passed.
- Full release validation: passed, including focused 409 passed, full 1608 passed / 12 xfailed, low-FD 1608 passed / 12 xfailed, benchmark smoke, stress smoke, and release stress.
- Deterministic adoption-race probe: failed on CPython 3.11 and 3.14 with a stale cached estimate after encoder adoption.

I did not reproduce the original restricted-egress network delay itself; this review covers the bounded-loader behavior and cache/concurrency contract.

zab-01:
@Tosko4 confirmed — the adoption-time `cache_clear()` cannot close an in-flight LRU insertion, so a count that entered the estimator before adoption could repopulate the post-clear cache with a stale estimate. Fixed in 9db1b13 by making the cache generation-aware: the memoized count is keyed on `(text, generation)` and the generation bumps under `_encoder_lock` at adoption, so a pre-adoption result can only ever be inserted under the retired key, which post-adoption lookups never touch. The `cache_clear()` is retained purely as memory hygiene for the now-unreachable old-generation entries.

Tests added per your asks, both event-driven:

- `test_adoption_race_stale_estimate_not_served` reproduces your exact ordering — pause a count inside `_fallback_token_estimate`, let the loader adopt and clear, release the fallback — and asserts the first and repeated post-adoption counts for the identical text return the encoder's value. Verified failing against 7c8c3fd before the fix (stale estimate served, as you reproduced).
- `test_concurrent_callers_single_loader_prompt_returns` pins the intended contract: exactly one `_load_encoder()` invocation across concurrent callers, bounded first-caller wait, prompt returns for later callers while the loader hangs, and exact counts after adoption.

Validation on this head: focused file 5 passed (3 consecutive runs), full suite 1610 passed / 12 xfailed, ruff clean.

Re-requesting review.

Tosko4 (approved):
Approved. The generation-keyed cache closes the adoption race from my previous review: an in-flight pre-adoption estimate can only land under the retired generation, while post-adoption lookups use the real encoder. The new event-driven regressions cover that exact ordering and the one-loader concurrent-caller contract.

Validation on exact head `9db1b13df97d33ff2f2146a4eb9606ed3330fb9f`:
- `python3 -m pytest tests/test_tokens_encoder_loading.py -q -o addopts=`: 5 passed; 20 consecutive focused runs also passed.
- Python 3.11 focused run: 5 passed.
- Full release validation: focused 409 passed; full and low-FD suites each 1610 passed / 12 xfailed; benchmark smoke and smoke/release stress gates passed.
- Ruff, compile checks, fast-success exact-count probe, and a real-tiktoken adoption smoke passed.

Material limitation: GitHub reports no CI checks for this head, and I did not reproduce the original cold-cache restricted-egress delay itself. The hung-loader timeout and fallback behavior are covered deterministically.

Issue #375, opened 2026-07-14
roadmap: cache-stable tool-output and full-sweep parity with Lossless Claw


100yenadmin:
## Goal

Bring the highest-value, narrowly portable Lossless Claw behaviors into Hermes-LCM without replacing Hermes-native storage or changing defaults:

1. opt-in provider-facing stubs for durably externalized textual tool results
2. dry-run-first historical sidecar backfill
3. token-bounded protected fresh tails
4. bounded search over externalized payloads
5. bounded atomic threshold full sweeps

We maintain Martian Engineering's Lossless Claw implementation and are porting the proven pattern into Hermes-LCM. The intent is behavioral parity where it materially improves prompt size, recoverability, and cache stability—not a storage-layer or runtime transplant.

## Non-goals

- Replacing Hermes-LCM's existing externalized-payload store.
- Copying Lossless Claw's entire runtime, scheduler, or database schema.
- Enabling new behavior by default in the first PRs.
- Background/deferred compaction; #287 remains the separate async design discussion.
- Duplicating read-tool recovery covered by open PRs #364 and #365.
- Merging, releasing, or changing any live deployment from this tracker.

## Ordered work

- [ ] #374 / #380 — active replay stubbing
- [ ] #377 / #381 — token-bounded fresh tails
- [ ] #376 / #382 — historical externalization backfill
- [ ] #378 / #383 — externalized-payload search
- [ ] #379 / #384 — bounded atomic threshold full sweeps (stacked on #381)

## Acceptance criteria

- Every behavior is default-off or backward-compatible by default.
- Raw message rows and existing DAG lineage remain recoverable.
- Tool-call/result pairing and structured media behavior remain provider-valid.
- Operator mutations are dry-run-first, idempotent, and reversible where safe.
- Full sweeps are bounded by iterations and wall clock and publish one assembled context per invocation.
- Each child issue has a focused PR, tests, docs, and honest validation evidence.
- GitHub Actions is the canonical Python-version validation surface.

## Validation gates

Each PR follows the repository template and, where applicable:

- focused pytest for the changed behavior
- `ruff check .`
- `pytest -q`
- low-file-descriptor pytest
- `python -m compileall -q .`
- script and shell syntax checks
- benchmark/stress smoke
- `git diff --check`
- current-head CI and review-thread readback

## Evidence and provenance

[...]

PR #380, opened 2026-07-14
feat: stub externalized tool results in active replay

### LINKED ISSUE #374: feat: opt-in active-replay stubbing for externalized textual tool results
# feat: opt-in active-replay stubbing for externalized textual tool results

## Problem

Hermes-LCM already externalizes oversized payloads before SQLite/FTS storage (#139 / #148), and it returns compact active stubs for whole-message generic `raw_payload` externalization (#214). However, ordinary oversized textual tool results remain raw in the provider-facing active replay even after a durable externalized payload and compact stored reference exist.

Current `_ingest_messages()` creates provider replay messages separately from `protect_messages_for_ingest()`. It substitutes the protected content into active replay only for the existing generic `raw_payload` active-stub case. `_assemble_context()` then appends the selected tail and performs tool-pair cleanup, but no textual tool-result stub substitution. This means large tool output can continue to consume the active context, force earlier compaction/eviction, and churn the stable prompt prefix even though lossless recovery is already available through `lcm_expand` / `lcm_describe`.

This is deliberately narrower than #139. Storage externalization is implemented. The missing surface is provider-facing replay of already-durable textual tool results.

## Prior art

Martian-Engineering/lossless-claw implemented this in several layers:

- [#628](https://github.com/Martian-Engineering/lossless-claw/pull/628): replace externalized evictable historical tool-result rows with compact references while preserving tool-call/result pairing, content shape, fresh tail, and drilldown.
- [#862](https://github.com/Martian-Engineering/lossless-claw/pull/862): run the large-tool interceptor on live messages before normal/degraded assembly so current-turn/fallback paths cannot bypass stubbing.
- [#946](https://github.com/Martian-Engineering/lossless-claw/pull/946): do not re-externalize `lcm_describe` drilldown output.
- [#950](https://github.com/Martian-Engineering/lossless-claw/pull/950): recover full live `read` output before externalizing a host-truncated fragment.

Hermes-LCM should reuse its existing externalized payload store and tools rather than port Lossless Claw's storage layer.

## Proposed direction

Add a separate opt-in active-replay policy for oversized textual tool results whose payload has already been externalized durably.

Suggested constraints:

- Default off; no behavior change for existing users.
- Reuse the durable externalized ref produced by ingest protection; do not create a second payload.
- Fail open: if persistence or ref validation fails, retain the original tool result.
- Preserve role, `tool_call_id`, assistant `tool_calls`, ordering, and structured provider content shape.
- Preserve the intentional #226 contract for structured image/media and tool-result media unless explicitly configured otherwise.
- Exempt `lcm_expand` / `lcm_describe` recovery output from immediate recursive re-stubbing, or provide an explicit bounded expansion path.
- Apply consistently to normal active assembly and overflow/degraded fallback paths.
- Make the threshold token-aware if practical; at minimum document its relationship to the current character-based externalization threshold.
- Keep historical backfill/migration a separate, dry-run-first operator surface; future ingest should need no new schema.

## Acceptance criteria

- An oversized textual tool result is stored once as a recoverable externalized payload and appears as a compact provider-facing reference when the new option is enabled.
- The option disabled is byte/structure compatible with current active replay.
- Tool-call/result pairing remains provider-valid.
- Multi-block content remains provider-valid.
- Structured media behavior from #226 remains unchanged.
- Normal assembly and overflow/degraded fallback cannot bypass enabled stubbing.
- `lcm_expand` / `lcm_describe` can recover the full original content, and their recovery response does not enter an accidental stub loop.
- Storage failure leaves the original result inline.
- Focused benchmarks report active prompt tokens saved and assembly latency.

## Non-goals

- Replacing existing Hermes externalization storage.
- Changing compaction policy or implementing sweep compaction in this issue.
- Automatically rewriting existing databases without an explicit backup/dry-run/apply workflow.

100yenadmin:
## Summary

- add **opt-in, token-aware stubbing of oversized textual tool results in provider-visible active
  replay** — externalize first, fail open if durable storage can't be confirmed, replace inline
  content with a Hermes-native recoverable reference
- preserve tool-call pairing, structured text shapes, media payloads, fresh-tail behavior, and
  `lcm_describe`/`lcm_expand` recovery (recovery results are never re-stubbed)
- during a compression-boundary cooldown, a **durable replay cleanup still publishes — with zero
  summarizer spend** (cleanup-only pass; proven by a test that mocks the summary LLM and asserts it
  is never called)
- plain (no-replay-diff) preflight ordering is **unchanged from upstream**: an earlier draft reordered
  the cooldown/overflow checks there; that reorder is reverted out of this PR (happy to propose it
  separately with its own test if the semantics interest you)
- deterministic synthetic benchmark: **79.78% fewer provider-visible prompt tokens** (300,288 →
  60,709), 8/8 eligible payloads recovered equal after normalization

Tracker: #375
Closes #374

## Why

Hermes-LCM already externalizes large outputs durably, but provider-visible replay can keep carrying
the full text until compaction reaches it — paying the token cost every turn. This policy replaces
qualifying active tool results with recoverable references at ingest and historical-assembly time,
leaving raw SQLite history and DAG lineage unchanged. Disabled by default
(`LCM_LARGE_OUTPUT_ACTIVE_REPLAY_STUBBING_ENABLED=false`); media-bearing results stay inline;
storage-confirmation failure keeps the original content.

### Where this hooks into the pipeline

```mermaid
flowchart LR
    I[Ingest] --> C[Compact] --> D[Condense] --> ES[Escalate] --> A[Assemble] --> R[Retrieve]
    S[stub substitution<br/>opt-in] --> X[(externalized<br/>payload sidecars)]
    I -->|oversized textual<br/>tool result| S
    A -->|recoverable ref<br/>in replay| S
    classDef hot fill:#b45309,stroke:#7c2d12,color:#fff
    class I,A,S hot
```

`lcm_describe` / `lcm_expand` recover the original content on demand; raw history is never rewritten.

## Validation

- [x] focused suite `tests/test_active_tool_stubbing.py` — **25 passed**, including four **flag-off
      regression tests**: cooldown still blocks plain threshold/overflow preflight; a durable replay
      cleanup publishes during cooldown; that publish makes **zero** summary-LLM calls
      (`assert_not_called` on the mocked summarizer); non-cleanup replay diffs stay blocked
- [x] full `pytest -q` — **1631 passed, 1 skipped, 12 xfailed**; 3 failures are the known macOS
      baseline (1 tokenizer-dependent + 2 `/var` vs `/private/var`), reproduced identically on the
      untouched upstream clone at `31675c6` — Linux CI is the canonical gate
- [x] `ruff check .` — pass
- [x] `scripts/validate_release.sh --full --keep-going` — diff/compile/shell gates, focused pytest
      (409), benchmark smoke, both stress tiers all pass; the two full-suite variants carry only the
      same three macOS baseline assertions
- [x] deterministic benchmark repeated twice — 300,288 → 60,709 both runs; 8/8 recovery equality
- [x] rebased onto current `main` (`31675c6`), no conflicts

(Fork workflow runs need maintainer approval — validation output above is locally reproducible until
then.)

## Notes

- Opt-in and backward compatible: no database rewrite, no transcript GC required, no live profile
  touched by tests or benchmarks.
- Replay-branch preflight ordering is deliberate and now documented in-code: a boundary skip cools
  down summary-*producing* work, but must not stop the host from adopting a cleanup that ingest has
  already made durable — that pass is deterministic and spends nothing.
- Overlaps: #370 touches the same `compress()`/assembly region — happy to rebase after it lands (the
  rebase cost is ours). #367 rewrites the config field table our two new entries sit near — same
  offer. Sibling #381 shares engine/compaction surface; #382/#383 don't overlap this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,6 +335,8 @@ Most installs only need `plugins.enabled` and `context.engine: lcm`.
 | `LCM_SENSITIVE_PATTERNS` | `api_key,bearer_token,password_assignment,private_key` | Comma-separated named sensitive pattern catalog entries to apply when redaction is enabled |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_ENABLED` | `false` | Store oversized ingest payloads, including tool results, media blocks, and generic raw content, in plugin-managed JSON files |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_THRESHOLD_CHARS` | `12000` | Externalization threshold for normalized payload text |
+| `LCM_LARGE_OUTPUT_ACTIVE_REPLAY_STUBBING_ENABLED` | `false` | Replace token-heavy textual tool results with recoverable externalized refs in active replay; current-turn ingest is immediate and historical assembly respects the protected fresh tail; requires large-output externalization |
+| `LCM_LARGE_OUTPUT_ACTIVE_REPLAY_STUB_THRESHOLD_TOKENS` | `25000` | Token-aware threshold for active-replay tool-result stubbing |
 | `LCM_LARGE_OUTPUT_TRANSCRIPT_GC_ENABLED` | `false` | Rewrite already-externalized summarized tool rows to compact placeholders |
 | `LCM_DOCTOR_CLEAN_APPLY_ENABLED` | `false` | Permit destructive `/lcm doctor clean apply` in trusted operator contexts |
 | `LCM_EMPTY_LIFECYCLE_GC_ENABLED` | `true` | Master toggle for automatic pruning of lifecycle rows for sessions that never ingested any messages or summary nodes |
@@ -512,6 +514,21 @@ Externalization for ordinary large tool output is opt-in. When enabled,
 oversized tool results are written to plugin-managed JSON files and referenced
 from summaries. They remain inspectable through
 `lcm_describe(externalized_ref=...)` and `lcm_expand(externalized_ref=...)`.
+
+Active-replay stubbing is a second, independently opt-in replay policy. When
+both externalization and active-replay stubbing are enabled, newly ingested
+textual tool results above the token threshold are durably externalized and
+replaced immediately in provider-visible replay, including results in the
+protected fresh tail. This lets a stub-only replay change converge even when no
+leaf is eligible for compaction. A historical assembly pass applies the same
+policy to older tool results before budgeting, while respecting the protected
+fresh tail. Tool-call ids and compatible structured text block types/keys are
+retained; raw SQLite rows and DAG lineage are not rewritten by the historical
+pass. Structured image/media results remain inline, preserving the provider
+replay contract established by Hermes-LCM PR #226. If durable externalization
+cannot be confirmed, replay keeps the original payload inline. Results from
+`lcm_describe` and `lcm_expand` also remain inline so recovery does not
+recursively produce another ref.
 
 The storage-boundary payload guard is separate from that opt-in. LCM always
 scans messages at the store boundary before writing `messages.content` or
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -9,6 +9,22 @@ The benchmark harness is offline by default:
 - no live Hermes config mutation
 - writes isolated to the requested output directory
 
+## Active tool-result stubbing benchmark
+
+The focused active-replay benchmark builds ten deterministic synthetic tool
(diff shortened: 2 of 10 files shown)


Tosko4 (approved):
Approved at `ec01105` after the current-main refresh and exact-head validation (315 passed).

stephenschoettler:

PR #382, opened 2026-07-14
[v1·externalize 3/5] dry-run historical tool-output externalization backfill

### LINKED ISSUE #376: feat: add dry-run historical tool-output externalization backfill
## Problem

Forward ingest externalization does not pre-create durable sidecars for textual tool-result rows already present in `lcm.db`. Active replay can externalize eligible rows lazily, but a migrated or long-running installation may pay that discovery/write cost during context assembly and may retain a large unprepared historical corpus indefinitely.

Parent tracker: #375

## User or operator impact

Operators with existing LCM history need a bounded, inspectable way to prepare old textual tool outputs for active-replay stubbing without rewriting raw rows, changing DAG lineage, or pointing an ad-hoc migration at a live database.

## Proposed direction

Add a standalone operator script that:

- scans tool-result rows from an explicit target database
- uses the Hermes-native externalized-payload format and deduplication helpers
- defaults to dry run and writes only with `--apply`
- accepts optional session and token-threshold filters
- never rewrites `messages.content`, `messages.tool_calls`, summary nodes, or lifecycle state
- emits a scrubbed manifest containing created refs, content hashes, counts, and token estimates, never raw content
- is idempotent across repeated runs
- supports manifest-based rollback only when file digests still match and no message or summary node references the sidecar
- refuses unsafe rollback rather than deleting a recoverability dependency

This is a prewarm/operator surface for #374, not a new storage layer.

## Alternatives and tradeoffs

- Lazy-only externalization avoids an operator command but moves sidecar creation onto the provider-facing assembly path.
- Rewriting historical message rows would make refs explicit but is materially riskier and unnecessary because Hermes can find sidecars by content/session/tool identity.
- Copying Lossless Claw's blob migration would duplicate storage Hermes already owns.

## Backward compatibility

No schema rewrite and no default runtime behavior change. Dry run is the default. Raw rows remain authoritative and recoverable.

## Validation or evidence

A local migrated profile contained hundreds of eligible historical tool rows; no private rows, session IDs, or raw content will be included in public evidence. Tests will cover dry run, apply, idempotency, manifest safety, referenced-sidecar rollback refusal, missing/corrupt files, and no-row-mutation invariants.

## Implementation status

PR incoming.

## Additional notes

We maintain Martian Engineering's Lossless Claw implementation and are porting the proven operator pattern into Hermes-LCM while retaining Hermes-native storage.

100yenadmin:
## Summary

- add an **offline, dry-run-first** backfill script that externalizes oversized historical tool
  outputs into sidecar files — `scripts/backfill_externalized_tool_outputs.py`
- pre-created sidecars are **digest-matched by the engine's existing externalization lookup** during
  later ingest/compaction, so history gets the same treatment as new messages with **zero engine
  changes**
- guarded rollback mode: manifest-scoped, refuses still-referenced sidecars, and reports **per-file
  partial failures** (`succeeded`/`failed`/`skipped` counts + `failed_paths`), attempting every
  remaining file even after an error and exiting non-zero if any deletion failed
- deterministic tests: **13 focused** (was 6), now covering the already-externalized dry-run counter,
  invalid/missing/symlink rollback refs, schema/non-applied manifest guards, aggregate reporting,
  and partial-failure continuation

Tracker: #375
Closes #376

## Why

Hermes-LCM externalizes oversized tool outputs at ingest going forward, but databases that predate
that feature (or ran with it disabled) still carry large historical rows inline, inflating
provider-visible replay until compaction reaches them. This script lets an operator externalize that
history **offline and reversibly**: dry-run first, idempotent, sidecars only — the engine adopts them
through its normal digest lookup, no schema or engine behavior change.

### Where this hooks into the pipeline

```mermaid
flowchart LR
    I[Ingest] --> C[Compact] --> D[Condense] --> ES[Escalate] --> A[Assemble] --> R[Retrieve]
    B[one-shot backfill<br/>dry-run first] --> X[(externalized<br/>payload sidecars)]
    X -.->|digest match at<br/>ingest/compaction| I
    classDef hot fill:#b45309,stroke:#7c2d12,color:#fff
    class B hot
```

The script never touches the pipeline at runtime — it only pre-creates the sidecar files the
existing `externalize.py` digest lookup already knows how to adopt.

## Validation

- [x] `pytest -q tests/test_historical_externalization_backfill.py` — **13 passed**
- [x] full `pytest -q` — **1620 passed, 1 skipped, 12 xfailed**; 2 failures are the known macOS
      `/var` vs `/private/var` path assertions, reproduced identically on untouched upstream `main`
      at `31675c6` (same two tests) — Linux CI is the canonical gate
- [x] `ruff check .` — pass
- [x] `python -m compileall` + script `py_compile` + shell syntax — pass
- [x] `scripts/validate_release.sh --full --keep-going` — all gates pass (focused pytest 409 passed;
      benchmark, stress smoke, and release stress tiers pass; the two full-suite variants carry only
      the same two macOS baseline assertions)
- [x] rebased onto current `main` (`31675c6`), no conflicts, linear history

(Fork workflow runs need maintainer approval — the validation output above is locally reproducible
until then.)

## Notes

- Opt-in by construction: an operator-invoked script; nothing runs automatically.
- No database rewrite: message rows are untouched; the script writes sidecar files + a manifest.
- Rollback is deliberately conservative: refuses sidecars still referenced by the DB, skips
  digest mismatches and symlinks, and reports rather than hides partial failures.
- No file overlap with sibling PRs #380/#381/#384 or with open #370/#367 (script + its tests only,
  plus a one-line validator hook).


Tosko4 (approved):
Approved at `bda9142` after the current-main refresh and exact-head validation (13 passed).

Tosko4 (changes requested):
Requesting changes on `b57b00f` because the destructive backfill/rollback contract still has release blockers.

- The manifest is not a durable ownership journal (`scripts/backfill_externalized_tool_outputs.py:194-309`). Each run creates a new manifest ID and publishes only once, after all sidecar writes. Reusing the same manifest path replaces the earlier rollback record with an empty manifest; a manifest publication failure strands created sidecars; apply failures still return exit code 0. The manifest also does not bind the database or storage root used during apply.
- Rollback does not use the canonical ref scanner. `_ref_is_referenced()` checks `messages.content` and summaries only (`scripts/backfill_externalized_tool_outputs.py:313-330`). I reproduced a rollback deleting a sidecar whose live ref existed only inside nested `messages.tool_calls`.
- Schema and filesystem checks fail closed too late or not at all. A database marked schema version 999 was scanned successfully. `_write_manifest()` resolves the path before checking `is_symlink()` (`scripts/backfill_externalized_tool_outputs.py:156-170`), and I reproduced it overwriting the symlink target. Rollback also separates pathname checks from the later `unlink()`, leaving leaf/parent replacement races.
- The privacy/operator contract needs to be explicit and tested. Historical content is copied into a new sidecar retention surface without applying the currently enabled sensitive-pattern policy, while manifest refs can include the sanitized tool-call stub despite the operator guide saying tool-call IDs are omitted. The new targeted test file is also missing from the default focused release-validation gate.

- `python3 -m pytest -q tests/test_historical_externalization_backfill.py -o addopts=`: 16 passed.
- `scripts/validate_release.sh --full --keep-going` in an isolated home: all gates passed, including 409 focused tests, 1661 full tests, 1661 low-FD tests, and release stress with zero failures.
- GitHub CI: all six current checks green.
- Temporary database/filesystem probes reproduced the ownership-loss, nested-ref deletion, future-schema, symlink-target, and successful-exit-on-failure cases above.

I did not run apply or rollback against a live Hermes profile. Please make ownership crash-recoverable and target-bound, reuse the canonical schema/ref safety paths, use no-follow descriptor-bound filesystem operations, define the redaction/retention policy, and add the corresponding adversarial regressions before the next review.

Tosko4 (approved):
Approving the current head after the requested durability and privacy follow-ups. The manifest/rollback safety, canonical ref handling, descriptor-bound filesystem work, and sensitive-pattern policy are now covered by the subsequent hardening commits. Current CI is green and there are no unresolved review threads.

100yenadmin:
Addressed all four blockers. Head is now `0e7a293`. Blocker → commit(s) → regression tests:

**1. Durable ownership journal** — `b57b00f`, `009cae0`, `6f5c924`
Write-ahead journal: each sidecar is journaled to `pending_items` and fsync'd before it is created; the manifest is crash-recoverable and reuse of the same path never clobbers a prior applied journal. The manifest binds db path + storage root + schema version (verified at rollback) and apply now exits non-zero on any failure. Publish/exchange boundaries no longer replace or strand a regular file.
- `test_same_path_apply_rerun_preserves_rollback_manifest_ownership`
- `test_same_path_apply_appends_new_owned_sidecars_without_losing_old_ones`
- `test_interrupted_manifest_publication_recovers_pending_owned_sidecar`
- `test_same_manifest_cannot_be_reused_for_another_database_or_storage_root`
- `test_backfill_apply_write_failure_returns_nonzero_with_failed_path`
- `test_manifest_publication_does_not_replace_regular_file_that_appears_at_publish_boundary`
- `test_manifest_update_restores_regular_file_swapped_in_at_exchange_boundary`

**2. Rollback uses the canonical ref scanner** — `009cae0`
`_collect_referenced_refs` now enumerates refs from `messages.content`, `messages.tool_calls`, and summaries through the plugin's canonical `extract_all_externalized_payload_refs` path, including nested `tool_calls`.
- `test_rollback_refuses_sidecar_referenced_in_nested_tool_calls`
- `test_rollback_refuses_referenced_sidecar`

**3. Fail closed, early, race-free** — `009cae0`, `6f5c924`
Schema-version gate up front via `refuse_schema_version_too_new` (schema 999 now refused before any scan/write). Manifest and sidecar reads/writes use `O_NOFOLLOW` descriptor-bound opens; rollback deletes via an `os.open(dir_fd)` + rename-to-quarantine + `unlinkat` pattern with a final identity recheck, closing the leaf/parent replacement race.
- `test_backfill_refuses_future_database_schema_before_scan_or_sidecar_write`
- `test_manifest_symlink_is_refused_without_touching_target`
- `test_rollback_rejects_symlink_manifest_without_reading_target`
- `test_rollback_does_not_delete_regular_file_swapped_after_validation`
- `test_rollback_fails_closed_when_quarantine_is_replaced_after_identity_check`

**4. Explicit privacy/retention contract** — `0e7a293`
The currently enabled `LCM_SENSITIVE_PATTERNS` policy is now applied to tool-result content before each sidecar is written, exactly as live ingest does, so no un-redacted secret reaches the new retention surface; every manifest digest, ref, and provenance proof is derived from the redacted content actually stored. The manifest omits raw session/tool-call ids and its refs carry a `historical-backfill` marker rather than a tool-call stub; the active redaction policy is bound and reuse under a changed policy is refused. Retention story documented in the operator guide, and the targeted test file is added to `scripts/validate_release.sh` focused gate.
- `test_apply_redacts_sensitive_content_before_writing_sidecar`
- `test_manifest_never_leaks_tool_call_or_session_ids`
- `test_backfill_refuses_manifest_reuse_under_changed_redaction_policy`

Validation on this head (isolated `HERMES_HOME`, CI agent-stub, single process): `tests/test_historical_externalization_backfill.py` 33 passed; full suite 15 failed / 1660 passed, the 15 failures identical to the pre-existing env baseline (unrelated `test_auto_focus_topic`, `test_path_containment`, `test_path_security`); `ruff check` clean on touched files.

100yenadmin:
@Tosko4 re-requesting your review — the four blockers from your `b57b00f` changes-requested review are addressed on head `0e7a293` (mapping and regression tests in the comment above). The formal re-review request API is blocked by my fork-account permissions on this repo, hence this mention.

Tosko4:
> @Tosko4 re-requesting your review — the four blockers from your `b57b00f` changes-requested review are addressed on head `0e7a293` (mapping and regression tests in the comment above). The formal re-review request API is blocked by my fork-account permissions on this repo, hence this mention.

It's approved

Tosko4 (approved):
@stephenschoettler Re-approving the current exact head `0e7a293`. It is approved, mergeable/clean, all six checks are green, and there are no unresolved review threads. My later “It’s approved” comment only confirmed this approval; it did not invalidate it. Please merge this PR.

stephenschoettler:
Thanks, this was approved, checks passed, and it has now been merged as `77d75fa5`.

Issue #386, opened 2026-07-15
Design: pluggable embeddings + semantic/hybrid retrieval — provider abstraction, BLOB vector store, staged train


100yenadmin:
Title: Design: pluggable embeddings + semantic/hybrid retrieval — provider abstraction, BLOB vector
store, staged train

## What

Add **opt-in semantic and hybrid retrieval** to hermes-lcm: a pluggable embedding provider
(`voyage` HTTP / `ollama` / `fastembed` local / `none`), a dependency-free vector store (float32
BLOBs in regular tables), a dry-run-first backfill command, and `semantic`/`hybrid` modes on the
existing retrieval tools — with recall proven by a committed eval script, and a docs guide covering
**free** (Voyage free tier) and **local no-signup** (fastembed/Ollama) setups.

We maintain Martian Engineering's Lossless Claw and are porting the proven pattern into Hermes-LCM
(the largest remaining piece of #375's parity direction; design issue first per CONTRIBUTING — the
stacked PR train follows, referencing this issue).

## Why

`lcm_grep` is FTS/BM25: exact-term recall is excellent, paraphrase recall is not ("have we discussed
auth token rotation?" misses "credential refresh"). In our Lossless Claw deployment, hybrid retrieval
(FTS ∪ KNN → rerank) measurably lifted recall on a stratified query set. LCM's summary corpus is the
ideal target: thousands of dense leaf summaries, not millions of chunks.

## Design decisions (and the data behind them)

**1. Vector math: brute-force cosine over float32 BLOBs in a regular table is the DEFAULT.**
No new hard dependency, no SQLite extension. `sqlite-vec` is an optional accelerator only
(`LCM_SQLITE_VEC_ENABLED=false`): it is pre-1.0/alpha and — decisively — **cannot load on macOS
stock-Python** (Apple compiles SQLite with `SQLITE_OMIT_LOAD_EXTENSION`), which many Hermes users run.
Benchmark on the corpus sizes that matter:

**Benchmark** (median ms, top-50 cosine, unit-normalized float32 BLOBs in SQLite; 1 warmup + 5 measured; Homebrew Python 3.12, numpy 2.5.1, sqlite-vec 0.1.9; full method + raw data ship with the train):

*D = 512*
| N | pure-python | bounded-scan (2k) | numpy cold | numpy warm | sqlite-vec |
|---|---|---|---|---|---|
| 5,000 | 47 | 21 | 5.8 | 0.07 | 2.9 |
[...]

Issue #385, opened 2026-07-15
Design: temporal rollups + natural-time retrieval (lcm_recent) — day/week/month memory over LCM history


100yenadmin:
## What

Add **calendar-time memory** to hermes-lcm: day/week/month rollup summaries built from existing leaf/
condensed summaries, plus a natural-time retrieval tool (`lcm_recent`) that answers "what did we work
on yesterday / last week / on 2026-07-01" without content-keyword guessing.

We maintain Martian Engineering's Lossless Claw and are porting the proven pattern into Hermes-LCM
(part of #375's parity direction; this is the design issue first, per CONTRIBUTING — the staged PR
train follows, stacked and referencing this issue).

## Why

Today retrieval is content-indexed (`lcm_grep`, `lcm_describe`, `lcm_expand`): excellent for "find the
linker error", blind for "summarize yesterday". Agents and operators both reach for time-shaped recall
constantly, and the DAG already contains everything needed — it just has no calendar aggregation. In
our Lossless Claw deployment the equivalent feature became one of the most-used recall surfaces.

## Design

**Schema** (new `SCHEMA_VERSION` step, idempotent `ensure_*` helpers per the `db_bootstrap.py` pattern):

- `lcm_rollups` — `rollup_id`, `period_kind` (`day|week|month`), `period_start`, scope
  (per-conversation for dailies; global aggregates for week/month), `summary`, `token_count`,
  `status` (`building|ready|stale|failed`), `built_at`, `source_fingerprint`
- `lcm_rollup_sources` — rollup → contributing summary-node ids (provenance + staleness detection)
- a small state row for last-build cursors per period kind

**Builder** — daily rollups synthesized from that day's leaf/condensed summaries (target ~5k tokens,
hard cap ~15k), then week/month aggregates built **from the dailies** (~20k cap). New ingest marks the
affected day `stale`; rebuilds happen inside the **existing synchronous maintenance seams** — this
proposal adds **no background thread** (deliberately consistent with the staged posture in
`docs/async-background-compaction.md`; a worker loop, if ever wanted, is its own discussion).
Synthesis rides the existing summarize-with-escalation auxiliary path, so the circuit-breaker and
spend-guard already govern it.

**Tool surface** — `lcm_recent`: parses `today | yesterday | 7d | week | month | date:YYYY-MM-DD`
(+ relative windows), returns ready rollups; when a rollup is absent/stale/disabled it **falls back to
[...]

PR #387, opened 2026-07-15
[v1·temporal 1/4] rollup storage substrate (schema + RollupStore)

100yenadmin:
## Summary

- storage substrate for **temporal rollups** (design: #385 — train PR 1 of 4): `lcm_rollups` /
  `lcm_rollup_sources` / `lcm_rollup_state` tables via a new `SCHEMA_VERSION` step, plus a
  transactional `RollupStore`
- day→week/month **staleness cascade**, windowed ready-rollup reads, build cursors, app-level source
  purge integrity (matches the repo's no-SQL-FK convention)
- **no behavior change**: nothing constructs or calls `RollupStore` yet; the new
  `LCM_TEMPORAL_ROLLUPS_ENABLED` flag defaults off and is inert in this PR
- 12 focused tests: migration idempotency, v5→v6 upgrade, newer-version refusal, CRUD/rebuild,
  staleness cascade, uniqueness, cursors, purge, config defaults/overrides

Design anchor: #385 (temporal rollups + `lcm_recent`)
Part of #375

## Why

Time-shaped recall ("what did we do yesterday / last week") needs calendar aggregation the DAG doesn't
have. Per #385, the train lands in the smallest reviewable units — this first PR is schema + store
only, so the migration and data model can be reviewed in isolation before any behavior exists.

```mermaid
flowchart LR
    RS[(lcm_rollups /<br/>sources / state)]
    S1[PR-1 · this:<br/>schema + RollupStore] --> RS
    S2[PR-2: builder +<br/>maintenance wiring] -.-> RS
    S3[PR-3: lcm_recent] -.-> RS
    S4[PR-4: introspection] -.-> RS
```

## Validation

- [x] `pytest -q tests/test_rollup_store.py` — **12 passed**
- [x] full `pytest -q` — **1618 passed, 1 skipped, 12 xfailed**; 3 failures are the known macOS
      baseline, reproduced identically on untouched upstream `main` at `31675c6`
- [x] `ruff check .` — pass
- [x] `scripts/validate_release.sh --full --keep-going` — all gates pass (focused 409; benchmark +
      both stress tiers; the two full-suite variants carry only the same three macOS baselines —
      low-fd gate completed normally this run)
- [x] five stale `SCHEMA_VERSION == 5` literals in `tests/test_lcm_core.py` now use the shared
      constant (version tracking only, no behavioral test change)

(Fork workflow runs need maintainer approval — validation is locally reproducible until then.)

## Notes

- Opt-in flag included here only so the config surface is stable across the train; it gates nothing
  until PR-2.
- `lcm_rollup_sources` uses app-level integrity (transactional purge/replace in `RollupStore`),
  mirroring the existing summary-source convention rather than SQL FKs.
- Heads-up: our embeddings design (#386) will also need a schema step; whichever train lands second
  rebases its version number — the migration steps are written to make that a two-line change.

### Schema-migration disclosure (adversarial-review follow-up)

To be precise about the "no behavior change when disabled" claim: like every schema-backed feature in
this repo, the migration is **unconditional** — opening an `lcm.db` with this build bumps
`SCHEMA_VERSION` to the next value and creates the new tables even with the feature flag off. Those
tables stay **empty** and nothing reads or writes them until the feature is enabled, so there is no
*behavioral* change — but the on-disk schema does advance. One consequence worth stating: a user who
tries this build and then rolls back to an older release will hit `refuse_schema_version_too_new` (the
older build won't open a newer-schema DB). That is inherent to any additive schema step here, not
specific to this PR; flagged explicitly so it's a conscious choice rather than a surprise.


100yenadmin:
Fix round (adversarial self-review) — pushed a `fix(rollups): store-level correctness` commit:
- **mark_stale now UPSERTs**: a never-built period had no row, so `mark_stale_for_day` (UPDATE-only) touched nothing and automatic maintenance never built it. It now inserts a `stale` row for missing day/week/month so the first rollup actually gets built.
- **rebuild preserves the last good summary**: `upsert_building` no longer nulls an existing summary before the LLM runs — a failed rebuild keeps the previous serveable summary instead of blanking it.
- added `idx_lcm_rollups_pending` (partial index) so the maintenance scan doesn't table-scan all rollup history. 13 store tests.

Tosko4 (changes requested):
Thanks for the cleanup here. I re-reviewed this exact head as the storage substrate for the temporal train and reran the focused store tests plus downgrade and lifecycle probes. The new stale-row creation fixes the initial scheduling hole: invalidating a day now seeds day/week/month work, which is a real improvement.

I’m still requesting changes at this architecture gate because the durable publication contract is not safe yet:

- `db_bootstrap.py:744-749` still creates the rollup tables and advances the core schema to v6 during ordinary startup, even with temporal rollups disabled. A committed v5 database opened by this head becomes v6, and exact base code then refuses it with `SchemaVersionTooNewError`. This also collides with the embeddings train independently claiming core v6.
- Scope identity is still an opaque `scope TEXT`, while cursor state is global per period kind. The later train uses session IDs as “conversation” identity, which does not survive rotation or encode timezone/profile policy.
- `upsert_building()` / `mark_ready()` still have no generation or lease token. Invalidation while a row is `building` does not supersede that build, and the old builder can still publish `ready` afterward.
- The last-known-good bytes survive a rebuild attempt now, but reads only expose `status='ready'`; they disappear during `building` and remain unavailable after `failed`.
- Purging source-backed rollups deletes the row without restoring rebuild debt or rewinding the cursor.

Focused validation: `tests/test_rollup_store.py` passed, and the cumulative temporal focused suite passed 62/62. The adversarial publication probe still reproduced `building → invalidation leaves building → late old builder publishes ready`.

Smallest safe path: keep this as feature-owned derived state rather than an unconditional core-schema upgrade, define structured scope identity, and add generation/CAS publication before the builder train depends on it.

100yenadmin:
@Tosko4 addressed — pushed `fix(rollups): lazy feature tables + generation/lease + per-scope cursor`:
- **Schema:** core `SCHEMA_VERSION` stays **5**; the v6 numeric step is gone. Rollup tables are created lazily+idempotently from RollupStore init on the enabled path, recorded via a named `temporal_rollups_v1` migration step. Disabled install → no tables, version stays 5, base code opens it; no cross-train v6 collision.
- **Build race:** added a `generation` column. Invalidation increments it; `upsert_building` returns a build token capturing the generation; `mark_ready(token)` is a compare-and-set that publishes **only if generation is unchanged** (returns False when superseded). `reclaim_stale_building` flips lease-expired `building` rows back to `stale` so a crashed build is retried.
- **Scope identity:** investigated — `conversation_id` exists in `lcm_lifecycle_state` but is not reachable from `summary_nodes` (session-keyed, carry-over reassigns forward), so a family key would need a core-schema column + backfill, which the schema fix above forbids. Kept session-scoped rollups, made cursor state **per-(period_kind, scope)** (no more shared global cursor), and documented the rotation boundary honestly. Happy to revisit a family key if you'd add the column upstream.

Tosko4 (changes requested):
Re-review verdict for exact head `edd9fe80`: **changes still required**.

There is real progress here. I independently verified that the core schema remains v5, the temporal tables are created lazily under `temporal_rollups_v1`, disabled installs remain inert, and temporal/embedding initialization works in either order (including 20 concurrent first-use probes).

Two publication-safety blockers remain in this substrate:

- Two concurrent `upsert_building()` calls for the same period receive the same `(rollup_id, generation)`. Both can publish; in my probe the older late builder overwrote the newer ready summary and source list.
- Terminal updates are not consistently generation-guarded. A late `mark_failed()` changed a newer ready row to `failed`, and an old `record_incomplete_aggregate()` changed a newer ready aggregate back to `stale`.

These are independent reproductions against the exact store blob, not inferences from automated review. The remaining safe path is one exclusive build owner/token plus generation-CAS for every terminal outcome, not only `mark_ready()`.

100yenadmin:
Addressed the automated review's inline findings on this head (pushed `fix(rollups): complete the generation/lease model + perf indexes + purge re-stale`):
- **The generation model is now complete, not just on `mark_ready`**: the build **claim** itself issues a fresh generation (racing claimants can't share a token — only the latest can publish), `mark_failed` is CAS-guarded like `mark_ready` (a superseded builder's late exception can't flip a newer row to failed), and the aggregate-deferral write is generation-guarded too.
- **No-source stale jobs resolve** instead of lingering and consuming per-pass budget.
- **Purge no longer orphans windows past the cursor**: deleting a built rollup re-seeds a stale row so the window rebuilds.
- **Indexes**: `lcm_rollup_sources(node_id)` for purge lookups; `scope` included in the ready and pending indexes so per-conversation reads and the stale scan don't walk every scope.

Tosko4 (changes requested):
Targeted re-review of exact head `7843bd2d`: the earlier ownership/publication blockers are fixed. Competing claims now receive distinct generations, late ready/failure outcomes are generation-guarded, and the focused race tests pass.

One #387-owned migration blocker remains under the architecture decision on #375. Feature initialization still accepts an existing same-name `lcm_rollups` table without validating its full shape, publishes `temporal_rollups_v1`, and only fails when the store is used. I reproduced this with a table that had the indexed identity/status columns but omitted `built_at`: `RollupStore` initialized and marked the migration complete, then the first build failed with `table lcm_rollups has no column named built_at`.

Please make feature initialization atomic with marker publication, or validate the complete versioned table/index/constraint shape before publishing the marker and provide an explicit repair/refusal path. A malformed or interrupted feature schema must not be recorded as complete.

Tosko4 (changes requested):
Follow-up correction on exact head `7843bd2d`: my earlier statement that late terminal outcomes were fully guarded was too broad. Competing claims now receive distinct generations, but two store-level blockers remain in addition to the feature-migration issue.

- After `mark_ready()` succeeds, `mark_failed()` with the same token still updates the row because its guarded update checks only `rollup_id + generation`, not `status='building'`. I reproduced `ready_write=true`, then `late_failed_write=true`; the final row was `failed` while retaining the published summary. A downstream exception after publication can therefore hide a valid result.
- `upsert_building()` deletes the last-known-good `lcm_rollup_sources` rows before replacement publication. In a rebuild/purge race, the purge found zero affected rollups and the in-flight build then published deleted-node content as `ready`.

Please require `status='building'` for every generation-guarded terminal transition and preserve queryable source lineage until replacement publication commits atomically. The malformed/partial feature-schema blocker from the previous review also remains open.

100yenadmin:
Addressed (pushed `fix(rollups): status-guard terminal transitions + preserve rebuild lineage + verify feature schema`):
- **Terminal transitions now require `status='building'`**, not just `(rollup_id, generation)`: `mark_ready`, `mark_failed`, and the no-source resolve all include `AND status='building'`. Your repro is now a test — after `mark_ready` succeeds, a late `mark_failed` with the same token is a no-op and the published summary stands.
- **Rebuild lineage preserved atomically**: `upsert_building` no longer deletes `lcm_rollup_sources` at claim time — the prior lineage stays queryable (so a concurrent purge still sees the rollup and re-stales it), and the sources-replace + `status='ready'` happen in one transaction. This closes the deletion race end-to-end (the #388 tail you reassigned here).
- **Feature-schema verified, not marker-trusted**: init re-ensures + verifies the required tables/indexes exist and repairs a marker-set-over-dropped-table state.

Tosko4 (changes requested):
The original terminal-state and rebuild-lineage repairs are good on `c4187544`: late same-token terminal writes no longer overwrite `ready`, and an ordinary rebuild claim keeps the previous source lineage until replacement is published.

Two concrete #387-owned blockers remain:

1. **Feature-shape validation still trusts same-named objects.** A pre-existing `lcm_rollups` with the expected object/index names but no `built_at` passed initialization with `temporal_rollups_v1` present, then failed on first build. Please validate the versioned columns, constraints, and index definitions and repair or refuse before publishing the marker.
2. **A deleted build token can alias a new period.** `resolve_no_source()` hard-deletes the `building` row. SQLite then reused that `rollup_id`; the next period received the same `(rollup_id=1, generation=0)`, and a late guarded `mark_failed()` from the deleted build changed the new period to `failed`. A build identity must remain non-reusable, or deletion must durably invalidate every old token.

Requesting changes for those remaining store/migration contracts; the reported same-token and rebuild-lineage fixes themselves pass.

100yenadmin:
Invariant remediation for T-01, T-06, T-07, T-09, T-10, T-13, T-14, and T-15 store-layer facets is published at b265d5b13e96dc5ea37719fa052c9c305d70e716. The repaired stack passed 148 focused temporal tests, 290 core tests, Ruff, diff and ancestry checks; combined schema/integration proof keeps core schema at 5 and default-off operation inert. Independent final temporal review found no P0-P3 issue at 97% confidence. This is exact-head PR proof only, not merge or release proof; maintainer review state was left unchanged.

Tosko4 (approved):
Rechecked exact head `b265d5b1` against the previously reported store, schema, token-ownership, and invalidation blockers.

The malformed-schema path now refuses safely before first use, build tokens carry a non-reusable nonce, late terminal writes cannot alias replacement rows, and durable invalidation prevents deleted or superseded source content from publishing ready. The focused temporal train rerun passed all 148 tests.

100yenadmin:

100yenadmin:

PR #388, opened 2026-07-15
[v1·temporal 2/4] rollup builder + bounded maintenance

100yenadmin:
## Summary

- **RollupBuilder** (design #385 — train PR 2 of 4, stacked on #387): UTC daily rollups synthesized
  from each day's leaf/condensed summaries, keyed by a node's **newest covered source-message
  timestamp** (`COALESCE(latest_at, created_at)`, matching `dag.py`'s source-window fallback);
  week/month aggregates built **from ready dailies only**
- synthesis rides the existing **summarize-with-escalation** path and shares the engine's summary
  circuit breaker + spend guard; hard token caps via escalation's deterministic-truncate final rung
- **two minimal flag-gated engine hooks**: post-ingest staleness marking, and bounded catch-up
  (`LCM_ROLLUP_BUILDS_PER_PASS`, default 2 — dailies before aggregates, oldest first) at the existing
  lifecycle-bind maintenance opportunity; the empty path is a single indexed query
- **failure isolation**: a rollup failure marks the row `failed` and never affects the turn;
  everything inert with `LCM_TEMPORAL_ROLLUPS_ENABLED=false` (default)
- 9 focused tests: time keying, deterministic output, caps + oversize fallback, ready-only
  aggregation + refingerprinting, end-to-end staleness→bounded rebuild, flag-off inertness (spy on
  zero queries), failure isolation

Design anchor: #385 · Stacked on: #387 · Part of #375

## Why

PR-1 gave rollups a home; this PR fills it — bounded, synchronous, and quietly. No background thread
(consistent with `docs/async-background-compaction.md`'s staged posture): stale rollups catch up a
few per pass inside maintenance work the engine already does.

```mermaid
flowchart LR
    I[Ingest] -->|mark day stale<br/>flag-gated| RS[(lcm_rollups)]
    M[lifecycle-bind maintenance<br/>≤ builds_per_pass] --> RB[RollupBuilder<br/>escalation-capped]
    RB --> RS
```

## Validation

- [x] focused builder + store tests — **21 passed**
- [x] full `pytest -q` — **1627 passed, 1 skipped, 12 xfailed**; 3 failures are the known macOS
      baseline (same three as #387, reproduced on untouched upstream `main`)
- [x] `ruff check .` — pass
- [x] `scripts/validate_release.sh --full --keep-going` — all gates pass (focused 409, benchmark +
      both stress tiers; full-suite variants carry only the same three baselines)

## Notes

- Merge order: #387 first (this stacks on it).
- Rollup scope = the active DAG session id, so ingest staleness and rollup selection share one key.
- LLM spend: zero when disabled; when enabled, rollup synthesis is governed by the same circuit
  breaker + spend guard as regular summarization — it cannot outspend the budget you already granted.


100yenadmin:
Fix round (adversarial self-review) — pushed `fix(rollups): builder + engine safety`:
- **failed rollups are retried** (30s backoff) — the maintenance scan now selects stale AND retry-eligible failed rows, not just stale.
- **wall-clock budget on the maintenance pass**: new `LCM_ROLLUP_MAINTENANCE_BUDGET_MS` (default 5000) stops starting new builds once exceeded, so a slow summarizer can't stall the session-start hot path for minutes (rollups are best-effort and may lag — documented).
- **node-deletion invalidation**: session-reset node deletion now marks referencing rollups stale (previously they referenced deleted nodes forever). 14 builder tests + 5 retain-depth.

Tosko4 (changes requested):
I reviewed this increment against #387 rather than charging inherited storage findings to this PR. The bounded maintenance structure is useful and the focused builder/store suite passes, but the builder lifecycle still has several correctness blockers of its own.

- Invalidation is still attached to raw-message ingest (`engine.py:4139-4148`) while rollups consume published summary nodes. Publishing a later summary can therefore leave an older rollup ready and apparently current.
- Maintenance selects only `stale|failed`; a process crash after `upsert_building()` leaves a `building` row that is never reclaimed.
- Invalidation during a build does not supersede the active generation. A late old builder can publish ready after newer input arrived.
- `_daily_sources()` selects every summary node in the date range, including parent and covered child nodes, so the same source coverage can be summarized twice.
- Week/month builds proceed when any daily input is ready, and a later successful daily rebuild does not stale its containing aggregate. That can publish a ready aggregate with known missing or outdated days.
- Reset/carry-forward handling remains session-bound, including a default 1,000-node enumeration limit.

Focused validation passed 27/27; the exact cumulative temporal suite passed 62/62. The missing-summary-publication, crash-stranding, duplicate-frontier and partial-aggregate cases still reproduce outside the happy-path tests.

Smallest safe path: start with an explicit/manual UTC daily builder on a generation-safe store. Defer automatic maintenance, aggregates and reset/rollover handling until publication and lifecycle ownership are explicit.

100yenadmin:
@Tosko4 addressed — `fix(rollups): staleness keyed to summary publication + aggregate completeness`:
- **Staleness now keyed to summary PUBLICATION, not raw ingest:** the two `_dag.add_node` sites (leaf at `compaction.py`, condensation at `engine.py`) call `_invalidate_rollups_for_published_node`, keying on the node's covered-message newest timestamp. A later summary now stales the affected day/week/month.
- **`building`-reclaim:** maintenance reclaims lease-expired `building` rows (crash recovery) via the #387 generation/lease.
- **Generation supersession:** builder captures the token at start, mark_ready CAS (see #387).
- **Source dedup:** `_daily_sources()` now selects a non-overlapping covering set (no parent+child double-count).
- **Aggregate completeness:** week/month publish `ready` only when all constituent dailies that actually have content are ready; a daily rebuild/stale cascades to its week AND month. (Content-aware: days with no summary content don't block, so sparse conversations still aggregate — flagged in case you'd prefer strict calendar completeness.)

Tosko4 (changes requested):
Re-review verdict for exact head `0837da2d`: **changes still required**.

The move to publication-time invalidation, expired-build reclaim, and aggregate completeness/cascade handling are meaningful fixes. The builder still has two independently reproduced snapshot/frontier races:

- `build_day()` reads `_daily_sources()` before acquiring the build token. If a summary is published in that gap, the builder can claim the newly incremented generation and then publish its older source snapshot. My probe published successfully with source `[1]` while newly published node `2` was omitted.
- `_daily_sources()` selects every summary node in the date window. A retained child and its condensed parent were both selected, so the same underlying message coverage is summarized twice rather than using a non-overlapping source frontier.

Deletion invalidation also needs to supersede in-flight builds reliably; clearing source mappings during claim makes source-ID-based invalidation unable to identify some building rows.

I’m keeping inherited store ownership findings on #387 rather than duplicating them here.

100yenadmin:
Addressed the automated review's inline findings (pushed `fix(rollups): capture-token-first builds, publication-only staleness, source dedup, scope gating`):
- **Staleness is publication-only now**: ground-truthing showed both `add_node` sites already fired the publication hook — the actual bug was the *premature raw-ingest mark* (stale before the summary exists → rebuild from old sources → ready-omitting-the-leaf). That raw-ingest hook is removed; invalidation keys purely to summary-node publication.
- **Builds capture their generation token BEFORE reading sources**, so an invalidation between snapshot and claim supersedes the build.
- **Deletion-staleness bumps the generation** (an in-flight build can't publish deleted-node content) and captures **all** deleted node ids (no 1000-row truncation).
- **Maintenance is gated off for bypassed/stateless sessions** (matches the ingest-write gating).
- Rotation scope: an on-rollover re-scope collides with the UNIQUE key and would need the core-schema column your #387 review ruled out — documented as the boundary; the #389 read-side span keeps results correct across it.

Tosko4 (commented):
Targeted re-review of exact head `4bb826f3`: the #388-local blockers from the previous head are resolved.

I verified that the build token is captured before the source snapshot, an invalidation/deletion arriving in flight supersedes the old build, same-day condensed children are excluded when their parent covers them, and no-source jobs no longer remain stuck as maintenance debt. The temporal focused suite is green.

I did not find a new #388-local blocker in this delta. The stacked train is still not merge-ready because its #387 feature-migration substrate remains blocked; I’ll keep that finding on the owning PR rather than duplicate it here.

Tosko4 (changes requested):
Follow-up correction on exact head `4bb826f3`: the prior #388 fixes are real, but my earlier “no new local blocker” conclusion was incorrect.

- Source dedup is still grouped by the node’s `latest_at` day. With a child covering July 15 and a parent condensing that child while spanning July 15–16, the July 15 source set contained the child and July 16 contained the parent. Adjacent daily rollups therefore duplicate the same covered lineage across a multi-day request.
- Publication invalidation still receives only `latest_at` (with `created_at` as fallback). For a newly published summary spanning July 15–16, I reproduced July 15 remaining `ready` while only July 16 became `stale`.

The deletion race is also not closed end-to-end because #387 clears source lineage when claiming a rebuild; that inherited store issue belongs on #387. Please make the daily frontier interval-aware and invalidate every UTC day intersected by the summary’s coverage interval.

100yenadmin:
Addressed (`fix(rollups): interval-aware coverage helper for daily frontier + publication staleness`) — one shared helper (`rollup_periods.py`: `covered_days`/`canonical_frontier`) now drives both:
- **B1 daily frontier is interval-aware**: the canonical frontier suppresses a child whose lineage a spanning parent already covers, so adjacent dailies no longer duplicate it (your Jul15-child / Jul15-16-parent repro is a test). **Deliberate semantic worth your eye:** a multi-day node feeds only its representative day (the parent → Jul16), so Jul15's *daily* legitimately doesn't exist; a Jul15 query still surfaces that content via the interval-aware fallback (which keeps the covering parent). If you'd prefer per-day attribution instead of representative-day consolidation, say so — it's a frontier-policy change, not a rework.
- **B2 publication staleness now stales every UTC day the coverage interval intersects** (+ containing week/month) via `covered_days` — your Jul15-16 summary now stales both days, not just Jul16.

Tosko4 (changes requested):
The reported interval/frontier repairs pass on `6ea1f02a`: the spanning parent suppresses its covered child, uses the documented representative day, remains available through interval-aware fallback for the earlier day, and publication stales every intersected day plus its aggregates.

Three normal lifecycle paths still need #388-owned recovery:

1. **First-build deletion race.** A new build has no source rows until `mark_ready()`. Deleting its source during summarization therefore invalidated zero rollups, and the late publish succeeded as `ready` with the deleted source. Register the claimed generation's sources before blocking work, or revalidate their liveness atomically at publication.
2. **Disabled → enabled reconciliation.** A ready rollup remained `ready` with its old source after another summary for that day was published while temporal rollups were disabled and the feature was reopened. Re-enable needs a durable publication watermark/frontier reconciliation rather than assuming every historical publisher ran the optional hook.
3. **Lost invalidation debt.** `mark_stale_for_published_summary()` catches every storage failure and returns zero. There is no durable retry/reconciliation record, so a transient failure can leave a stale `ready` rollup indefinitely.

The aggregate-cascade concern from the current inline review does **not** reproduce: deleting a source correctly staled its day, week, and month. Requesting changes only for the three concrete lifecycle gaps above.

100yenadmin:
Invariant remediation for T-01, T-02, T-03, T-04, T-05, T-07, T-12, and T-15 builder/engine facets is published at eefc2951ac34eda50451892664d4fd4c5b4140d4. The repaired stack passed 148 focused temporal tests, 290 core tests, Ruff, diff and ancestry checks, including interval, canonical-frontier, atomic invalidation, bounded-work, and concurrent scope regressions. Independent final temporal review found no P0-P3 issue at 97% confidence. This is exact-head PR proof only, not merge or release proof; maintainer review state was left unchanged.

Tosko4 (approved):
Rechecked exact head `eefc2951` against the previously reported builder and lifecycle blockers.

First-build deletion, publication ordering, multi-day invalidation, disable/re-enable reconciliation, pending-outbox fail-closed retrieval, and canonical frontier behavior now follow the intended contract. The focused temporal train rerun passed all 148 tests.

100yenadmin:

100yenadmin:

PR #389, opened 2026-07-15
[v1·temporal 3/4] lcm_recent natural-time retrieval

100yenadmin:
## Summary

- **`lcm_recent`** — natural-time retrieval over LCM history (design #385 — train PR 3 of 4, stacked
  on #388): `today | yesterday | Nd | week | month | date:YYYY-MM-DD | last Nh`, parsed to UTC
  `[start, end)` windows
- serves **ready** day/week/month rollups newest-first, with per-rollup token lines, provenance
  (which rollups served, status), limit clamping, and the shared 20k-char response ceiling
- **transparent degradation is the load-bearing behavior**: missing/stale/disabled rollups and
  sub-day windows fall back to existing depth-0 summary retrieval in the same window — never an
  error, and provenance says `fallback: true`
- serving is **read-only and makes zero LLM calls** (building stays in #388's bounded maintenance)
- 66 focused tests: every period form, invalid inputs, ready/stale/disabled/sub-day modes, scopes,
  bounds, provenance; docs with three usage examples

Design anchor: #385 · Stacked on: #388 (which stacks on #387) · Part of #375

## Why

This is the PR the first two exist for: "what did we work on yesterday / last week" becomes a single
tool call that answers from prebuilt rollups when they exist and behaves exactly like today's
retrieval when they don't.

```mermaid
flowchart LR
    Q[lcm_recent<br/>natural-time query] --> P[UTC window parser]
    P -->|ready| RS[(lcm_rollups)]
    P -.->|absent / stale /<br/>disabled / sub-day| F[existing leaf-summary<br/>retrieval fallback]
```

## Validation

- [x] focused `lcm_recent` + public-registration/packaging contracts — **66 passed**
- [x] full `pytest -q` — **1652 passed, 1 skipped, 12 xfailed**; 3 failures are the known macOS
      baseline, reproduced identically on untouched upstream `main` (isolated-HOME run)
- [x] `scripts/validate_release.sh --full --keep-going` — all gates pass (focused 409, benchmark +
      both stress tiers; full-suite variants carry only the same three baselines)

## Notes

- Merge order: #387 → #388 → this.
- Beyond the spec'd files, the repo's hard-coded public-tool surfaces (`engine.py` tool list,
  `__init__.py`, `plugin.yaml`) needed registration-only lines so the tool is reachable and the
  synchronized tool-contract suite stays valid — no engine lifecycle/compaction behavior touched.
- `week`/`month` serve only exact calendar aggregates; any non-ready row in the window forces the
  fallback (conservative by design — no partially-fresh answers).


100yenadmin:
Fix round (adversarial self-review) — pushed `fix(rollups): drop unreachable global scope`: `lcm_recent` advertised `scope='global'` but nothing ever writes a global-scoped rollup, so it silently always fell back. Now only `scope='conversation'` is accepted/advertised, with cross-session rollups noted as future work. 29 lcm_recent tests.

Tosko4 (changes requested):
The scope cleanup here is good: the unreachable global mode is gone, and the public surface now consistently describes current conversation/session behavior. The natural-time parsing and bounded fallback direction are useful.

I’m still requesting changes for two retrieval-contract issues introduced in this slice:

- The rollup path cannot prove complete window coverage. `_recent_has_unready_rollups()` notices only existing non-ready rows, while `_recent_ready_rollups()` accepts any non-empty ready result. A two-day request with only one ready day therefore returns rollup mode instead of falling back for the complete window.
- Fallback is limited to depth-0 summaries from the current session, excluding retained higher-depth/carry-forward summaries after rotation.

It also remains dependent on #387/#388 for trustworthy publication and aggregate completeness; I’m not duplicating those lower-train findings here.

Focused validation: `tests/test_lcm_recent.py` passed 29/29, but the missing-day window probe still returned rollup mode without a fallback reason.

Smallest safe path: land the parser plus canonical summary fallback first, then add rollup serving once the store can provide an authoritative complete-window census and durable conversation identity.

100yenadmin:
@Tosko4 addressed — `fix(rollups): lcm_recent requires full-window coverage + broader fallback`:
- **Complete-window proof:** `_recent_has_unready_rollups` now detects MISSING days (no row), not only existing non-ready rows, and rollup mode is served only when the ENTIRE requested window is covered by ready rollups; otherwise it falls back for the whole window. A 2-day request with one ready day now falls back instead of returning partial rollup output.
- **Fallback breadth:** the leaf-summary fallback now includes retained higher-depth / carry-forward summaries in-window, not just depth-0 current-session summaries.

Tosko4 (changes requested):
Re-review verdict for exact head `12f18394`: **changes still required**.

The previous missing-day coverage bug is fixed, and fallback now includes retained higher-depth summaries. Two PR-local retrieval gaps remain and reproduce on this exact SHA:

- Conversation rollup serving checks only `engine.current_session_id`, while fallback spans the current plus last-finalized session. A ready current-session rollup therefore enabled rollup mode and masked matching content from the finalized session that fallback would have returned.
- Fallback filters summaries by `latest_at` falling inside the window rather than interval overlap. A summary whose `[earliest_at, latest_at]` crossed the requested day was omitted entirely.

So the parser/fallback direction is solid, but rollup mode still cannot prove complete conversation-window coverage across lifecycle rotation. Lower publication/frontier blockers remain owned by #387/#388.

100yenadmin:
Addressed the automated review's inline findings (pushed `fix(rollups): lcm_recent spans the fallback session set + overlap-based leaf window`):
- **Rollup mode and fallback now check the same session span** (current + last finalized), so rollup-mode answers can't silently cover less than the fallback would.
- **The leaf-summary window filter is overlap-based**: summaries whose covered span intersects the window are included, not only those whose `latest_at` lands in it — multi-day summaries now appear for every day they cover.
- The missing-day / Nd-window completeness check and building/failed retry paths flagged on the older heads were fixed in the previous round (verified per-item — dispositions in the fix notes).

Tosko4 (commented):
Targeted re-review of exact head `ce0229ef`: the #389-local serving blockers from the previous head are resolved.

I verified whole-window fallback when any expected daily is missing, fallback across the current + last-finalized session span, and interval-overlap selection for summaries crossing a window boundary. Those direct regressions pass on this head.

I did not find a new #389-local blocker in this delta. The temporal train remains blocked by the owning substrate/rebuild findings on #387 and #391; I’m not duplicating those inherited findings here.

Tosko4 (changes requested):
Follow-up correction on exact head `ce0229ef`: the original window/session blockers are fixed, but my earlier “no new local blocker” conclusion missed two serving issues.

- The leaf fallback selects every overlapping summary without suppressing a child covered by an overlapping higher-depth parent. My exact probe returned both `parent-content` and `child-content`, so duplicated lineage can consume the public `limit` ahead of independent summaries.
- Rollup provenance is serialized outside the section budget. With 1,000 ready rollups and `limit=1`, the response was **39,039 characters** against the advertised **20,000-character** cap, contained all 1,000 provenance rows, and had room for zero content sections.

Please apply one interval-aware canonical frontier to fallback selection and bind provenance to returned sections, or replace it with bounded aggregate/truncation metadata.

100yenadmin:
Addressed (`fix(rollups): canonical frontier for lcm_recent fallback + provenance bound to returned sections`):
- **C1 fallback uses the same canonical frontier** as the daily builder (shared helper), so a child covered by an overlapping higher-depth parent is suppressed — your parent-content/child-content probe now returns only the canonical node, and the public `limit` isn't consumed by duplicated lineage.
- **C2 provenance is bound to returned sections**: `provenance.rollups` lists only the rollups backing the sections actually returned, plus an O(1) `rollups_covered` count — your 1,000-rollup/`limit=1` case now stays within the char cap instead of emitting 39k chars of provenance.

Tosko4 (commented):
Verified on `4699e580`.

The fallback now uses the shared canonical frontier, so the parent/covered-child case returns only the canonical parent and no longer spends the public limit twice on the same lineage. Provenance is derived from the sections actually returned; the 1,000-rollup / `limit=1` case stayed bounded and reported one backing rollup.

I found no remaining #389-local retrieval blocker. Because this PR is cumulative from `main`, it remains blocked by the inherited #387/#388 lifecycle gates rather than receiving a full approval.

100yenadmin:
Invariant remediation for T-03, T-05, and T-12 recent/fallback facets is published at e829139719597e660bcbc8ec52aa790d8ab335ae. The repaired stack passed 148 focused temporal tests, 290 core tests, Ruff, diff and ancestry checks, including bounded limit=1 work, canonical fallback, interval-aware selection, and capped provenance. Independent final temporal review found no P0-P3 issue at 97% confidence. This is exact-head PR proof only, not merge or release proof; maintainer review state was left unchanged.

Tosko4 (approved):
Rechecked exact head `e8291397` against the prior fallback, coverage, deduplication, and provenance blockers.

Fallback remains fail-closed, uses the canonical frontier, does not double-spend the public limit, and bounds provenance to the sections actually returned. The cumulative temporal train rerun passed all 148 focused tests.

100yenadmin:

100yenadmin:

PR #381, opened 2026-07-14
feat: cap the protected fresh tail by tokens

100yenadmin:
## Summary

- cap the **protected fresh tail by tokens** as well as message count: the boundary resolver walks
  backward accumulating tokens and stops at the configured budget, so a handful of enormous recent
  messages can no longer pin an unbounded prompt region out of compaction's reach
- **tool-call groups retreat atomically** — the boundary never splits an assistant tool-call from its
  tool results (with an orphan-bailout when a group can't be completed)
- the newest message is always kept; behavior is **opt-in**
  (`LCM_FRESH_TAIL_MAX_TOKENS`, default off → existing count-only behavior is unchanged)
- exposed in `lcm_inspect`/rotate/config surfaces; 9 focused boundary-resolver tests

We maintain Martian Engineering's Lossless Claw and are porting the proven pattern into Hermes-LCM.

Tracker: #375
Closes #377

## Why

`fresh_tail_count` protects the N most recent messages from compaction — but N messages can be 2K
tokens or 200K. Under a token-denominated context budget, a count-only tail makes compaction's
effectiveness depend on message-size luck. A token bound makes the protected region predictable
without changing what "fresh" means for well-behaved histories.

### Where this hooks into the pipeline

```mermaid
flowchart LR
    I[Ingest] --> C[Compact] --> D[Condense] --> ES[Escalate] --> A[Assemble] --> R[Retrieve]
    C -.->|fresh-tail boundary:<br/>count AND token bound,<br/>tool groups atomic| I
    classDef hot fill:#b45309,stroke:#7c2d12,color:#fff
    class C hot
```

Only the compaction-boundary computation changes; assembly and retrieval are untouched.

## Validation

- [x] focused `tests/test_fresh_tail.py` — **9 passed** (boundary resolver, token accumulation,
      tool-group retreat + orphan bailout), re-verified against the newly-merged #371 non-blocking
      tiktoken loader (this branch calls the exact `tokens.py` functions #371 changed — no breakage)
- [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` —
      **1014 passed, 1 skipped**
- [x] full `pytest -q` — **1617 passed, 1 skipped, 12 xfailed**; 2 failures are the known macOS
      `/var` vs `/private/var` baseline, verified byte-identical on untouched upstream `31675c6` —
      Linux CI is the canonical gate
- [x] `ruff check .` — pass
[...]

PR #390, opened 2026-07-15
[v1·embeddings 1/5] embedding vector storage substrate + KNN ladder

100yenadmin:
## Summary

- **embedding storage substrate** (design #386 — embeddings train PR 1, benchmark-driven):
  `lcm_embedding_profile` (dim locked at registration, immutable) / `lcm_embedding_meta`
  (model-fingerprinted) / `lcm_embedding_vectors` (float32 BLOBs) via a new schema step
- **the #386 compute ladder, exactly as benchmarked**: NumPy full scan when importable (with a
  `(model, max(rowid), row_count)` matrix cache invalidated on writes — correctness never depends on
  the cache), bounded most-recent pure-Python scan otherwise, `coverage: full|bounded|none` reported
  on every result
- L2-normalization on write; DELETE+INSERT under one transaction (no upsert); filter-aware overfetch
  (k×10 cap 500) with post-scan joins for time/conversation filters; app-level purge for hard-deleted
  nodes
- **default-off and inert**: nothing constructs the store yet — no provider, no HTTP, no tool wiring
  (train PRs 2–4)
- 14 tests incl. hand-computed cosine ordering, NumPy-absent fallback, bounded-window enforcement,
  cache invalidation, migration idempotency/upgrade/version-guard

Design anchor: #386 · Part of #375

## Why

Per #386's benchmark: NumPy meets a 300 ms interactive budget even cold-loading 50k×1024 from SQLite
(≤69 ms), unbounded pure-Python fails past ~20k vectors, and sqlite-vec can't load on stock/system
Python — so the store is dependency-free by default with NumPy as an optional accelerator and an
honest `coverage` signal when running bounded. This PR is the data model + math only, reviewable in
isolation before any behavior exists.

```mermaid
flowchart LR
    VS[(lcm_embedding_profile /<br/>meta / vectors BLOB)]
    L1[NumPy full scan<br/>+ matrix cache] --> VS
    L2[bounded pure-Python scan<br/>coverage: bounded] -.fallback.-> VS
    L3[coverage: none<br/>callers stay FTS-only] -.fallback.-> L2
```

## Validation

- [x] `pytest -q tests/test_vector_store.py` — **14 passed**
- [x] full `pytest -q` — **1620 passed, 1 skipped, 12 xfailed**; 3 failures are the known macOS
      baseline, reproduced identically on untouched upstream `main`
- [x] `ruff check .` — pass
- [x] `scripts/validate_release.sh --full --keep-going` — compile/shell/diff gates, focused pytest
      (409), benchmark smoke, both stress tiers pass; the two full-suite gates hit a known
      local-load-only wedge in an 8-thread Barrier migration test (frozen logs, terminated; the test
      then passed 3/3 in isolation and a fresh bounded full run showed only the same three baselines
      — same phenomenon documented on #381/#384)

## Notes

- **Schema-version heads-up**: this PR and the temporal train's #387 independently claim the next
  schema version (`v6_embeddings` vs `v6_temporal_rollups`). Whichever lands second renumbers to v7 —
  isolated to a two-line constant change here by design. Happy to rebase whichever order you merge.
- Upstream `main` has no `summary_nodes.suppressed_at` column today; the post-scan join enforces
  suppression when the column exists and is valid without it (hard-deleted summaries are always
  excluded; purge removes their vectors).
- `conversation_ids` filters target `summary_nodes.session_id`; `since` uses
  `COALESCE(latest_at, created_at)` — matching DAG recency semantics.

### Schema-migration disclosure (adversarial-review follow-up)

To be precise about the "no behavior change when disabled" claim: like every schema-backed feature in
this repo, the migration is **unconditional** — opening an `lcm.db` with this build bumps
`SCHEMA_VERSION` to the next value and creates the new tables even with the feature flag off. Those
tables stay **empty** and nothing reads or writes them until the feature is enabled, so there is no
*behavioral* change — but the on-disk schema does advance. One consequence worth stating: a user who
tries this build and then rolls back to an older release will hit `refuse_schema_version_too_new` (the
older build won't open a newer-schema DB). That is inherent to any additive schema step here, not
specific to this PR; flagged explicitly so it's a conscious choice rather than a surprise.


100yenadmin:
Fix round (adversarial self-review) — pushed `fix(vector-store): KNN correctness`:
- **bounded (no-numpy) scan now takes the NEWEST vectors**: backfill writes newest-content-first, so newest rows got the lowest rowids and the bounded 'most recent N' scan was actually returning the OLDEST. Now ordered by `embedded_at DESC`.
- **filters applied before the top-k score cap**: since/conversation filters previously ran AFTER truncating to the top-500-by-score, so true filtered matches outside that window were dropped (could return <k or empty while coverage='full'). Now filter → score → top-k. 15 vector-store tests incl. a case with 501 higher-scoring unfiltered vectors ahead of the matches.

Tosko4 (changes requested):
Thanks for the KNN cleanup. I re-reviewed this exact head as the embeddings architecture gate. The dependency-free SQL window is now genuinely bounded at the database-work level, and the NumPy path filters session/time before scoring. Those are meaningful improvements.

I’m still requesting changes because the storage identity and compatibility contract remain unsafe:

- Ordinary startup still creates embedding tables and advances the core schema to v6 even when embeddings are disabled. A committed v5 database opened here becomes unreadable by exact base code, and the temporal train independently claims the same v6 marker.
- Profiles/vectors are keyed only by `model_name`; provider, revision, task, codec/dtype and byte order are not canonical identity. Registering the same model name under another provider silently keeps the old provider metadata, and switching back does not reactivate the intended profile.
- The NumPy cache key is `(model, max_rowid, row_count)`. Another process can rewrite a vector while those values stay unchanged, leaving one process with stale similarity results.
- The no-NumPy path still applies its bounded candidate cap before session/time eligibility filtering, so eligible older rows can be hidden behind newer out-of-scope rows.
- The NumPy metadata lookup emits one SQL parameter per embedded ID; 33,000 IDs exceeded this runtime’s SQLite variable limit.
- Legacy `summary_nodes` without `latest_at` still fail filtered KNN, and the `CAST(node_id AS TEXT)` predicate prevents direct rowid lookup.

Focused validation: `tests/test_vector_store.py` passed 15/15. The committed downgrade, cross-process cache, bounded-eligibility and large-ID probes still reproduce.

Smallest safe path: feature-owned/feature-gated derived schema, strict profile+codec identity and validated CRUD first. Add cache/filter/KNN claims after cross-process invalidation, legacy compatibility and bounded eligibility are explicit.

100yenadmin:
@Tosko4 addressed — `fix(vector-store): schema v5 + canonical identity + durable data-version`:
- **Schema:** core `SCHEMA_VERSION` back to **5**; embedding tables created lazily via a named `embeddings_v1` step, gated on enabled. No core-version bump → base code opens the DB, no v6 collision with the temporal train (which does the same).
- **Canonical identity:** profiles/vectors now keyed by `_identity_hash(provider, model, revision, dim, dtype, byteorder, task)` — re-registering the same model name under a different provider creates a distinct profile (no silent metadata clobber), and switching config back to a prior identity **reactivates** that profile with its vectors intact (no re-backfill).
- **Cross-process cache:** added a durable per-identity `data_version` counter bumped **atomically inside every vector write**; the numpy cache key includes it, and reads use an autocommit connection so a write from another process is visible — no more stale similarity results. The 33k-id metadata lookup is now a chunked temp-table JOIN (scales past the SQLite variable limit).

Tosko4 (changes requested):
Re-review verdict for exact head `31a84e70`: **changes still required**.

The architecture correction is real: core schema stays v5, embedding tables are lazy under `embeddings_v1`, profile rows now use canonical identities, cache invalidation has a durable data version, and the large-ID temp-table path removes the SQLite parameter ceiling.

Three independently reproduced substrate blockers remain:

- Without NumPy, the global bounded window is selected before conversation/time/source eligibility. With a one-row window, a newer out-of-scope vector hid an older eligible vector and KNN returned no result.
- `_lcm_id_scratch` is one connection-scoped temp table shared by concurrent operations. A deterministic two-thread interleaving made the A query return `[]` after the B query replaced the scratch contents.
- KNN resolves by model string, not provider identity. With provider B configured and only provider-A vectors registered under the same model name, semantic retrieval did not degrade; it scored and returned the provider-A corpus.

Canonical storage identity is therefore fixed, but retrieval still needs to carry and enforce that identity, and filtered scratch work must be isolated or serialized.

100yenadmin:
Addressed the automated review's inline findings on this head (pushed `fix(vector-store): legacy-db compat, PK-index binds, per-call scratch, filter-before-bound, identity read path, purge wiring`):
- **Legacy-DB compatibility**: time-scoped KNN no longer crashes on v5 DBs lacking `summary_nodes.latest_at` (PRAGMA-detected fallback to `created_at`), and the source filter detects/skips when `messages.source` is absent on a VectorStore-only worker DB.
- **Index usability**: the `node_id` candidate filter binds parameters instead of CASTing the column, so the INTEGER-PK lookup is used; the bounded no-numpy scan now orders from an indexed column so the bound is real.
- **Concurrency**: per-call unique scratch temp tables (concurrent `knn()` calls no longer clobber each other); the active profile resolves under the write lock; the bounded path filters before the recency bound.
- **Identity in the read path**: KNN resolves the profile by full provider identity, not `model_id` alone — switching provider A→B with the same model name can no longer serve the wrong profile's vectors.
- **Purge wired**: `purge_embeddings_for_nodes` is now actually called from the deletion paths (session-reset depth deletes + `doctor clean apply`), with orphan-filtering before ranking as defense-in-depth.
Three of the flagged items were already handled by the previous round (scalar error redaction, role degrade, orphan filter) — noted per-item in the outcome table.

Tosko4 (changes requested):
Targeted re-review of exact head `6a548ef4`: the concrete vector-store fixes are real. I verified filter-before-bound behavior, full provider-identity resolution, per-call scratch tables under concurrent KNN, legacy shape handling, and purge/orphan exclusion.

One #390-owned migration blocker remains under the architecture decision on #375. Feature initialization still accepts an existing same-name profile table without validating its full shape, publishes `embeddings_v1`, and only fails when the store is used. I reproduced this with `lcm_embedding_profile(identity_hash, provider, model_name)`: `VectorStore` initialized and marked the migration complete, then `register_profile()` failed with `no such column: dim`.

Please make feature initialization atomic with marker publication, or validate the complete versioned table/index/constraint shape before publishing the marker and provide an explicit repair/refusal path. A malformed or interrupted feature schema must not be recorded as complete.

Tosko4 (changes requested):
Follow-up correction on exact head `6a548ef4`: retrieval now resolves full provider identity, but my earlier statement that provider identity and legacy handling were fully fixed was too broad.

- `record_embedding()` accepts only a bare model name and resolves the active identity after the provider call. I produced a vector under provider A, switched the same model/dimension to provider B before publication, and the provider-A vector was stored under provider B’s identity.
- When a legacy `messages` table lacks `source`, `_source_allowed_ids()` treats every candidate as eligible. A query for a source that definitely did not exist returned the legacy summary instead of degrading or failing closed.
- The dependency-free “bounded” path materializes every live candidate ID before filtering and applying `bounded_scan_rows`; with a 100-row corpus and a bound of 10 it still loaded 100 IDs. Vector loading is bounded, but SQL work and host-memory enumeration are not.

Please carry the exact durable identity from provider resolution through publication CAS, fail/degrade source filtering closed when provenance cannot be verified, and make the fallback candidate enumeration genuinely bounded. The feature-schema validation/marker blocker from the previous review also remains open.

100yenadmin:
Addressed (`fix(vector-store): durable identity through write CAS, fail-closed source filter, SQL-bounded enumeration, schema verify`):
- **Identity carried through the write CAS**: `record_embedding(..., identity=)` publishes under the identity captured at provider-resolution, inside the write transaction — your A→B-mid-flight repro now stores the vector under A's identity (or rejects), never rebinds it onto B. (Backfill wires the captured identity through.)
- **Source filter fails closed**: when `messages.source` is absent, `_source_allowed_ids` returns empty rather than treating all candidates as eligible — your nonexistent-source query no longer returns the legacy summary. (Flipped the old fail-open test.)
- **Enumeration bounded at SQL**: filters in `WHERE` + `ORDER BY embedded_at DESC` + `LIMIT` — your 100-corpus/bound-10 case now enumerates ~10, not 100.
- **Schema verified + repaired**, not marker-trusted (mirror of #387).

Tosko4 (changes requested):
The captured-identity, legacy fail-closed source handling, generic SQL-bound, and per-call scratch repairs are good on `455c2c0d`.

Two concrete #390-owned blockers remain:

1. **Feature-shape validation still trusts same-named objects.** A pre-existing `lcm_embedding_profile` with the expected object/index names but no `dim` passed initialization with `embeddings_v1` present, then failed in `register_profile()`. Validate the versioned columns, constraints, and index definitions and repair or refuse before publishing the marker.
2. **The bounded dependency-free path caps before source eligibility.** With bound 10, ten newer wrong-source vectors, and one older eligible vector, `knn(..., source='keep')` returned no hit. Candidate IDs are limited by recency before `_apply_source_filter()` runs, so an eligible vector just beyond the bound is invisible. Push source eligibility into the bounded SQL candidate query before `LIMIT`.

Requesting changes only for those remaining migration/filter contracts; the reported identity publication and unfiltered bounded-enumeration fixes pass.

100yenadmin:
Invariant remediation for E-01, E-04, E-07, E-08, E-09, and E-10 vector-store facets is published at 0c7119f693a2b0804f0455f92efe84407a365628. The repaired stack passed 159 focused embeddings tests, 290 core tests, Ruff, diff and ancestry checks, the 250001-row bounded planner regression, and deterministic recall with no regression. Core schema remains 5 and default-off operation is inert. Independent final embeddings review found no P0-P3 issue at 98% confidence. This is exact-head PR proof only, not merge or release proof; maintainer review state was left unchanged.

Tosko4 (approved):
Rechecked exact head `0c7119f6` against the previously reported schema, identity, bounded-selection, legacy-filter, and scratch-state blockers.

Malformed same-name feature objects are rejected before use, captured provider identity is preserved, bounded candidate materialization is SQL-capped, legacy source handling fails closed, and per-call scratch state is isolated. My bounded exact-head rerun passed 49 vector-store tests; the very large planner stress case was intentionally not repeated in this pass.

100yenadmin:

100yenadmin:
Thanks for `f9a240a` — the source-recency ordering is the right call (newest-first backfill genuinely inverted the bounded window; verified your test logic and ran the combined suites: 87/87 green, and our 3,872-summary production-import archive orders correctly under it).

One hardening pushed on top (`716e059`): the ordering now uses `COALESCE(sn.latest_at, sn.created_at)` per-row. The DAG migration adds `latest_at` without backfilling legacy rows, and a bare `latest_at DESC` sorts those NULLs last — silently dropping legacy summaries out of the bounded window on upgraded databases (our imported archive happens to have 0 NULLs, but upgraded-in-place DBs won't be so lucky). Regression test included (`test_bounded_scan_keeps_null_latest_at_legacy_rows_by_created_at`); propagated up the train the same way you did.

100yenadmin:

PR #392, opened 2026-07-15
[v1·embeddings 2/5] pluggable embedding providers + warmup

100yenadmin:
## Summary

- **pluggable embedding providers** (design #386 — embeddings train PR 2 of 5, stacked on #390):
  `voyage` (raw HTTP, no SDK), `ollama` (`/api/embed`), `fastembed` (optional local dep) — resolved
  from `LCM_EMBEDDING_PROVIDER`/`LCM_EMBEDDING_MODEL`, dormant when unset
- Voyage client ports our production discipline: document/query `input_type` asymmetry,
  `truncation: false` (fail loud, never silently clip), 80k-token request bin-packing with 27k
  per-doc skip-and-report, bounded retries, Retry-After budgeting, classified errors, **PII-scrubbed
  error logging**
- **local models never download lazily**: FastEmbed runs `local_files_only=True` on normal calls;
  only `/lcm embed warmup` may download — the one command where dimension discovery happens and the
  profile gets registered (#371's lesson applied at model scale)
- provider-local circuit breaker + call-rate spend guard (own instances, mirroring the escalation
  shapes); injectable transport — **zero live calls or downloads in tests**
- 21 all-mocked tests; `VOYAGE_API_KEY` read from env only, never stored or logged

Design anchor: #386 · Stacked on: #390 · Part of #375

## Validation

- [x] `pytest -q tests/test_embedding_provider.py` — **21 passed** (110 across the stacked
      provider/vector/command bundle)
- [x] full `pytest -q` — **1641 passed, 1 skipped, 12 xfailed**; 3 failures are the known macOS
      baseline (same three as #390, reproduced on untouched upstream `main`)
- [x] `ruff check .` — pass
- [x] `scripts/validate_release.sh --full --keep-going` — all gates pass; full-suite variants carry
      only the same three baselines

## Notes

- Merge order: #390 → this.
- Provider resolution does no network access; nothing embeds anything yet — backfill and search modes
  are the next two cars.
- The `/lcm embed warmup` output includes a one-line cost note per provider (Voyage free-tier /
  local-$0) — the full setup guide is the train's docs car.


100yenadmin:
Fix round (adversarial self-review) — pushed `fix(embedding-provider): resilience + safety`:
- **token safety margin**: per-doc/per-batch caps now at 0.9× the Voyage limits (24.3k/72k), since the local count can under-count (char-estimate before tiktoken loads; non-Latin scripts) and `truncation:false` makes an over-cap request a hard 400.
- **broader PII scrub**: error-body redaction is now recursive and covers free-text under `detail`/`msg`/`error`/`message` (was key-name-based, leaking echoed content under those envelopes); status + error `kind` preserved.
- **bounded interactive call path**: query embedding uses an explicit HTTP timeout, one attempt, no Retry-After sleep. 22 provider tests.

Tosko4 (changes requested):
There are good fixes in this pass. Provider error bodies now redact scalar strings correctly, and interactive Voyage/Ollama requests use one attempt with an explicit timeout. The earlier scalar-error logging concern is resolved on this head.

The provider slice still has four PR-local blockers:

- Voyage batches are bounded by estimated tokens but not by the provider’s item-count cap; a 1,001-document batch was sent as one request.
- Retry timeout is still per attempt rather than an absolute operation deadline. A `0.02s` timeout with three attempts took about `1.5s` including backoff.
- Ollama requests do not set `truncate: false`, so oversized input can be silently truncated rather than failing clearly.
- FastEmbed query embeddings still use generic `model.embed()` rather than the query-specific API, so query/document task asymmetry is lost.

Identity collision findings belong to #390 and are intentionally not duplicated here. Focused provider/vector tests passed 37/37.

Smallest safe path: provider resolution and explicit warmup after adding item caps, fail-loud truncation, correct query/passage APIs and one monotonic deadline across retries/backoff.

100yenadmin:
@Tosko4 addressed — `fix(embedding-provider)`:
- **Voyage item-count cap** (1000/request, configurable) in addition to the token cap — a 1001-doc batch now splits.
- **One absolute operation deadline** across all attempts+backoff (replaces the per-attempt timeout) — a 0.02s budget returns in ~0.02s, not 1.5s.
- **Ollama `truncate: false`** so oversized input fails loudly instead of being silently truncated.
- **FastEmbed query API** (`query_embed`) for query embeddings so query/document task asymmetry is preserved.

Tosko4 (commented):
Re-review verdict for exact head `c0629697`: **the #392-local blockers are resolved**.

I independently verified all four requested provider fixes:

- 1,001 Voyage documents split into requests of 1,000 + 1;
- a 20 ms interactive budget prevents retry/backoff from extending the operation;
- Ollama sends `truncate: false`;
- FastEmbed uses `embed()` for documents and `query_embed()` for queries.

The scalar provider-error redaction concern is also superseded on this head; the private scalar strings are removed in the focused test.

Validation: the provider suite passed 26/26, including the four targeted regressions. I found no new provider-local blocker. I’m not approving the stacked train because #390’s vector-store identity/filter/concurrency gates remain below it.

Tosko4 (changes requested):
Follow-up on exact head `c0629697`, correcting my earlier comment after broader provider probes. The focused fixes remain valid: the default Voyage cap splits 1,001 items into 1,000 + 1, the interactive query path has an operation deadline, Ollama fails loudly instead of truncating, FastEmbed uses the query API, and scalar error content is scrubbed. Two provider-local contracts are still open:

1. Normal `embed_query()` and `embed_documents()` calls do not pass an operation deadline into `_request()`. With `timeout=0.02` and a synthetic network failure, a normal query made three attempts and took 1.500 seconds because the 0.5s and 1.0s backoffs remained outside any total budget.
2. `max_batch_items` accepts values above Voyage's hard 1,000-item API limit. Configuring it to 2,000 sent one request containing 1,001 inputs. The provider reference caps the list at 1,000: https://docs.voyageai.com/reference/embeddings-api

Please enforce one absolute deadline for normal provider operations as well as interactive queries, and clamp or reject Voyage batch limits above 1,000. Both cases need direct regressions through the public provider methods.

Tosko4 (changes requested):
Targeted re-review of exact head `dc99cd30`: the focused interactive deadline and default 1,000-item split pass, but the two provider-local contracts from the previous review remain open.

1. Normal `embed_query()` / `embed_documents()` still call `_request()` without an operation deadline. With `timeout=0.02` and three retryable failures, a normal query made three attempts and took **1.501s** because the 0.5s + 1.0s backoffs remained outside a total budget.
2. `max_batch_items` still accepts values above Voyage’s hard 1,000-input limit. Configuring 2,000 emitted one request containing **1,001 inputs**.

Please apply one absolute deadline to normal provider operations as well as the interactive path, and clamp or reject any Voyage item cap above 1,000. The current tests cover the safe defaults/lower custom values but not these public-constructor escape paths.

100yenadmin:
Addressed (`fix(embedding-provider): one absolute deadline on normal ops; clamp Voyage item cap at 1000`):
- **Normal `embed_query`/`embed_documents` now share the one absolute monotonic deadline** (retries+backoff inside it), same as the interactive path — your 0.02s-budget/3-retry case now returns in ~0.02s, not 1.5s.
- **Voyage `max_batch_items` clamped to ≤1000** (the provider's hard limit is authoritative regardless of config) — your 2000 config now never emits a >1000-input request.

Tosko4 (changes requested):
The retry/backoff repair and Voyage's hard 1,000-item cap both pass on `53dd82b5`.

Two #392-local contract gaps remain:

1. **A normal multi-batch `embed_documents()` call has no shared outer deadline.** Each provider sub-batch starts a fresh `deadline_budget_s=self.timeout`. With a 100ms configured timeout and two one-item sub-batches, both requests received their own ~100ms budget and the call completed after ~288ms. Either carry one monotonic deadline through the full call or explicitly define/configure this as a per-request budget and stop describing it as operation-wide.
2. **Disabled warmup still performs provider work.** With provider/model configured and `embeddings_enabled=false`, `/lcm embed warmup` called the provider, created/registered the feature profile, and reported `status: ready`. The opt-in flag should fail closed before provider calls/downloads/schema writes, or the operator contract must explicitly define warmup as the enabling action rather than calling disabled mode inert.

Requesting changes only for those boundaries; the single-request retry deadline and hard item cap are fixed.

100yenadmin:
Invariant remediation for E-03 and E-06 provider/deadline facets is published at 66482f37f98beb69c22f66be085fc75172837a44. Provider ambiguity is not automatically resent, hard limits are authoritative, and callback/decode/retry work shares one absolute budget. The repaired stack passed 159 focused embeddings tests, 290 core tests, Ruff, diff and ancestry checks. Independent final embeddings review found no P0-P3 issue at 98% confidence. This is exact-head PR proof only, not merge or release proof; maintainer review state was left unchanged.

Tosko4 (approved):
Rechecked exact head `66482f37` against the previously reported provider deadline, batching, hard-cap, and disabled-warmup blockers.

A public provider operation now shares one absolute deadline across preprocessing, dispatch, retries, decoding, and internal sub-batches; Voyage remains clamped to 1,000 items; and warmup is inert while embeddings are disabled. The exact-head provider suite passed all 47 tests.

100yenadmin:
Follow-up review fix published in e596c9d. Voyage authentication failures no longer count toward the transient circuit breaker, so repeated 401/403 responses remain actionable auth errors. Regression coverage includes three consecutive auth failures plus the existing transient circuit-breaker behavior. The full embedding-focused suite passed (170 tests), Ruff passed, and the repaired stack passed combined temporal/embeddings integration.

100yenadmin:

100yenadmin:

PR #393, opened 2026-07-15
[v1·embeddings 3/5] /lcm embed backfill (dry-run-first)

100yenadmin:
## Summary

- **`/lcm embed backfill`** (design #386 — embeddings train PR 3 of 5, stacked on #392): populates the
  vector store from existing depth-0 summaries, **dry-run first** — the default invocation previews
  counts, batches, and provider-aware cost estimates with **zero provider calls and zero writes**
  (the same dry-run-first shape as #382)
- `--apply`: provider calls happen **outside any write transaction**;each vector row commits
  independently (one failure never loses a batch); Voyage auth errors abort immediately with an
  operator-readable message; transient errors skip-and-report and the next run resumes them
- **single-flight claim**: `BEGIN IMMEDIATE` metadata claim with 10-minute stale takeover, released
  only by the owning invocation (error paths included)
- idempotent by construction (`NOT EXISTS` against the current profile's meta rows); `--limit`
  bounds each invocation; operator-guide section documents the workflow and safe resume
- 9 mocked-provider tests: dry-run purity, cost math, batching/meta, idempotence, per-row isolation,
  auth abort + claim cleanup, transient continuation, over-cap reporting, stale-claim takeover

Design anchor: #386 · Stacked on: #392 (→ #390) · Part of #375

## Validation

- [x] `pytest -q tests/test_embedding_backfill.py` — **9 passed** (119 across the stacked
      provider/vector/command bundle)
- [x] full `pytest -q` — **1650 passed, 1 skipped, 12 xfailed** (canonical + `ulimit -n 1024` low-FD
      runs); 3 failures are the known macOS baseline, rerun and reproduced on untouched upstream
      `main` at `31675c6`
- [x] release validator — all gates pass across the completed packet + a wedge-recovery rerun
      (one full-pytest child hit the known local-load Barrier wedge, was terminated per the
      documented playbook, and every remaining gate was rerun directly: release stress
      `failure_count: 0`)

## Notes

- Merge order: #390 → #392 → this.
- No engine/tool/schema changes — a command only. Search modes are the next car; nothing reads the
  vectors yet.


Tosko4 (changes requested):
The dry-run inventory is useful, and owner-exact lease release is handled correctly. I’d keep that read-only surface.

Apply mode is not ready yet:

- Work discovery happens before ownership is claimed, so the selected batch can be stale by the time the claim is acquired.
- The ten-minute lease has no heartbeat. A second owner can take over while a slow first worker is still active.
- Status starts as `complete`, and ordinary provider/batch failures do not change it. The probe returned `status=complete, embedded=0, failed=1, remaining=1`.
- There is no checkpoint between provider acceptance and `record_embedding()`, so a crash in that gap can resend and rebill a batch.
- There is no operation-wide runtime budget or lease refresh.

Focused cumulative validation passed 46/46.

Smallest safe path: keep dry-run only. Add claim-before-discovery/requery, renewable owner-CAS lease, truthful partial status, operation budgets and resumable publication before enabling `--apply`.

100yenadmin:
@Tosko4 addressed — `fix(embedding-backfill): correct apply-mode` (kept apply-mode rather than gutting to dry-run-only):
- **Claim before discovery:** acquire the lease first, then discover/requery pending rows so the batch can't be stale by claim time.
- **Renewable heartbeat lease:** owner-CAS with periodic refresh; a second owner can only take over a genuinely expired lease, not one being heartbeated.
- **Truthful status:** no longer starts `complete`; reports `partial`/`failed` with accurate counts when any batch fails.
- **Crash-safe, no double-bill:** rows are marked `in_flight` before the provider call and cleared only on `record_embedding`; a crash in the gap leaves them re-discoverable by the idempotent NOT-EXISTS scan without rebilling already-recorded rows.
- **Operation runtime budget** + lease-refresh cadence (config).

Tosko4 (changes requested):
Re-review verdict for exact head `6844cb8d`: **changes still required**.

Claim-before-discovery, owner-CAS lease renewal, truthful partial/failed status, durable inflight marking, and the operation budget are substantive fixes.

The original provider-acceptance publication gap remains. If the provider accepts a batch but `record_embedding()` fails before the vector is committed, the inflight row is rediscovered and automatically sent to the provider again. My probe observed two provider calls for one eventually recorded vector.

That is at-least-once billing, not crash-safe resumable publication. If the provider does not offer an idempotency key, a bounded V1 can report such rows as uncertain and require an explicit retry rather than silently rebilling them.

I’m keeping vector-store issues on #390 and provider request-shape issues off this PR; this verdict is specific to backfill apply semantics.

100yenadmin:
One behavior fix from the automated review landed here (it's backfill behavior, so this PR owns it): `--apply` runs now use their own operation budget instead of tripping the interactive per-minute spend guard — a large backfill no longer stalls after ~60 fast batch calls. The interactive guard is unchanged for query embedding.

Tosko4 (changes requested):
Targeted re-review of exact head `de6aaa5b`: claim-before-discovery, owner-CAS lease renewal, truthful partial status, and durable in-flight markers are all improvements. They do not close the uncertain-publication seam.

I reproduced the exact remaining case with one row: the provider returned an embedding successfully, local `record_embedding()` then failed, and the next apply run sent the same row to the provider again. Across the two runs the provider was called twice; the first reported `failed` with one row in flight, and the second completed and published it.

An in-flight marker makes the work recoverable, but it cannot distinguish “provider never accepted” from “provider accepted/billed, local publication failed.” Retrying the latter can duplicate paid remote work. Please add a durable acceptance/idempotency/reconciliation contract, or explicitly stop and require operator resolution for this ambiguous state rather than claiming crash-safe no-double-bill behavior.

Tosko4 (changes requested):
Additional exact-head finding for `de6aaa5b`: the uncertain provider-acceptance/publication blocker remains, and lease ownership is also not revalidated after a blocking provider call.

I let the provider call outlive the lease and had a successor acquire it. The old worker then returned from the provider, published the embedding, and reported `status: complete`; the successor’s claim remained present. Renewal before each synchronous batch does not prevent a stale owner from writing after the call.

Please add an owner/generation CAS after provider return and before every vector/in-flight publication, with an independent heartbeat where the provider supports it. This is separate from the already-reported remote-acceptance ambiguity: even a successful response must not be published by a worker that no longer owns the lease.

100yenadmin:
Addressed (`fix(embedding-backfill): re-check lease ownership after provider return; publish under captured identity`): after the provider returns and before every publication, ownership is re-checked via owner CAS (owner id is a fresh uuid per acquire) — your stale-owner repro now discards the result and does not publish or release the successor's claim. Vectors publish under the identity captured before the provider call (ties into #390's write CAS).

Tosko4 (changes requested):
The stale-owner and captured-identity repairs pass on `d4aac828`: after a successor acquired the lease during the blocking provider call, the old worker published zero rows, reported `lease_lost`, and left the successor's claim intact.

The backfill still has an ordinary remote-acceptance/publication ambiguity:

- when the provider returned a vector and local `record_embedding()` failed, the next run resent the same document, producing two provider calls for one final row;
- when a multi-request provider batch accepted its first sub-batch and the second failed, retry resent the already accepted first sub-batch because no partial checkpoint survives;
- `for_backfill=True` bypasses the interactive spend guard but still applies `embedding_query_timeout_s` (3 seconds by default) to each document request, while the separate backfill operation budget defaults to unlimited. A normal slow bulk request can therefore fail on a query-latency setting;
- dry-run and apply use different Voyage safety thresholds. A 25,000-token document was counted as one estimated batch and 25,000 cost tokens, while apply skipped it before any provider call because its effective document budget is 24,300 tokens.

Please give backfill its own explicit request/deadline policy, make dry-run use the exact apply eligibility/batching rules, and persist provider acceptance at a granularity that can be reconciled or retried idempotently. If a provider cannot support that, surface an operator-visible uncertain state instead of automatically resending accepted work. Requesting changes for those remaining backfill guarantees; the post-call lease and identity fixes pass.

100yenadmin:
Invariant remediation for E-02 and E-03 backfill/publication facets is published at 13a4e5a3291559288f49eaeb8f1136d5188979bd. Publication now validates immutable identity and lease ownership transactionally; uncertain remote acceptance remains quarantined and explicit retry authorization is reserved for the exact selected IDs. The repaired stack passed 159 focused embeddings tests plus 33 targeted final-review cases. Independent final embeddings review found no P0-P3 issue at 98% confidence. This is exact-head PR proof only, not merge or release proof; maintainer review state was left unchanged.

Tosko4 (changes requested):
Rechecked exact head `13a4e5a3`. The important publication work is fixed: accepted sub-batches are yielded and persisted independently, ambiguous remote acceptance is quarantined as `uncertain`, retry requires explicit authorization for the exact selected rows, and publication validates lease plus immutable identity transactionally.

One normal bulk-path blocker remains. `resolve_provider(..., for_backfill=True)` still gives Voyage, Ollama, and FastEmbed the interactive `embedding_query_timeout_s`, which defaults to 3 seconds. The backfill lease and operation budget do not alter that provider deadline, so an ordinary non-trivial backfill or local model load can fail every batch against a query-latency budget even though bulk work has its own lifecycle and budget.

Please give backfill/document embedding its own timeout contract, or explicitly thread the bulk budget into provider resolution. Keep the interactive query deadline separate. The exact-head backfill suite passed 33 bounded tests; the 250,001-row maintenance stress case was intentionally not repeated in this pass.

100yenadmin:
@Tosko4 addressed on exact head 1ea76d82540cc0684d303b7827cdfe37829d24e9. `resolve_provider(..., for_backfill=True)` now uses a dedicated bulk/document-operation deadline (`embedding_backfill_timeout_s`, default 120s; `LCM_EMBEDDING_BACKFILL_TIMEOUT_S` override) instead of the 3s interactive query deadline. Interactive query behavior is unchanged. The exact scaled repro that previously failed after 3.010s now succeeds after 3.102s under the bulk deadline. Regressions cover Voyage, Ollama, and FastEmbed resolution plus a slow local document encode. Validation: 163 embeddings tests, 290 core tests, deterministic recall unchanged, Ruff/diff checks clean, and 294 combined temporal+embeddings tests. #394 and #395 were re-stacked mechanically at 11a9325 and 6e23bfd; #395 documents the separate bulk/query contracts. Please re-review the new exact head when convenient.

Tosko4 (approved):
Rechecked exact head `1ea76d82`. This closes the remaining bulk-path blocker.

`resolve_provider(..., for_backfill=True)` now uses the dedicated `embedding_backfill_timeout_s` contract, defaulting to 120 seconds and configurable through `LCM_EMBEDDING_BACKFILL_TIMEOUT_S`. The interactive query timeout remains separate at 3 seconds. The bulk deadline is wired across Voyage, Ollama, and FastEmbed, with provider-resolution and slow local document-encode regression coverage.

Validation: 85 provider/backfill tests passed, with the previously identified 250,001-row maintenance stress test intentionally excluded. Ruff, `py_compile`, and the delta diff check passed.

Approving.

100yenadmin:
Follow-up review fixes published in 702ed73. First-run in-flight schema inspection now occurs after BEGIN IMMEDIATE, closing the concurrent DDL race, and both embedding purge paths remove matching durable in-flight rows. Deterministic two-connection and standalone/shared-transaction purge regressions pass. The full embedding-focused suite passed (170 tests), Ruff passed, and the repaired stack passed combined temporal/embeddings integration.

100yenadmin:

PR #394, opened 2026-07-15
[v1·embeddings 4/5] semantic + hybrid lcm_grep (RRF)

100yenadmin:
## Summary

- **`lcm_grep` gains `semantic` and `hybrid` modes** (design #386 — embeddings train PR 4 of 5,
  stacked on #393); `full_text` stays the default and its serialized output is **byte-for-byte
  identical** (regression-tested)
- semantic: query embedded under a **hard wall-clock budget** (`LCM_EMBEDDING_QUERY_TIMEOUT_S`, 3 s)
  → KNN via the #390 ladder, cosine **confidence bands** (≥0.65/0.5/0.35) and store `coverage`
  surfaced per result; timeout/transient/missing-provider → **degrades to full-text with an explicit
  flag, never an error**; auth errors surface readably
- hybrid: FTS ∪ semantic, node-id dedup, **reciprocal-rank fusion (k=60)** — no external reranker,
  no new dependency in this car (Voyage rerank-2.5 noted as a possible later enhancement)
- **the proof artifact ships in-tree**: `scripts/eval_retrieval_recall.py` + 30 committed stratified
  queries (exact-term / paraphrase / multi-hop-ish) over a fixed-seed corpus with a deterministic
  mock embedder — offline, no network; a test asserts byte-for-byte determinism and that hybrid
  recall ≥ FTS recall on paraphrases. Synthetic smoke result: FTS paraphrase recall\@5 **0.100** vs
  semantic/hybrid **1.000** (smoke values, not provider benchmark claims; live-provider runs are
  env-double-gated)

Design anchor: #386 · Stacked on: #393 (→ #392 → #390) · Part of #375

## Validation

- [x] focused mode/provider/vector/schema bundle — **50 passed** (11 new)
- [x] full `pytest -q` — **1661 passed, 1 skipped, 12 xfailed** (normal + low-FD validator gates);
      3 failures are the known macOS baseline, reproduced on untouched upstream `main`
- [x] release validator — all gates pass, release stress `failure_count: 0`, no wedge this run
- [x] `full_text` byte-identity regression across feature on/off

## Notes

- Merge order: #390 → #392 → #393 → this.
- Zero cost when unconfigured; local providers make both modes $0; the docs car (next, final)
  carries the full free/local setup guide.


100yenadmin:
Fix round (adversarial self-review) — pushed `fix(embedding-search): degrade + latency`:
- **coverage='none' now degrades to FTS**: a user with embeddings enabled but no backfill yet (no profile) got EMPTY results in semantic/hybrid instead of falling back. Now coverage='none' / empty candidates degrade to full-text with the `degraded_to_fts` flag; hybrid returns its FTS arm rather than fusing against nothing.
- **no leaked non-daemon thread**: the query-embed timeout previously left a non-daemon worker running (up to retries), which concurrent.futures' atexit hook would join and could hang process exit. The provider call is now bounded directly and any helper thread is a daemon.
- **provider cached on the engine** by (provider, model) so the circuit breaker + spend guard actually accumulate across calls (were reset every call). `_lcm_grep_full_text` verified byte-identical (SHA-256). 15 search tests.

Tosko4 (changes requested):
This pass improves the failure behavior: missing coverage now degrades explicitly, provider instances are cached, built-in interactive HTTP paths use one attempt, and timed-out workers no longer block interpreter shutdown. The committed offline recall evaluation is also useful evidence for the retrieval direction.

Two correctness contracts still block this slice:

- Semantic search parses source, conversation, role and time filters, but KNN receives only `time_from` and session ID. Role/conversation are echoed rather than enforced, while source and `time_to` are applied after the KNN cap. Current-head probes returned the wrong conversation, returned an assistant-backed summary for `role=user`, and degraded after discarding a top out-of-scope hit even though a lower eligible vector existed.
- The configured query timeout still covers only query embedding. Vector-store construction and KNN are synchronous outside that budget. A configured `0.02s` timeout with a deliberately slow KNN returned after `0.12s`. Daemon threads avoid shutdown blockage, but they do not cancel provider work; repeated timeouts can accumulate live workers.

Lower vector-store findings remain owned by #390. Focused search/provider/vector/backfill tests passed 61/61, and the offline evaluation still reproduces semantic/hybrid paraphrase recall@5 of 1.0 versus FTS 0.1.

Smallest safe path: either reject unsupported filters/scopes explicitly for V1, or enforce eligibility before ranking. Use one absolute deadline across provider, KNN, fallback and fusion with bounded worker ownership.

100yenadmin:
@Tosko4 addressed — `fix(embedding-search): enforce filters before ranking + one absolute deadline`:
- **Filters enforced before ranking:** conversation, source and time (`from`+`to`) are now passed into the KNN filter and applied BEFORE the top-k cap, so ineligible vectors can't consume a slot ahead of an eligible lower one, and there are no wrong-conversation hits. `role` degrades to the full_text path (which enforces role) rather than being echoed-and-ignored — so no filter is silently dropped; it's enforced by the arm that can.
- **One absolute deadline** across embed + KNN + fallback + fusion (previously only query-embed was bounded; KNN was outside it). A slow KNN under a tiny total budget now degrades to FTS within budget, and worker ownership is bounded so repeated timeouts don't accumulate live workers.
- `_lcm_grep_full_text` output verified byte-identical (sha256).

Tosko4 (changes requested):
Re-review verdict for exact head `4ef78abd`: **changes still required**.

There are good fixes here: role-scoped semantic queries now degrade to the exact full-text path, source/time eligibility is pushed into KNN before ranking on the full path, and worker capacity is bounded.

Three semantic-contract blockers remain:

- `conversation_id` is reduced to session IDs. In a session containing multiple gateway lanes, a query for lane A returned a summary sourced only from lane B.
- The public schema still says broader scopes and `time_from`/`time_to` return raw-message hits only. Semantic mode nevertheless returns summary nodes for those requests, including cross-session nodes that current-session expansion rejects.
- The advertised absolute operation deadline is not end-to-end. With a 20 ms setting and deliberately slow KNN, the call degraded only after about 122 ms.

Provider/model identity mixing is reproduced too, but I assigned that substrate/API issue to #390 rather than charging it twice here.

100yenadmin:
Contract-consistency fixes from the automated review (aligns with your review's 'enforce or reject explicitly'): the semantic arm now honors the advertised raw-only `lcm_grep` contract — `session_scope='all'`, explicit non-current sessions, and `time_from`/`time_to` queries degrade the semantic arm to the raw full-text path (explicit flag, never silently summary-backed cross-session hits), and conversation filtering no longer leaks wrong-lane hits in multi-conversation sessions. Two round-2 tests that encoded the contradictory behavior were rewritten to assert the contract.

Tosko4 (changes requested):
Targeted re-review of exact head `2439e471`: the raw-only scope contract and lane handling are now explicit and correct. Role, broader-session, time, and conversation-scoped semantic requests degrade to the full-text path that can enforce those dimensions; source eligibility is applied before vector ranking.

The end-to-end deadline blocker remains. The absolute deadline covers query embedding and KNN, but `degraded()` runs full-text search after that deadline with no remaining-budget check or bound. With a **0.02s** configured budget, a KNN that exceeded it and a synthetic **0.08s** fallback produced a correct degraded result after **0.171s**. Hybrid also performs its full-text arm before starting the semantic deadline.

Please either define the budget narrowly as semantic-arm-only and remove the end-to-end/degrade-within-budget guarantee, or enforce one deadline across the full request, including fallback and hybrid orchestration.

100yenadmin:
Addressed (`fix(embedding-search): document the semantic-arm-only latency budget`) — I took your first option (narrow the contract) deliberately: the full_text fallback runs synchronously on the shared, **uncancellable** SQLite connection, so a hard end-to-end cutoff isn't safe and 'degrade within budget' was self-contradictory. The budget is now defined as **semantic-arm-only** (query-embed + KNN), the overpromising comment/docs are corrected, and the semantic attempt stays deadline-bounded and tested. If you'd rather I enforce end-to-end instead, that'd mean making the fallback interruptible (a separate connection + cancellation) — happy to, just a bigger change.

Tosko4 (commented):
Verified on `f9b81c2c`.

The contract is now explicit and coherent for this slice: one absolute deadline bounds the semantic arm (query embedding + vector-store construction + KNN), and a timeout abandons that arm before the synchronous full-text fallback runs to completion. I agree with narrowing the contract instead of claiming the current shared-connection FTS path is cancellable.

I found no remaining #394-local orchestration blocker. Because this PR is cumulative from `main`, it remains blocked by the inherited #390/#392/#393 gates rather than receiving a full approval.

100yenadmin:
Invariant remediation for E-05 whole-request search-deadline facets is published at 61dc3f5561074bd6a2452e78ce9d24d8df7a0781. The entry deadline now covers resolution, query embedding, bounded KNN, hydration, FTS fallback, both hybrid arms, and fusion, with no later stage starting after expiry. The repaired stack passed 159 focused embeddings tests, 290 core tests, Ruff, diff and ancestry checks, and deterministic recall with no regression. Independent final embeddings review found no P0-P3 issue at 98% confidence. This is exact-head PR proof only, not merge or release proof; maintainer review state was left unchanged.

Tosko4 (commented):
Rechecked exact head `61dc3f55` against the prior search-deadline and degradation blockers.

The one request deadline now covers provider resolution, query embedding, bounded KNN, hydration, FTS fallback, both hybrid arms, and fusion. Raw-only filters degrade to the exact full-text path instead of being silently ignored.

No blocker remains in this local slice. The cumulative PR remains blocked by the owning backfill timeout issue in #393.

Tosko4 (approved):
Rechecked exact head `11a9325b`. The #394 incremental patch has the same stable patch ID as the previously reviewed local slice, and the parent #393 bulk-timeout blocker is now resolved.

The cumulative embeddings validation at the refreshed #395 tip passed 161 bounded vector/provider/backfill/search tests. Two previously identified very large stress cases were intentionally excluded; Ruff and diff checks passed.

Approving.

100yenadmin:
Follow-up review fixes published in c3dbd1a. Semantic capacity exhaustion now has an independent bounded FTS worker lane; hybrid preserves already-computed FTS results when the semantic arm times out while auth errors remain operator-visible; successful warmup replaces stale provider cache state. Exact fallback/cache regressions, the 83-test provider/search suite, the 170-test embedding-focused suite, Ruff, deterministic recall, and combined temporal/embeddings integration all passed.

100yenadmin:

100yenadmin:
**Re: `content_scope='externalized'` × `mode='semantic'/'hybrid'` (combined-head contract question from the tracker)**

You were reading it right — and it needed fixing, not just documenting. On the combined head, neither the semantic nor the hybrid arm referenced `content_scope` at all, so `content_scope='externalized'` with `mode='semantic'` silently ignored the requested scope and returned history-only semantic hits. (The combination is unreachable on either train's own head — `content_scope` comes from the externalized-payload-search PR (#383), the `mode` dispatcher from this one — so it's a pure integration seam.)

**To your sub-question ("do they require embeddings to actually work?"):** externalized payloads are *never* embedded — `lcm_embedding_meta.embedded_kind` is CHECK-constrained to `'summary'` — so there is no vector corpus a semantic externalized-scope search could run against, regardless of whether embeddings are enabled/backfilled. "Threading the scope through the semantic arm" therefore can't produce semantic payload hits today; it would require payload embeddings, a separate feature.

**Contract chosen: degrade, not reject.** The semantic arm already refuses to approximate filters it cannot honor exactly (`role`, `time_from/to`, `conversation_id`, broader `session_scope`) by degrading to `full_text` with an explicit `degraded_reason`. Non-`history` `content_scope` now follows the same rule:

- `mode='semantic'` + `content_scope='externalized'|'both'` → degrades to the full-text arm (which owns bounded payload scanning), with `degraded_to_fts: true` and `degraded_reason: "content_scope beyond history is served by full_text retrieval"`.
- `mode='hybrid'` → the semantic arm's degrade surfaces as the full-text-arm result — which honors `content_scope` on the combined head — plus the same explicit marker. No fused history-only semantic hits can leak into a payload-scoped request.

Landed as `cef6d34` on this branch with focused regressions (`test_semantic_content_scope_degrades_to_full_text`, `test_hybrid_content_scope_degrades_to_full_text_arm` — the modes suite is 32/32), and propagated to #395 (`a2c7001`) and the combined release-test build (`727e19c` on the fork's `release-test-2026-07-17`/`main`).

Live verification on a real host against an actual externalized payload: all three calls (`semantic+both`, `hybrid+externalized`, `full_text+externalized` control) now return the full-text `externalized_scan` diagnostics block, with the two degraded modes carrying the explicit marker — and the payload scan itself still enforces #383's active-session-only rule (`rejected_session_mismatch` reported rather than silently matched).

Hard reject remains available if you'd rather fail loudly than degrade, but degrade keeps `content_scope='both'` usable in one call and matches every other unhonorable-filter path in the arm. Happy to switch it if you prefer.

100yenadmin:

PR #395, opened 2026-07-15
[v1·embeddings 5/5] free & local embeddings setup guide

100yenadmin:
## Summary

- **`docs/embeddings-setup.md`** — the operator guide for the embeddings feature (design #386 —
  **final train car, 5 of 5**, stacked on #394): three provider walkthroughs with copy-paste config —
  **Voyage free tier** (200M free tokens per model group on the Voyage-4 generation + rerank-2.5, no
  credit card; verified 2026-07 with their pricing page linked as source of truth), **fastembed** as
  the pip-only no-signup local default (with an honest model-download-size note and the
  explicit-warmup workflow), and **Ollama** for users already running it
- cost table (hybrid ≈ $0.001/query on Voyage past free tier; $0 local; $0 when off), performance and
  footprint numbers from the design benchmark, the coverage/degrade semantics in plain operator
  language, and provider-switch → re-backfill semantics
- cross-linked from `docs/retrieval-tools.md`

We maintain Martian Engineering's Lossless Claw and are porting the proven pattern into Hermes-LCM.

Design anchor: #386 · Stacked on: #394 (→ #393 → #392 → #390) · Part of #375 · **Completes the embeddings train**

## Validation

- [x] docs-only diff (`docs/embeddings-setup.md` new + one link line); no code paths touched
- [x] every factual claim sourced: Voyage numbers verified against docs.voyageai.com 2026-07 (pricing
      page linked as the source of truth in-text); performance numbers are the #386 benchmark's;
      command examples match the shipped `/lcm embed warmup|backfill` and `lcm_grep` mode surfaces
- [x] all internal links resolve within the stacked tree

## Notes

- Merge order: #390 → #392 → #393 → #394 → this.
- We'll re-verify the Voyage free-tier numbers against their pricing page at whatever time this
  merges and refresh if they've moved.

PR #401, opened 2026-07-17
[v2 1/9] shared retrieval core (enabler)

100yenadmin:
Part of the recall train (tracker: #375, "forever memory" follow-up to the #380–#395 wave). **Stacks on the wave** — cumulative from main like the prior trains; the slice diff for review is vs the wave head (`docs/embeddings-setup-guide` + subsequent fixes).

**This slice:** pure enabler refactor. Extracts the retrieval plumbing shared by `lcm_grep` (and the upcoming `lcm_recall`) into `retrieval_core.py`: RRF fusion, KNN invocation + node hydration, deadline helpers, snippet shaping. `tools.py` keeps thin wrappers, so every public contract, JSON shape, degrade path, and error string is byte-identical — full suite pass/fail set is unchanged vs base, verified by diffing sorted FAILED lists, not counts.

Zero behavior change; exists so two search surfaces never grow divergent ranking implementations.


100yenadmin:

PR #403, opened 2026-07-17
[v2 3/9] lcm_recall — cross-conversation memory search

100yenadmin:
Recall train slice 3 — **the feature** (stacks on chunk-vectors; tracker #375).

**`lcm_recall`**: search the agent's entire memory across ALL conversations and all time by meaning. Three arms over the whole local DB — FTS (raw, all sessions) + summary KNN + chunk KNN (both unfiltered) — fused with RRF (chunk hits dedupe against FTS by store_id), optional `rerank-2.5-lite` stage (`LCM_RERANK_ENABLED`, default OFF pending our benchmark harness), then a soft prior: `final_score = rank × (1 + scope_bias·is_current_conversation) × recency(half-life 30d, floor 0.5)`. Scope/recency are ranking boosts, never filters. Honest degrade matrix mirrors `lcm_grep` (embeddings off → FTS arm with explicit reason). `lcm_grep` is untouched except one additive forward-pointer sentence; its param surface, degrade cells, and tests are byte-identical.

Design decision record: extending `lcm_grep`'s scope semantics was adversarially debated and rejected — it would reverse the reviewer-verified degrade cell (semantic × scope=all → raw-only) and jam soft-prior semantics into a hard-filter tool. A cold-agent reach-for test (8 scenarios) confirmed the dedicated tool wins exactly the "have I ever discussed…" class while grep keeps exact/known-range lookups; the description's not-for clauses come from that test.

One load-bearing find during acceptance: the summary arm inheriting `embedding_bounded_scan_rows=2000` made older archive vectors *structurally unreachable* (recency-bounded enumeration). Recall gets its own `recall_scan_rows` (default 25,000, deadline-guarded); regression-locked; grep unchanged.

Live proof on a real 347k-message / 3,872-summary imported archive, fresh session, real host: 8 hits, 7 cross-session over 5 distinct conversations, both canonical months-old memories recalled by a paraphrase query. 14 new tests; full-suite failure set identical to base.


100yenadmin:

PR #404, opened 2026-07-17
[v2 4/9] LongMemEval retrieval-quality harness

100yenadmin:
Recall-train companion (independent slice; stacks on the wave; tracker #375).

**LongMemEval retrieval harness** in the repo's benchmark culture: offline-first (explicit `fetch` step, HF dataset `xiaowu0162/longmemeval` PINNED at revision `2ec2a55`), retrieval metrics only (R@1/5/10, NDCG@10 against labeled evidence sessions — no LLM judge), deterministic stub / local-fastembed / voyage provider arms, aggregate-only exports.

Purpose: gate the recall train's quality knobs with OUR numbers instead of vendor claims — which arm wins (FTS / summary vectors / chunk vectors / hybrid RRF / +rerank), whether the reranker default earns its latency, and which chunk policy to recommend. First local numbers (25q, bge-small): FTS R@10 = 0.60; summary/hybrid vector arms R@10 = 0.84–0.88. README documents the MemDelta (arXiv:2606.29914) caveat: retrieval-benchmark verdicts flip with embedding/base-model choice, so we publish exact configurations.

16 tests; benchmarks/README section; full-suite failure set identical to base.


100yenadmin:

PR #405, opened 2026-07-17
[v2 5/9] non-blocking FTS integrity scan on bind

100yenadmin:
Operational fix (independent slice; stacks on the wave; tracker #375).

**Non-blocking FTS integrity scan on session bind.** Profiled on a 986MB/347k-row production-import DB: 100% of the cold-bind cost is the FTS5 `integrity-check` special command (the code's own comment cites #235); the existing 24h marker throttle helps, but every cache-miss (first bind + each expiry) blocked the bind synchronously — measured 10.3s warm-cache, minutes cold-cache.

Fix: on a throttled cache-miss, the cheap structural check stays synchronous (correctness gate), the deep scan moves to a daemon thread with its OWN connections (FTS5 integrity-check needs a writable handle — savepoint-rollback keeps it logically read-only; a separate brief connection stamps the marker). Corruption found in background is never auto-rebuilt: a persisted `fts_integrity_failed:<table>` flag surfaces through both doctor paths pointing at the explicit `/lcm doctor repair apply`. One scan per (db, table) with staleness unwedging; `LCM_FTS_INTEGRITY_BACKGROUND=false` restores the exact old behavior.

Measured: bind 10.257s → 1.721s on the same DB. +6 tests (dispatch/stamp, corruption→flag, kill-switch, single-flight, stale unwedge); sync-assertion tests pinned with the kill-switch, assertions unchanged.


100yenadmin:

PR #406, opened 2026-07-17
[v2 6/9] interim schema-stamp remediation

100yenadmin:
Operational fix (independent slice; stacks on the wave; tracker #375).

**Interim-build schema-stamp detection + guided remediation.** DBs touched by interim development builds carry `schema_version=6` with a v5 core and early-variant feature tables; current builds refused them with guidance ("upgrade the plugin") that is wrong for this case.

- `classify_version_mismatch()`: `interim_stamp` = core v5 shape matches EXACTLY and every extra table belongs to a known feature family — feature-table internal shape is deliberately not certified by the numeric stamp (each feature's marker-gated init owns that). Anything else stays `genuinely_newer` (refuses, safe direction).
- Refusal message now branches on the classification and names the remediation.
- `/lcm doctor repair schema-stamp [apply]`: dry-run default, backup-first, drops only family caches that FAIL their final-shape verifier (derived data — rollups rebuild from the DAG, vectors re-backfill; each drop reported with its rebuild hint), then resets the stamp. Refuses `genuinely_newer`.

Hard acceptance on a real interim-stamped operator DB (independently re-verified): classify → 6 derived-cache drops → reopens clean at v5 → RollupStore + VectorStore construct, verifiers clean. 13 tests incl. end-to-end recovery; full-suite failure set identical to base.


100yenadmin:

PR #407, opened 2026-07-17
[v2 7/9] proactive memory injection (default-off)

100yenadmin:
Recall train slice 4 — the "it just has it" surface (stacks on #403; tracker #375).

**Proactive memory injection** (default-OFF: `LCM_PROACTIVE_RECALL_ENABLED`): at active-context assembly, LCM embeds the newest user message, runs the `lcm_recall` pipeline (k=6, scope_bias 0.3 — zero duplicated ranking/deadline logic), and injects one bounded `<relevant-memories>` block (≤ `LCM_PROACTIVE_RECALL_BUDGET_TOKENS`, default 500) with 1-3 labeled memory hints + `lcm_expand` handles. Pull-based memory tools can't deliver "remembers without being asked" — agents don't call memory tools when they don't know a memory exists; assembly-time injection is the mechanism the current top LongMemEval systems share.

Safety posture: block placed between the cache-stable summary prefix and the volatile fresh tail (summary-prefix caching intact); wrapped in an `_INJECTED_CONTEXT_TAGS` tag so `strip_injected_context_blocks` removes it before ingest — injected memory never re-enters the lossless store (tested); relevance floor + dedupe vs active context; any failure or timeout injects nothing and never blocks assembly; telemetry counters in `lcm_status`. `LCM_PROACTIVE_RECALL_PROVIDER` allows a cheap/local embedder for the injection path independent of the search provider.

One reviewer-worthy judgment call: `LCM_PROACTIVE_RECALL_MIN_SCORE` defaults to 0.01 on the RRF scale (single-arm RRF scores sit ~0.014-0.016); with rerank enabled scores become [0,1] relevance and the operator guide documents raising it (~0.3).

+14 fixture tests over the real recall pipeline; full-suite failure set byte-identical to base; ruff clean.


100yenadmin:

PR #409, opened 2026-07-17
[v2 8/9] adversarial-sweep: storage integrity + data-safety

100yenadmin:
Adversarial-review fix batch (storage + data-safety), stacked on the recall train (tracker #375). An 81-agent adversarial sweep (6 review dimensions → two-skeptic verification per finding → judge) confirmed 26 findings on the combined head; this PR fixes the storage/safety set, one commit per finding, each with the verifier's reproduction as its regression test:

- **Schema-stamp classifier could DROP tables on a genuinely-newer DB** (extra/unknown column now = `genuinely_newer`, never dropped; the real interim-DB remediation path verified unregressed — it has renamed, not extra, columns)
- `lcm_chunk` family registered in the schema-stamp verifier map (was allowlisted but never verified/repaired)
- `fts_integrity_failed` flag now cleared by successful explicit repair (was a permanent false-positive)
- Explicit repair joins in-flight background scans; integrity errors classified as corruption only on corruption signatures (lock/busy → `unchecked`)
- Background-scan dispatch stamps `scan_started_at` before `thread.start()` (cross-process duplicate-scan window)
- GC content-rewrite and chunk archival now atomic per message (stale-offset slicing window closed)
- `--retry-uncertain` chunk recovery persists real char spans (+ `char_end<=char_start` rejected at the write seam)
- **Chunk backfill consent gate**: `--corpus chunks|both --apply` on cloud providers requires `--confirm-raw-text`; docs now state plainly that the chunk corpus sends raw verbatim text (unlike summaries) and that ingest-time redaction is not retroactive
- `--corpus both` next-hint coherent; `--corpus`/`--policy` documented in the operator guide

Full-suite failure set identical to the environmental baseline; ruff clean. One sweep finding (per-vector commit amplification in backfill publishing) is deferred with a coordination note — it needs a batch-publish seam in vector_store.py and is perf-only.


100yenadmin:

PR #410, opened 2026-07-17
[v2 9/9] adversarial-sweep: lcm_recall ranking + perf

100yenadmin:
Adversarial-review fix batch (recall ranking + performance), stacked on the recall train (tracker #375). Fixes the ranking/perf set of the 26 sweep-confirmed findings, one commit per finding with reproductions as regression tests:

- **Ranking quartet** in `lcm_recall`: chunk-vs-FTS dedupe kept the worst-ranked chunk (dict-comprehension last-wins → best-wins); RRF double-counted multi-chunk messages (per-arm identity collapse before fusion); merged hits mixed one arm's snippet with another's offsets (better-ranked arm now supplies both); rerank window was selected pre-prior and spliced voyage scores onto the RRF scale (window now post-prior; rerank is a pure reorder, documented in provenance)
- `scope_bias` boost survives session rotation (conversation-scope session set, not single-id equality)
- Bounded chunk-corpus truncation now surfaces in `degraded_reasons` with scanned/total (was silent at 25k of 93k-261k)
- Pooled `VectorStore` instances (LRU, data_version-invalidated) so matrix caches survive across calls — back-to-back recall 44.4ms → 20.6ms measured; bounded LRU caches replace clear-on-miss
- Batched chunk hydration (one JOIN, was N+1); no-op migration-marker writes skipped; dedicated `LCM_RECALL_QUERY_TIMEOUT_S` (8s) so the 3-arm tool stops inheriting the single-arm 3s budget
- Test-gap closures: direct JSON `lcm_doctor` background-flag test; `VoyageProvider.rerank` wire tests; `include='summaries'` + embeddings-off regression (FTS arm now runs regardless of `include` when embeddings are unavailable, doc updated); expand-handle docs distinguish `lcm_expand` vs `lcm_load_session` per hit kind

Full-suite failure set identical to the environmental baseline; ruff clean; live cross-session recall smoke on a real 347k-message archive copy re-verified post-fix.


100yenadmin:

PR #423, opened 2026-07-21
feat: add grounded answer-ready recall foundation

100yenadmin:
## Summary

- Add opt-in `lcm_recall(detail="answer_ready")` output that hydrates bounded exact-source windows for raw-message hits. Omitted `detail` and explicit `detail="snippets"` retain the existing response bytes.
- Add immutable exact references binding session, message, source span, and content digest. References fail closed after deletion, source drift, or content mismatch.
- Keep observation time, event occurrence time, write time, and explicit question/as-of time separate in storage and tool output.
- Add dependency-free `lcm_compute` operations for grounded date, count, compatible-unit arithmetic, difference, and ordering questions. Ambiguous, incomplete, stale, or ungrounded input returns evidence-only fallback instead of a computed claim.

## Why

The ordinary recall path can retrieve useful evidence, but it does not currently offer a bounded, immutable grounding contract for deterministic downstream reasoning. This change adds that contract as an explicit opt-in while preserving the existing default recall API and bytes.

## Compatibility and safety

- Default `lcm_recall` output has a byte-for-byte regression test.
- Invalid detail modes and stale, deleted, or content-mismatched references fail closed.
- Relative dates are anchored only to persisted observation time. Caller-supplied anchors must match, and unavailable observation time remains `null` rather than becoming an invented epoch.
- Host-supplied candidate answers are accepted only when they exactly equal the canonical deterministic answer, preventing extra narrative claims from entering a verified result.
- Database migration is additive: nullable observation-time columns are added, legacy rows keep their write timestamp, and no observation time is invented.
- Synthetic tests cover legacy migration, backup/restore, source deletion and invalidation, profile rebinding, rotation, and schema-stamp remediation without using a live database.
- Secret-redacted content cannot be reconstructed through exact references.

## Review guide

1. `tools.py` and `schemas.py`: opt-in recall contract and unchanged default response path.
2. `exact_refs.py`: immutable source binding and fail-closed validation.
3. `occurrence_time.py` plus storage changes: separation of observed, occurred, ingested, and question/as-of time.
4. `reasoning.py`: dependency-free grounded computation and ambiguity fallback.
5. `tests/test_v1_upstream_foundation.py`: compatibility, provenance, migration, invalidation, secret-safety, and computation contracts.

## Validation

- [x] Focused foundation suite: 117 passed.
- [x] Tool-registration regressions that previously exposed stale expectations: 4 passed.
- [x] Both affected registration/packaging test files: 36 passed.
- [x] Canonical CI on `37cfcfd0442ba979d602ae86e898663621dfba28`: all six jobs passed.
  - Python 3.11, 3.12, 3.13, and 3.14: 2,296 passed and 12 expected failures per matrix job.
  - Repository-wide Ruff lint passed.
  - Workflow lint passed.
  - Compile, shell syntax, and `git diff --check` static validation passed.
- [x] Current-head check annotations: none.
- [ ] `scripts/validate_release.sh --full`: not duplicated locally; the repository's canonical PR CI completed the repository-wide pytest matrix, lint, and static checks above.

Canonical CI: https://github.com/stephenschoettler/hermes-lcm/actions/runs/29822591430

## Benchmark evidence and claim boundary

The only raw historical result remains 440/500 on product `ffb11bc6674b6435eb32a9d79a4d9b708d770d03` with harness `bcec42f3176aac03d873fedd604b813b2edd72b1`.

A separately isolated, one-shot V1 reproduction scored **441/500 (88.20%)** using:

- this PR's product behavior at `12b39c22fc695ded1eb999860963af9abe461536`;
- MemoryBench harness `5d07dc179b0dd1aec5ca355bf189a3c7251dd8bb`;
- validation-only adapter overlay SHA-256 `a38843ccbd5bc2f49aaa2b622cfc085c6e837cb360f481915ffed531aa74a7b2`;
- Voyage `voyage-context-3` retrieval;
- Codex CLI `0.144.6`, `gpt-5.6-sol` medium/priority answerer and low/priority judge;
- one attempt per question and zero retries.

The current head `37cfcfd0442ba979d602ae86e898663621dfba28` differs from that scored product head only by two test expectation insertions for the already-registered `lcm_compute` tool. The validation adapter is not part of this PR. Therefore, 441 is evidence for the exact combined source and protocol above, **not a product-only score claim** and not evidence of deterministic improvement over historical 440.

The predeclared recovered-question smoke scored 48/49 and remains recorded as a failed smoke gate. The full 500 proceeded only after an explicit owner override; neither run was tuned or retried.

## Deliberate scope boundary

This PR does not include assertion extraction/lifecycle, query views, adaptive or stateful controllers, pre-answer hooks, broad evidence compilers/selectors, clause adoption, structural relation compilation, semantic handles, source frames, session capsules, grounded revision flows, benchmark-only bridges, or the validation adapter used by the reproduction.

Merge, release, installation, deployment, and runtime behavior remain separate gates.


100yenadmin:
V1 score-preservation and CI status:

- The owner-authorized, one-shot full V1 run scored **441/500 (88.20%)**, clearing the predeclared `>=440` threshold.
- That run used product head `12b39c22fc695ded1eb999860963af9abe461536`, MemoryBench harness head `5d07dc179b0dd1aec5ca355bf189a3c7251dd8bb`, and validation-only adapter overlay SHA-256 `a38843ccbd5bc2f49aaa2b622cfc085c6e837cb360f481915ffed531aa74a7b2`.
- The recovered-question smoke remains recorded as a failed 48/49 gate. The full run proceeded only under an explicit owner override; neither run was retried or tuned.
- PR #423 now points to `37cfcfd0442ba979d602ae86e898663621dfba28`. The only change after the scored product head is a two-line test expectation fix adding the already-registered `lcm_compute` tool to two `EXPECTED_LCM_TOOLS` sets. No product behavior, API, schema, storage, benchmark code, or documentation changed.
- Canonical CI on `37cfcfd0442ba979d602ae86e898663621dfba28` is green across lint, workflow lint, and Python 3.11-3.14.

The 441 score belongs only to the exact combined product, harness, validation overlay, fixture, models, and protocol used for that run. It is **not attributable to this product PR alone**, is not evidence of deterministic improvement over the historical 440, and does not prove merge, release, installation, deployment, or runtime readiness.

Update: PR #423 is now ready for maintainer review. Independent approval is still required before merge.

100yenadmin:
@stephenschoettler This PR is now ready for maintainer review at exact head `37cfcfd0442ba979d602ae86e898663621dfba28`.

All six canonical CI jobs are green, the branch is current with `main` and mergeable, and the PR body now includes a focused review guide, default-byte/API and migration safety evidence, the quarantined scope, and the exact boundary around the separate 441/500 combined-protocol reproduction.

GitHub did not permit the fork author to use the formal reviewer-request API, so I am flagging it here. No merge or release action is requested by this comment.

Tosko4 (changes requested):
I reviewed exact head `37cfcfd0442ba979d602ae86e898663621dfba28`. The six CI jobs are green, and the clean full suite passed locally with `2296 passed, 12 xfailed`. The default/snippets byte-compatibility path, additive time-column migration, base reopen, packaging, and real host dispatch all looked good.

I still found three related blockers in the new immutable, deterministic, fail-closed contract:

1. **Exact refs can silently change meaning after a source row is rewritten.** The operand schema makes `quote` optional, `resolve_exact_ref()` only checks drift when a quote is supplied, and the ref itself contains only `store_id` plus character offsets. I created refs for `Amount $12` and `Amount $8`, rewrote those rows through the supported message-GC path to `Amount $99` and `Amount $1`, then reused the old refs without quotes. `lcm_compute` returned a verified `$100`; supplying the original quotes correctly fell back. A source-only provenance change also leaves the same ref valid. Please bind the ref to an immutable content/provenance revision, or require and verify immutable evidence identity on every compute operand. Add regressions for content rewrite and provenance drift.

2. **The selector can certify incomplete or duplicate evidence as complete.** A question naming Alice, Bob, and Carol returned a computed `$30` when only Alice and Bob were supplied, even though a matching Carol row existed, because the caller set `evidence_complete=true`. Separately, two distinct overlapping refs (`lcm:1:0-18` and `lcm:1:1-18`) over the same `$12` fact produced `$24`. The completeness flag should not override a finite evidence set stated by the question, and additive operations need to reject or safely deduplicate overlapping spans from the same source row.

3. **Temporal grounding can certify the wrong chronology.** Through the real `answer_ready` path, the message `I finished the dashboard 5 days ago. The contract was signed on 2024-03-01.` observed on March 20 assigned the dashboard occurrence to the unrelated contract date, March 1. Combined with a kickoff on March 10, `lcm_compute` returned `dashboard -> kickoff`; the actual order is `kickoff -> dashboard` because the dashboard date is March 15. Also, `As of 2024-03-01, what is the total page count?` accepted observations from March 10 when the optional `question_date` argument was omitted; the same call correctly fell back when that argument was supplied. Mixed temporal expressions need local event-to-date support or an evidence-only fallback, and an explicit as-of date in the question must be enforced rather than silently ignored.

These are all on the PR's new production path and can produce a computed claim instead of the promised evidence-only fallback. I think they need one consolidated repair batch, then a targeted rereview of these exact regressions. I did not find a migration, packaging, registration, or legacy-recall blocker.

100yenadmin:
Thanks for the careful, reproducible review — all three were real bugs on the reviewed head, and each pointed at the same contract promise (a computed claim only when evidence is verified, otherwise the evidence-only fallback). I rebased on the current PR head (`e328554`, which landed after your `37cfcfd` review) and re-ran each of your exploits before touching anything, so I only fixed what still reproduced.

**1. Exact-ref immutability — no longer reproduces on the current head.**
`e328554` binds every compute operand to an immutable `evidence_identity` (a digest over the full content and source-provenance fields), makes it **required** at resolve, and `resolve_exact_ref()` now fails closed regardless of quote presence. I replayed your exact attack against the current head:
- stale ref reused with no quote and no identity → `operands[0]: evidence_identity is required for every exact ref`
- stale ref with the pre-rewrite identity after a message-GC content rewrite → `evidence_identity does not match current content and provenance`
- source-only provenance change (session reassignment) → same fail-closed mismatch

Covered by `test_compute_rejects_identity_after_content_rewrite` and `test_compute_rejects_identity_after_provenance_rewrite`.

**2. Completeness certification — the overlap half was fixed; the finite-set half still reproduced, and this change closes it.**
- Overlapping spans (`lcm:1:0-18` + `lcm:1:1-18`) already fail closed on the current head via the same-row span-overlap check (`test_compute_rejects_overlapping_spans_from_one_source`).
- The Alice/Bob/Carol case still reproduced, but only because the finite-set detector recognized the entity list solely after `for`/`from`/`of`. Rephrasing the selector dodged the coverage check entirely — "total amount **by** / **among** / **spent by** Alice, Bob, and Carol" with only Alice and Bob supplied still computed `$30` under `evidence_complete=true`.

  Fix (`reasoning.py`): the completeness flag is treated as a claim, not a certificate. The selector anchor is broadened to any selector preposition; when an anchored slice still nests a selector ("of pages **for** Alice and Bob") the tighter enumeration wins; a bare proper-noun enumeration is honored when no preposition is present; and bare number/date items are excluded since a finite entity set is named. Incomplete evidence now falls back across all of these, while a fully-covered set still computes.

  Regressions: `test_compute_rejects_incomplete_named_evidence` (original `for` phrasing) plus `test_compute_rejects_incomplete_named_evidence_across_selectors` (parametrized over `for`/`by`/`among`/`spent by`/`paid to`) and `test_compute_accepts_complete_named_evidence_with_nonstandard_selector` (a complete `by`-set must still compute `$60`).

**3. Temporal grounding — no longer reproduces on the current head.**
`e328554` makes occurrence-time extraction fail closed on mixed expressions: a message carrying both a relative expression and an unrelated absolute date resolves to `unknown` (`ambiguous_mixed_temporal_expressions`) instead of adopting the other sentence's date, so the dashboard/kickoff chronology falls back (`order requires grounded dates`). The explicit as-of date in the question text is now enforced even when the optional `question_date` argument is omitted (`source was observed after the question-date boundary`), and a non-ISO as-of date fails closed rather than being ignored. Covered by `test_compute_falls_back_for_mixed_temporal_expressions` and `test_compute_enforces_explicit_as_of_date_without_optional_argument`, both verified against the real derivation path.

Full `tests/` suite and `ruff check .` are green on the affected surface. Ready for the targeted re-review of these exact regressions whenever you have a moment — thanks again.

Tosko4 (changes requested):
The evidence-identity repair and the original `occurrence_time` path are materially improved, and the targeted/full suites are green. I still reproduced three related contract seams on exact head `1858e15`:

1. `reasoning.py:710-719`: the schema-supported top-level `date` field bypasses the mixed-temporal fail-closed path. With `"I finished the dashboard 5 days ago. The contract was signed on 2024-03-01."`, supplying `date="2024-03-01"` still produces `dashboard -> kickoff`, although dashboard was 2024-03-15 and kickoff 2024-03-10. This needs to fall back rather than certify an incidental date from the same quote.

2. `reasoning.py:911-937`: finite named-set coverage is still fail-open. A question for three amounts belonging to Alice, Bob, and Carol computes from Alice, Bob, and Dave because the cardinality matches and selector coverage is skipped. Separately, substring matching lets `Anna` cover both `Ann` and `Anna`, so operands for only Anna and Bob satisfy “Ann, Anna, and Bob.” Coverage needs exact/injective selector matching and must apply whenever the question names a finite set, including known-cardinality plans.

3. The same validator now considers only `operand.label`, while the existing computation contract also derives labels from a string `value`/canonical `key`. Fully grounded Alice/Bob ordering operands without the optional `label` therefore fall back as missing both names. Please either preserve that compatible route or narrow the public operand contract consistently.

Suggested regression matrix: top-level-date mixed temporal evidence; wrong-name same-cardinality evidence; `Ann`/`Anna` injectivity; and value/key-derived labels. I stopped at these previously reported contract boundaries rather than opening another broad review cycle.

100yenadmin:
Thanks for the precise repros — all three reproduced on 1858e15 and are fixed in 8811f65.

**1. Top-level `date` mixed-temporal bypass.** When an operand supplies no `occurrence_time`, a bare top-level `date` now goes through the same fail-closed resolution as the occurrence path: it is certified only when the quote resolves to a single unambiguous day. A quote that mixes the date with relative phrasing (or a second explicit date) is treated as ambiguous and falls back, so `date="2024-03-01"` on the "…5 days ago… signed on 2024-03-01" quote no longer certifies an incidental date and `dashboard -> kickoff` no longer computes.

**2. Finite named-set coverage.** Coverage now runs whenever the question names a finite set — including known-cardinality sum/count/order plans, not only `requires_complete_evidence` ones — and matching is exact and injective. The same-cardinality wrong-name set (Alice/Bob/Dave for Alice/Bob/Carol) now reports Carol missing, and an `Anna` operand can no longer cover both `Ann` and `Anna`.

**3. Validator label derivation — you're right, and thanks for catching it.** That was a regression I introduced in round 1: the validator had narrowed to `operand.label` only, so fully grounded operands deriving their name from a string `value`/canonical `key` fell back as missing both names. I restored the compatible route rather than narrowing the public contract — the validator now derives labels exactly as the executor does (`label` → string `value` → canonical `key`).

Regression matrix added to `tests/test_pr423_review_blockers.py`: top-level-date mixed temporal, wrong-name same-cardinality, `Ann`/`Anna` injectivity, and value/key-derived labels (plus positive guards so fixes 1 and 3 don't over-narrow). Full-suite parity holds (identical pre-existing failure set with the change stashed) and ruff is clean.

100yenadmin:
CI note: the test (3.11) leg shows 'fail' at exactly 6h00m — a runner stall hitting the job timeout, not a test failure (3.12/3.13/3.14 passed the identical commit in 2-4 min). We lack rerun rights on this repo — could a maintainer re-run that one job when convenient?

PR #428, opened 2026-07-21
fix: keep FTS from starving semantic recall

Tosko4:
## Summary

- Cap the first FTS arm at 25% of the remaining recall budget when a requested semantic corpus actually has vectors.
- Keep the full deadline for FTS-only, unconfigured, and unbackfilled fallback paths.
- Add orchestration-level regressions for both sides of that contract.

## Why

On a production-size rc1 database, `lcm_recall(include="all")` could spend the complete shared 8-second deadline in FTS before provider resolution, summary/chunk retrieval, or reranking got a chance to run.

The resulting response looked like a Voyage/reranker problem, but summary-only recall on the same runtime worked normally. The actual issue was sequential deadline allocation: FTS received the whole request budget even when other configured arms still had work to do.

This keeps FTS useful while preventing it from starving the rest of the recall pipeline.

## Validation

The regression was RED before the fix:

```text
assert captured["fts_deadline"] < 108.0
E assert 108.0 < 108.0
```

Exact candidate results:

```text
new fairness regression: 1 passed
focused recall suite:     41 passed
full suite:               2291 passed, 12 xfailed
ruff / diff-check / compile: passed
```

The regression exercises `lcm_recall()` itself and verifies that:

- FTS times out and degrades independently;
- summary retrieval still returns a hit;
- reranking still runs;
- provenance reports the partial result honestly;
- embeddings-off, missing provider/model, or an empty requested vector corpus keeps the full FTS deadline;
- supported provider aliases select the same vector corpus as provider resolution.

## Risk and impact

This only changes internal deadline scheduling in `lcm_recall()`.

No schema, provider, scoring, RRF, response shape, feature default, or configuration surface changes. The global recall deadline is unchanged.

One deliberate trade-off remains: if matching vectors exist but the configured provider is unavailable at runtime, recall still uses the hybrid schedule and FTS gets 25% rather than the full deadline. Detecting provider health before FTS would move expensive provider resolution ahead of the cheap fallback and is outside this focused fix.

## Scope

In scope:

- FTS-first starvation in `include="all"` / hybrid recall;
- requested-corpus availability before applying the hybrid FTS cap;
- accurate partial-result timeout and provenance behavior;
- preserving full-budget FTS fallback where no semantic route is configured.

Out of scope:

- redesigning all recall arms into a parallel scheduler;
- changing provider retries or reranker behavior;
- runtime deployment or config changes.

## Rollout

No deployment is included in this PR. After review and merge, the production canary is the same `include="all"` query that reproduced the rc1 symptom, with arm provenance checked to confirm that a slow FTS arm no longer suppresses semantic retrieval and reranking.

## Reviewer focus

Please check the child-deadline math, the read-only corpus preflight, provider-alias normalization, and the fallback tests that preserve the original full FTS budget.

@stephenschoettler @100yenadmin this came directly from the rc1 production-size recall path. The Voyage/reranker route itself was healthy; FTS was simply eating the clock before it could run.

### DIFF
--- a/tests/test_lcm_recall.py
+++ b/tests/test_lcm_recall.py
@@ -498,6 +498,117 @@ def test_recall_uses_recall_timeout_budget(recall_engine, monkeypatch):
     assert payload["hits"]
 
 
+def test_slow_fts_arm_cannot_starve_semantic_recall_and_rerank(recall_engine, monkeypatch):
+    """A slow first arm degrades alone instead of consuming the whole recall budget."""
+    recall_engine._config.recall_query_timeout_s = 8.0
+    recall_engine._config.embedding_provider = "voyage"
+    recall_engine._config.embedding_model = "voyage-4-large"
+    recall_engine._config.rerank_enabled = True
+
+    clock = [100.0]
+    captured: dict[str, float] = {}
+
+    class VoyageProvider(MockProvider):
+        provider_id = "voyage"
+        model_id = "voyage-4-large"
+
+        def rerank(self, _query, documents, *, top_k=None, timeout, model="rerank-2.5-lite"):
+            return [(index, 1.0) for index, _document in enumerate(documents)]
+
+    provider = VoyageProvider()
+
+    def slow_fts(_engine, _query, *, candidate_limit, deadline):
+        del candidate_limit
+        captured["fts_deadline"] = deadline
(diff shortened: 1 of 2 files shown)


Tosko4:
@codex review

Issue #170, opened 2026-05-14
Feature: Extend import_lossless_claw to migrate OpenClaw LCM summaries (DAG nodes) into Hermes summary_nodes


willificent:
First off — **thank you** for building `import_lossless_claw.py`. It is genuinely thoughtful that hermes-lcm ships a migration path from OpenClaw's lossless-claw format. The message import worked flawlessly for us: 67,428 messages across 246 conversations, all resolved via the `lcm_imported_messages` tracking table. Idempotent, dry-run by default, properly handles empty stubs — it is a really well-made operator script.

## The Gap: Summaries

The import script covers messages but does not touch OpenClaw's `summaries` table — the compressed conversation memory (the DAG nodes that give LCM its drill-down recall). We just completed a manual migration of 1,417 OpenClaw summaries into Hermes `summary_nodes`, and the schema mapping is clean enough that it could be a natural extension of the existing script.

### OpenClaw → Hermes Schema

| OpenClaw `summaries` | Hermes `summary_nodes` | Notes |
|---|---|---|
| `summary_id` (text hash) | `node_id` (auto-increment) | Must build a mapping dict during insertion |
| `conversation_id` → `conversations.session_id` | `session_id` | Same resolution as message import |
| `depth` | `depth` | Direct |
| `kind` (`leaf` / `condensed`) | — | Used to determine `source_type` |
| `content` | `summary` | Direct |
| `token_count` | `token_count` | Direct |
| `source_message_token_count` | `source_token_count` | For leaf nodes |
| `descendant_token_count` | `source_token_count` | For condensed nodes (different source column!) |
| `summary_messages.message_id[]` | `source_ids` (JSON array of `store_id`s) | **Must resolve via `lcm_imported_messages`** |
| `summary_parents.parent_summary_id[]` | `source_ids` (JSON array of `node_id`s) | **Progressive: must insert leaves first** |
| — | `source_type` | `"messages"` for leaf, `"summaries"` for condensed |
| `created_at` (ISO datetime) | `created_at` (unix timestamp) | Parse and convert |
| `earliest_at` / `latest_at` | `earliest_at` / `latest_at` | Parse and convert |

### The Key Insight: Progressive Insertion

[...]

PR #434, opened 2026-07-23
fix: query-path embedding spend-guard starves retrieval after 60 calls/min -- make it configurable with a generous default

100yenadmin:
## Summary

- Make the query-path embedding spend guard explicit and configurable instead of implicitly inheriting the strict backfill default.
- Default query-path guard to a generous 600 calls/60s (env-overridable via `LCM_EMBEDDING_QUERY_SPEND_*`).
- Leave the backfill path's exempt bulk contract (`max_calls=0`) unchanged.

## Why

`resolve_provider(for_backfill=False)` (the query embedding path) implicitly used the strict default `EmbeddingSpendGuard()` (60 calls/60s, 60s backoff) -- a control meant for backfill economy, not interactive query traffic. Once a query loop crossed 60 calls in a 60s window, every further call was rejected pre-network with `ProviderRateLimited`, silently starving retrieval well within normal query volume.

A live probe against the query path reproduced this directly: calls 1-60 returned HTTP 200 as expected, and every call from 61 onward was rejected instantly and pre-network, with no retry or backoff recovering it within the window. The failure mode was silent -- retrieval simply degraded rather than erroring loudly.

## Fix

- `resolve_provider` now passes an explicit, configurable `EmbeddingSpendGuard` on the query path instead of falling through to the strict default.
- New `embedding_query_spend_*` config knobs (env `LCM_EMBEDDING_QUERY_SPEND_*`) default to 600 calls/60s/60s -- generous enough that normal query volume never trips it, while still bounding runaway loops.
- Backfill path is untouched: it keeps its exempt bulk contract (`max_calls=0`).
- A tripped guard still surfaces the typed `ProviderRateLimited` pre-network (zero-discard), never a silently swallowed counter.

## Tests

- `tests/test_embedding_provider.py` -- new coverage: query path stays unthrottled at 120 back-to-back embeds (root-cause repro), guard is configurable via constructor args and env, backfill stays exempt, and a tripped guard surfaces the typed rate-limit reason with the exact message.
- `tests/test_lcm_core.py` -- full existing suite, unaffected by this change, run alongside as a regression check.
- Both files green: `uv run pytest tests/test_embedding_provider.py tests/test_lcm_core.py` -- 355 passed.

## Risk and impact

Query-path behavior only; backfill's spend contract is unchanged. Default is deliberately generous so no currently-passing query workload should newly hit the guard.

### DIFF
--- a/config.py
+++ b/config.py
@@ -370,6 +370,9 @@ ENV_FIELD_SPECS: tuple[_EnvFieldSpec, ...] = (
     _EnvFieldSpec("recall_query_timeout_s", "LCM_RECALL_QUERY_TIMEOUT_S", float),
     _EnvFieldSpec("embedding_backfill_timeout_s", "LCM_EMBEDDING_BACKFILL_TIMEOUT_S", float),
     _EnvFieldSpec("embedding_max_batch_items", "LCM_EMBEDDING_MAX_BATCH_ITEMS", int),
+    _EnvFieldSpec("embedding_query_spend_max_calls", "LCM_EMBEDDING_QUERY_SPEND_MAX_CALLS", int),
+    _EnvFieldSpec("embedding_query_spend_window_seconds", "LCM_EMBEDDING_QUERY_SPEND_WINDOW_SECONDS", float),
+    _EnvFieldSpec("embedding_query_spend_backoff_seconds", "LCM_EMBEDDING_QUERY_SPEND_BACKOFF_SECONDS", float),
     _EnvFieldSpec("new_session_retain_depth", "LCM_NEW_SESSION_RETAIN_DEPTH", int),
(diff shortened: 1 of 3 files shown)


100yenadmin:
Closing in favour of the consolidated wave-1 PR #436. On inspection the fix's exact fields (`embedding_query_spend_max_calls` / `_window_seconds` / `_backoff_seconds`, generous defaults, env-mapped) turned out to be **already present in the wave-1 base** (verified at e99f342 and after — patch-identical), so every wave-1 benchmark result published in `bench/` was measured WITH this fix in place. Fork PR #170 carries provenance-marker commits acknowledging the original authorship. Thanks — the fix was right, it shipped earlier than any of us tracked, and #436 is where it reaches upstream.

PR #442, opened 2026-07-26
fix: lcm_grep crashes on a null/absent query argument

Kyzcreig:
## Summary
- Coerce the `query` argument null-safely in `lcm_grep` (full-text + semantic) and `lcm_recall` so a non-string query returns the documented `{"error": "No query provided"}` payload instead of escaping the tool as an `AttributeError`.
- Adds `tests/test_search_tool_arg_coercion.py` covering all three search tools across `full_text` / `semantic` / `hybrid`.

## Why

**Symptom.** With LCM enabled, a model that calls `lcm_grep` with a null query crashes the tool call:

```
AttributeError: 'NoneType' object has no attribute 'strip'
```

Nothing in the payload tells the model what went wrong, so the turn either aborts or the model retries the same malformed call. This is reachable in normal operation — the `query` field is a free-form model-authored string, and models routinely emit `null` when they are unsure what to search for (particularly right after a compaction, when `lcm_grep` is the recovery path being advertised in the LCM system note).

**Root cause.** `_lcm_grep_full_text` was the one search entry point still calling `.strip()` on the raw argument:

```python
query = args.get("query", "").strip()      # tools.py — full-text path
```

`args.get("query", "")` only supplies the default when the key is *absent*; an explicit `None` (or a number) is returned as-is and `.strip()` raises. `lcm_expand_query` already guards this correctly with `str(args.get("query") or "")`.

The sibling paths had a quieter variant of the same bug:

```python
query = str(args.get("query", "")).strip()  # _lcm_grep_semantic, lcm_recall
```

That does not raise, but `str(None)` is `"None"` — a truthy 4-character string — so the tool silently runs a real search for the literal term `None` and returns a confidently empty (or spuriously matching) result set instead of an error.

Because `mode=semantic` / `mode=hybrid` fall back to the full-text helper, they surfaced the crash as a degraded payload rather than a clean argument error:

```json
{"error": "full-text fallback failed: 'NoneType' object has no attribute 'strip'", "mode": "hybrid"}
```

**Fix.** Use the `str(args.get("query") or "")` form that `lcm_expand_query` already uses, at all three sites. `None`, `""`, and whitespace all fall through to the existing "No query provided" branch; a numeric query is coerced to its string form and searched normally rather than crashing.

## Validation

- [x] Focused validation: `pytest tests/test_search_tool_arg_coercion.py -q` -> `13 passed`
- [x] Regression-proof (fix reverted, tests kept): `6 failed, 7 passed` — the 7 that still pass are the intended controls (missing-key and whitespace cases, which were never broken)
- [x] Neighbour suites: `pytest tests/test_lcm_recall.py tests/test_embedding_search_modes.py tests/test_tool_contracts.py tests/test_search_tool_arg_coercion.py -q` -> `84 passed`
- [x] `ruff check tools.py tests/test_search_tool_arg_coercion.py` -> `All checks passed!`
- [ ] `pytest -q` — not run in full locally; the change is confined to three argument-coercion lines and the suites that exercise those tools are green.

## Notes
- No behaviour change for well-formed queries: for any `str` input, `str(x or "")` is `x` unless `x` is `""`, which already short-circuits to the same error branch.
- `lcm_expand_query`'s `prompt` argument was already null-safe; a test is included to pin that.

### DIFF
--- a/tests/test_search_tool_arg_coercion.py
+++ b/tests/test_search_tool_arg_coercion.py
@@ -1,0 +1,86 @@
+"""Hostile-argument coercion guards for the search-tool ``query`` argument.
+
+A model is free to emit ``{"query": null}`` (or a number) for any ``lcm_*``
+search tool. Every such tool must answer with the documented
+``{"error": "No query provided"}`` payload rather than letting an
+``AttributeError``/``TypeError`` escape the tool boundary, and must not treat a
+non-string sentinel as a literal search term.
+"""
+
+from __future__ import annotations
+
+import json
+from types import SimpleNamespace
+
+import pytest
+
+import hermes_lcm.tools as lcm_tools
+from hermes_lcm.config import LCMConfig
+from hermes_lcm.dag import SummaryDAG
+from hermes_lcm.store import MessageStore
+
+
+@pytest.fixture
+def engine(tmp_path):
+    config = LCMConfig(database_path=str(tmp_path / "coercion.db"))
+    store = MessageStore(config.database_path, ingest_protection_config=config)
+    dag = SummaryDAG(config.database_path)
+    engine = SimpleNamespace(
+        _config=config,
+        _store=store,
+        _dag=dag,
+        current_session_id="session-a",
+    )
+    try:
+        yield engine
+    finally:
+        dag.close()
+        store.close()
+
+
+@pytest.mark.parametrize("mode", ["full_text", "semantic", "hybrid"])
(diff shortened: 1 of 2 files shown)


Kyzcreig:
Hi! This PR has been open for about 35 days without anyone taking a look yet. No decision needed right now — I'd just like to get it on someone's radar. If a maintainer has a moment to glance at it and let me know whether it's headed in a useful direction (or who might be the right person to look), that would be a big help. Happy to add context or rebase if that makes reviewing easier. Thanks!

PR #443, opened 2026-07-26
fix: guard expand-query LLM response access; degrade instead of raising

Kyzcreig:
## Summary
- Guard the aux-LLM response read in `_synthesize_expansion_answer` so an unusable response degrades `lcm_expand_query` the same way a synthesis timeout already does, instead of leaking a `TypeError`/`AttributeError` out of the tool call.
- Adds `tests/test_expand_query_degradation.py` covering the malformed-response shapes plus a happy-path control.

## Why

**Symptom.** `lcm_expand_query` intermittently dies mid-call with an error that has nothing to do with the query:

```
TypeError: 'NoneType' object is not subscriptable
```

It is intermittent and load-correlated — it shows up when the auxiliary lane is contended (several expansions in flight, or the aux provider rate-limiting), and never reproduces on a quiet system. The retrieval half of the tool has already finished successfully at that point, so the caller loses a complete set of node and raw matches to a failure in the optional prose-synthesis step.

**Root cause.** The synthesis path reads the completion unguarded:

```python
response = call_llm(**call_kwargs)
content = response.choices[0].message.content   # tools.py
```

`call_llm` does not guarantee a well-formed completion. When the auxiliary provider returns an error-shaped or partial envelope, `choices` can be `None`, absent, an empty list, or a non-list sentinel, and each shape raises a different opaque exception from that one line:

| response shape | raised |
|---|---|
| `choices=None` | `TypeError: 'NoneType' object is not subscriptable` |
| `choices` attribute missing | `AttributeError` |
| `choices=[]` | `IndexError` |
| `choices=<non-list sentinel>` | `TypeError` |
| choice present, no `.message` | `AttributeError` |

`lcm_expand_query` wraps the synthesis call, but its `except` clause only catches `TimeoutError`, so all of the above propagate out of `handle_tool_call` uncaught.

**Fix.** Read `choices` / `[0]` / `.message` / `.content` defensively and raise a typed `_ExpansionSynthesisError` when the response yields no usable choice. `lcm_expand_query` catches that alongside the existing timeout branch and returns the same `_degraded_payload(...)`.

This deliberately reuses the existing degradation path rather than adding a new one: the tool already has a well-defined answer for "retrieval succeeded, synthesis didn't," and an unusable response is the same situation as a timed-out one from the caller's perspective. The caller keeps `matches`, `raw_matches`, and `node_ids`; only the synthesized prose is missing, and `degraded: true` says so explicitly.

## Validation

- [x] RED-proof — fix reverted, tests kept: `6 failed, 1 passed`. The one pass is the happy-path control, which must pass in both states.
- [x] GREEN: `pytest tests/test_expand_query_degradation.py -q` -> `7 passed`
- [x] Neighbour suites: `pytest tests/ -q -k "expand or tool_contract or recall"` -> `139 passed, 2177 deselected`
- [x] `ruff check tools.py tests/test_expand_query_degradation.py` -> `All checks passed!`

## Notes
- No behaviour change on well-formed responses — `test_well_formed_response_still_answers` pins that the guard is transparent on the happy path.
- The typed private exception keeps the guard from swallowing unrelated `TypeError`s raised deeper inside the call; only the "no usable choice" condition is converted into a degraded payload.

### DIFF
--- a/tests/test_expand_query_degradation.py
+++ b/tests/test_expand_query_degradation.py
@@ -1,0 +1,160 @@
+"""``lcm_expand_query`` degradation guards for unusable aux-LLM responses.
+
+The expansion synthesis call is made through ``agent.auxiliary_client.call_llm``,
+which can hand back an error-shaped or partial object under load. Reading
+``response.choices[0].message.content`` unguarded turns that into an opaque
+``TypeError``/``AttributeError`` that kills the whole tool call; the tool should
+instead degrade to the same payload it already returns on a synthesis timeout,
+so the caller keeps its node and raw matches.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from types import ModuleType, SimpleNamespace
+
+import pytest
+
+import hermes_lcm.tools as lcm_tools
+from hermes_lcm.config import LCMConfig
+from hermes_lcm.dag import SummaryDAG, SummaryNode
+from hermes_lcm.store import MessageStore
+
+
+@pytest.fixture
+def engine(tmp_path):
+    config = LCMConfig(database_path=str(tmp_path / "expand.db"))
+    store = MessageStore(config.database_path, ingest_protection_config=config)
+    dag = SummaryDAG(config.database_path)
+    engine = SimpleNamespace(
+        _config=config,
+        _store=store,
+        _dag=dag,
+        current_session_id="session-a",
+    )
+    store.append_batch(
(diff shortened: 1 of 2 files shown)


Kyzcreig:
Hi! This PR has been open for about 35 days without anyone taking a look yet. If someone has a few minutes, I'd appreciate a first pass just to get eyes on it — no decision needed at this stage, just visibility. It's currently open, not a draft, and merges cleanly. Happy to add context, split it up, or rebase if that would make reviewing easier. Thanks!

PR #445, opened 2026-07-26
fix: repair stale FTS triggers left by schema migrations

Kyzcreig:
## Summary
- Detect external-content FTS triggers that exist **by name** but whose **body has drifted** from the spec, and drop-then-recreate them during repair so the fix actually lands.
- Adds `tests/test_fts_trigger_drift.py` (11 tests).

## Why

**Symptom.** After upgrading across a schema change, LCM can land in a state where **every message write fails** and the health check insists nothing is wrong:

```
sqlite3.OperationalError: no such column: new.body_text
```

Every `INSERT INTO messages` aborts, so the session stops recording. Running a repair reports success and changes nothing:

```python
>>> external_content_fts_needs_repair(conn, spec)
False                                                    # <- claims healthy
>>> repair_external_content_fts(conn, spec, throttle=False)
{'rebuilt': False, 'degraded': False, 'triggers_recreated': False}
>>> conn.execute("INSERT INTO messages(content) VALUES ('x')")
OperationalError: no such column: new.body_text          # <- still broken
```

The database is unusable for writes and there is no in-product way out of it.

**Root cause.** Two independent gaps that compound:

1. **Detection is name-only.** `_fts_missing_triggers` compares the *set of trigger names* in `sqlite_master` against the names in the spec. A trigger left behind by an older schema — whose body still references a column that has since been renamed — is present by name, so the check is satisfied and `external_content_fts_needs_repair` returns `False`.

2. **Repair cannot replace an existing trigger.** `repair_external_content_fts` re-runs the spec's SQL, but those statements are `CREATE TRIGGER IF NOT EXISTS`. Against a trigger that already exists under that name, they no-op. The drifted body survives the repair, and `triggers_recreated: False` is reported.

So the one code path that could fix the problem is also the path that cannot see it.

**Fix.**

- `_fts_stale_triggers(conn, spec)` compares each stored trigger body against the spec's SQL and returns `True` on a real difference. `external_content_fts_needs_repair` now ORs it in alongside the structural and missing-name checks.
- `_normalize_trigger_sql` canonicalizes both sides before comparing. This matters: SQLite stores trigger source with `IF NOT EXISTS` stripped and the trailing `;` removed, so a naive byte compare would flag *every* trigger as stale and force a needless repair on every startup. Normalization drops `IF NOT EXISTS`, collapses whitespace, strips the terminator, and casefolds — so only genuine body differences survive.
- In `repair_external_content_fts`, when stale triggers are detected, `_drop_fts_triggers` runs before the recreate loop so the canonical body is guaranteed to be installed. `triggers_recreated` now reflects missing-or-stale.

The drop is deliberately gated on `triggers_were_stale` rather than done unconditionally, so the common healthy path keeps its existing no-op behaviour.

## Validation

- [x] **Bug reproduced on unmodified `main`** — detection returns `False`, writes abort, repair no-ops. Full transcript in the test-evidence notes.
- [x] RED-proof — fix reverted, tests kept: `9 failed, 2 passed`. The 2 passes are the intended controls (healthy-trigger and missing-trigger paths, which were never broken).
- [x] GREEN: `pytest tests/test_fts_trigger_drift.py -q` -> `11 passed`
- [x] Neighbour suites: `pytest tests/test_db_bootstrap_fts.py tests/test_lcm_core.py tests/test_fts_trigger_drift.py -q` -> `332 passed`
- [x] `ruff check db_bootstrap.py tests/test_fts_trigger_drift.py` -> `All checks passed!`

## Notes
- `test_repair_is_idempotent_and_leaves_no_stale_state` pins that a second repair reports `triggers_recreated: False` — i.e. normalization is not producing perpetual false-positive drift.
- `test_missing_trigger_still_reported_and_recreated` pins the pre-existing missing-by-name path is unchanged.
- The normalization is parametrized against the three cosmetic differences SQLite actually introduces (`IF NOT EXISTS` stripping, whitespace/indentation, keyword case), plus a negative case proving a real column-name difference is still caught.

### DIFF
--- a/db_bootstrap.py
+++ b/db_bootstrap.py
@@ -2907,8 +2907,50 @@ def _fts_missing_triggers(conn: sqlite3.Connection, spec: ExternalContentFtsSpec
     return bool(expected - existing)
 
 
+def _normalize_trigger_sql(sql: str) -> str:
+    """Canonicalize a CREATE TRIGGER statement for body comparison.
+
+    SQLite stores a trigger's source with ``IF NOT EXISTS`` stripped and the
+    trailing ``;`` removed, so a byte compare against the spec's SQL always
+    mismatches. Normalize both sides (drop ``IF NOT EXISTS``, collapse
+    whitespace, strip the terminator, casefold) so that the only differences
+    which survive are real ones — e.g. a body that still references a column
+    that has since been renamed.
+    """
+    text = re.sub(r"(?i)\bIF\s+NOT\s+EXISTS\s+", "", sql or "")
+    text = re.sub(r"\s+", " ", text).strip()
+    return text.rstrip(";").strip().lower()
+
+
+def _fts_stale_triggers(conn: sqlite3.Connection, spec: ExternalContentFtsSpec) -> bool:
+    """True if a trigger exists by name but its body has drifted from the spec.
+
+    ``_fts_missing_triggers`` only checks existence by name, so a trigger left
+    behind by an older schema passes that check while aborting every write to
+    the content table at runtime.
+    """
+    for trigger_sql in spec.trigger_sqls:
+        name = _extract_trigger_name(trigger_sql)
+        if not name:
+            continue
+        row = conn.execute(
+            "SELECT sql FROM sqlite_master WHERE type='trigger' AND name = ?",
+            (name,),
+        ).fetchone()
+        if row is None or row[0] is None:
+            # Absent entirely — that is _fts_missing_triggers' job, not "stale".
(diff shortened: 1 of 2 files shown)


Kyzcreig:
Hi! This PR has been open for about 35 days without any eyes on it yet. It's not a draft and currently shows as mergeable, so there's nothing blocking a read. Would someone be able to take a quick look and let me know if it's pointed at the right people? Happy to rebase, split it up, or add context if that would make it easier to review.

Issue #441, opened 2026-07-25
Gateway hygiene double-ingest correlates with a permanent leaf-compaction no-op loop; overflow recovery unreachable at stock defaults (session death at provider hard limit)


Milgauss:
## Summary

A live Telegram session ran into the provider hard context limit (272k, Codex OAuth gpt-5.5) with zero successful LCM compactions. The compaction threshold fired correctly and repeatedly for ~16 hours of wall clock; every attempt no-op'd with `no eligible raw backlog outside fresh tail`, while the real transcript held ~186k tokens of raw, never-compacted messages outside the 32-message fresh tail. Forced overflow recovery never engaged because at stock config it cannot: it requires an assembly cap, and both knobs that create one default to 0.

This is a different deadlock shape from #414 (there, everything sat inside a count-protected tail; here the backlog is outside the tail and gets misclassified as ineligible). The store shows the same `cli,telegram` double-ingest signature as #253, but produced by the automatic gateway session-hygiene compression agent rather than manual `/compress`.

## Versions and config

- hermes-lcm v0.19.0, hermes-agent v2026.7.20, `context.engine: lcm`
- Provider `openai-codex`, model gpt-5.5 (272,000 effective context)
- Fully stock plugin config: no `LCM_*` env vars set. `compression.threshold: 0.5` in config.yaml, autoraised to 0.85 for the Codex gpt-5.5 route (threshold 231,200 tokens)
- Platform: Telegram gateway session, tool-heavy (MCP tool outputs of 20k to 60k chars per call)

## Timeline (agent.log)

First threshold crossing, 2026-07-24 20:50, ~234k tokens. From then until death every gate fired and every attempt no-op'd:

```
gateway.run: Session hygiene: 130 messages, ~264,470 tokens (actual) — auto-compressing (threshold: 85% of 272,000 = 231,200 tokens)
hermes_plugins.hermes_lcm.compaction: LCM compression no-op: no eligible raw backlog outside fresh tail
agent.conversation_compression: Compression made no progress (session=...) — skipping boundary rewrite.
gateway.run: Gateway hygiene compression for session ... did not rotate or compact in place (no session_db on the hygiene agent) — preserving the original transcript instead of overwriting it with the summary (#21301).
agent.turn_context: Preflight compression: ~265,528 tokens >= 231,200 threshold (model gpt-5.5, ctx 272,000)
[...]

Issue #414, opened 2026-07-18
codex_gpt_long_context preset should include a token-bounded fresh tail


barronlroth:
## Summary

I hit a compaction deadlock while using the exact runtime values from the `codex_gpt_long_context` preset:

```text
LCM_CONTEXT_THRESHOLD=0.75
LCM_FRESH_TAIL_COUNT=24
LCM_LEAF_CHUNK_TOKENS=8000
```

The preset does not currently set `LCM_FRESH_TAIL_MAX_TOKENS`. In a long-running, tool-heavy Telegram session, the active transcript had only 22 messages, but one synthetic preserved-objective message had grown to about 1.35 MB and contained 131 nested `[Session Arc Summary ...]` markers.

Because 22 messages was below `LCM_FRESH_TAIL_COUNT=24`, the entire active transcript was protected. LCM attempted compression three times and each attempt was a no-op:

```text
LCM compression no-op: no eligible raw backlog outside fresh tail
```

The request remained around 397K tokens against a 372K effective model context window and ended with:

```text
Context length exceeded: max compression attempts (3) reached.
```

## Workaround and validation

I added:

```text
LCM_FRESH_TAIL_MAX_TOKENS=64000
```

After restarting, LCM reduced the same session from approximately 368,921 tokens to 82,665 tokens. I then ran a tool-heavy benchmark:

1. Asked the agent to inspect gateway status using tools and summarize it.
2. Followed up by asking what I had requested and what it found.

The first turn completed across four successful model calls with multiple tool calls, and the follow-up accurately recalled the exchange. The oversized preserved-objective scaffold was no longer active.

## Why this appears to be a preset gap

PR #381 correctly makes the token cap opt-in and default-off for backward compatibility. However, the PR rationale also notes that a count-only tail makes compaction effectiveness depend on message-size luck. That is exactly what happened here.

The model-family preset is the place where a safe token bound can be enabled without changing default behavior for users who do not select the preset.

## Suggested direction

Please consider:

- adding `fresh_tail_max_tokens` to the preset-managed fields;
- benchmarking an appropriate value for `codex_gpt_long_context`;
- including the resulting value in the preset's `runtime_env`;
- adding a regression fixture where the active transcript is below the message-count limit but above the token limit because of one oversized recent or synthetic message.

The validated 64K value above is useful field evidence, but I am not assuming it is universally correct. A context-relative cap or another benchmark-backed fixed value may be better.

PR #456, opened 2026-07-27
fix: yield the count-protected fresh tail under sustained threshold pressure (#441)

Milgauss:
## Why

A count-protected fresh tail that covers the session's whole token mass makes every compaction attempt no-op while the host reports over-threshold pressure on every turn, until the session dies at the provider hard limit. This is the fatal deadlock in #441 (same class as #414), reproduced by replaying the reporter's durable transcript against both v0.19.0 and current main: at or below `fresh_tail_count` messages every attempt logs `no eligible raw backlog outside fresh tail`; just above it, the 1-2 messages outside the tail sit below `leaf_chunk_tokens` and attempts log `raw backlog outside fresh tail is below leaf chunk threshold`. Forced overflow recovery could not engage because `max_assembly_tokens` and `reserve_tokens_floor` both default to 0 (tracked separately in #441).

The fix is a fresh-tail pressure yield. In exactly that state, and only that state, the engine re-resolves the tail with a token bound derived from the observed overage so compaction can progress. `fresh_tail_max_tokens` (#381) already guards this failure mode but is opt-in; the deadlock is silent, permanent, and fatal when the operator does not know to set it, which is why the yield defaults on.

The yield engages only when all of the following hold:

- `fresh_tail_pressure_yield_enabled` is on (default) and the caller passed host-observed tokens at or above `threshold_tokens` for this attempt,
- less than one working leaf chunk of eligible raw backlog sits outside the resolved tail (the tail is the only blocker),
- the blockage is sustained: `fresh_tail_pressure_yield_min_observations` (default 3) strictly consecutive entry-point invocations ended tail-blocked under pressure. Any invocation that is not tail-blocked (an eligible preflight, a sanitation-only pass, forced-overflow recovery, a non-tail blockage) resets the streak at its scope exit; LCM-bypassed traffic and exceptions are not observations in either direction. A session reset clears everything and advances an epoch so that a reset inside a nested invocation scope stays authoritative after the scope unwinds.

The armed bound itself is invocation-scoped: cleared on every exit path, saved and restored across nested invocations, never persisted. While armed, the leaf-chunk minimum is waived (the freed backlog can legitimately be smaller than one chunk). An explicit smaller `fresh_tail_max_tokens` still wins, and `resolve_fresh_tail_boundary` still enforces newest-message retention and tool-group integrity.

## Validation

- `tests/test_fresh_tail_pressure_yield.py` (23 tests): both deadlock shapes compact under pressure (red on unfixed main); no-pressure and kill-switch cases preserve the exact pre-fix no-op; the sustained gate (single observation preserves the classic no-op, the Nth consecutive blocked turn yields, relief resets); strict-consecutive regressions (blocked -> eligible -> blocked, non-tail blockage, sanitation-only progress, forced-overflow preflight, exception neutrality); invocation scoping (success, summarizer exception, nested isolation, session reset, and the composed nested-reset shape where the reset must win over the scope-exit restore); preflight/compress agreement under persisted-placeholder and threshold full-sweep pressure; deterministic rollup/compaction interplay against the merged #440.
- Full suite: 2331 passed, 12 xfailed, plus the two macOS-only path-test failures that reproduce identically on stock main. Same result under `ulimit -n 1024`.
- `ruff check`, `compileall`, script `bash -n`, and `git diff --check` pass.

## Risk / impact

Default-on, but the yield cannot fire outside the deadlock it exists for: it requires host-observed over-threshold pressure on the current attempt, a tail that is the only blocker, and the sustained streak. Healthy sessions never accumulate a streak because any non-blocked invocation resets it. The failure direction is bounded in both senses: if the gate is too strict the session simply stays in the pre-fix no-op for more turns; if it engages, the bound frees only the observed overage plus one working leaf chunk and evaporates at invocation exit. Engagement logs a WARNING with the observed/threshold/tail numbers.

## Scope boundaries

- No persistence: neither the streak nor the bound is written to the store; a process restart starts the evidence over.
- `fresh_tail_max_tokens` semantics are untouched; the yield composes with it and the smaller bound wins.
- The `max_assembly_tokens` / `reserve_tokens_floor` zero-default question stays in #441.
- No store schema, rollup, or condensation changes; the #440 interplay test exercises published rollups but changes nothing about them.

## Rollout / operator notes

- `LCM_FRESH_TAIL_PRESSURE_YIELD_ENABLED` (default `true`): rollback switch. Setting `false` restores the strict count tail immediately; no state to migrate.
- `LCM_FRESH_TAIL_PRESSURE_YIELD_MIN_OBSERVATIONS` (default `3`): consecutive blocked invocations before the yield engages; `1` restores first-observation yielding.
- Both are documented in the README and operator guide configuration tables.
- The engagement WARNING is the signal to tune `fresh_tail_count` or set `fresh_tail_max_tokens` deliberately.

## Reviewer focus

- The verdict-at-exit logic in `_fresh_tail_pressure_yield_invocation` (engine.py): "blocked" keeps the streak, "neutral" (bypassed traffic) leaves it, anything else resets it, and exceptions skip the verdict. This is the strict-consecutive contract.
- The guard in `_mark_preflight_compression_requested`: a preflight that advertises work settles the verdict as clear unless the work exists only because the yield armed this invocation. Without that exception, the yield-engaging preflight would reset the streak and the follow-up `compress` (its own invocation) would no-op against the advertised compaction.
- The epoch-guarded restore: `_clear_fresh_tail_pressure_yield_state` advances `_pressure_yield_reset_epoch`, and a scope exit restores its saved state only when no reset intervened.

https://claude.ai/code/session_011jYv3WkS7NKMqfZtU4zSSU

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,6 +329,8 @@ Most installs only need `plugins.enabled` and `context.engine: lcm`.
 | `LCM_CONTEXT_THRESHOLD` | `0.35` | Fraction of the context window that triggers LCM compaction |
 | `LCM_FRESH_TAIL_COUNT` | `32` | Recent messages protected from compaction |
 | `LCM_FRESH_TAIL_MAX_TOKENS` | `0` | Optional token cap for the protected fresh tail (`0` disables it); always retains the newest message and complete assistant/tool-result groups |
+| `LCM_FRESH_TAIL_PRESSURE_YIELD_ENABLED` | `true` | Default-on: when compaction is deadlocked because the count-protected tail covers the whole over-threshold session (#441), the tail yields to a derived token bound so compaction can progress; `false` restores the strict count tail (rollback switch) |
+| `LCM_FRESH_TAIL_PRESSURE_YIELD_MIN_OBSERVATIONS` | `3` | Consecutive tail-blocked compaction attempts under host-observed pressure before the yield engages; any attempt not blocked by the tail resets the count; `1` yields on first observation |
 | `LCM_INCREMENTAL_MAX_DEPTH` | `3` | Max DAG condensation depth (`-1` = unlimited, `0` = leaf only); enables hierarchical summarization |
 | `LCM_LEAF_CHUNK_TOKENS` | `20000` | Raw-backlog floor before leaf compaction; with dynamic chunking enabled, the base chunk target |
 | `LCM_DYNAMIC_LEAF_CHUNK_ENABLED` | `false` | Enable chunk-sized leaf compaction passes instead of compacting the whole non-tail raw backlog per pass |
--- a/compaction.py
+++ b/compaction.py
@@ -69,6 +69,7 @@ class CompactionMixin:
     def should_compress_preflight(self, messages):
         """Pre-flight check — also ingests messages into the store."""
         self._preflight_cleanup_only_due_to_boundary_cooldown = False
+        self._reset_fresh_tail_pressure_yield()
         self._maybe_reclassify_late_auxiliary_before_compaction_write()
         if self._bypasses_lcm_context_management():
             self._remember_lcm_bypass_message_prefix(self._bypass_lcm_session_id(), messages)
@@ -156,6 +157,7 @@ class CompactionMixin:
                     and self.threshold_tokens > 0
                     and replay_rough >= self.threshold_tokens
                 ),
+                observed_tokens=replay_rough,
             )
             if eligible:
                 return self._mark_preflight_compression_requested()
@@ -185,6 +187,7 @@ class CompactionMixin:
             eligible, reason = self._leaf_compaction_candidate_status(
                 messages,
                 allow_partial_leaf=self._config.threshold_full_sweep_enabled,
+                observed_tokens=rough,
             )
             if eligible:
                 return self._mark_preflight_compression_requested()
@@ -260,6 +263,7 @@ class CompactionMixin:
         *,
         force_overflow: bool = False,
         allow_partial_leaf: bool = False,
+        observed_tokens: Optional[int] = None,
     ) -> tuple[bool, str]:
         """Return whether a normal leaf compaction pass can actually run.
 
@@ -269,7 +273,37 @@ class CompactionMixin:
         backlog outside that tail is still smaller than the configured leaf
         chunk. In that case ``compress()`` would immediately no-op, so preflight
         should not advertise a compaction attempt yet.
+
+        When ``observed_tokens`` reports sustained over-threshold pressure and
+        the protected tail itself is why no pass can run, the fresh-tail
+        pressure yield re-resolves the tail with a derived token bound and the
+        check runs once more (see ``_maybe_engage_fresh_tail_pressure_yield``).
         """
+        eligible, reason = self._leaf_compaction_candidate_status_once(
+            messages,
+            force_overflow=force_overflow,
+            allow_partial_leaf=allow_partial_leaf,
+        )
+        if eligible:
+            return eligible, reason
+        if reason in (
+            "no eligible raw backlog outside fresh tail",
+            "raw backlog outside fresh tail is below leaf chunk threshold",
+        ) and self._maybe_engage_fresh_tail_pressure_yield(messages, observed_tokens):
+            return self._leaf_compaction_candidate_status_once(
+                messages,
+                force_overflow=force_overflow,
+                allow_partial_leaf=allow_partial_leaf,
+            )
+        return eligible, reason
+
+    def _leaf_compaction_candidate_status_once(
+        self,
+        messages: List[Dict[str, Any]],
+        *,
+        force_overflow: bool = False,
+        allow_partial_leaf: bool = False,
+    ) -> tuple[bool, str]:
         if not messages:
             return False, "empty message list"
         fresh_tail_start = self._fresh_tail_start(messages)
@@ -318,7 +352,10 @@ class CompactionMixin:
             working_leaf_chunk_tokens = self._working_leaf_chunk_tokens(raw_tokens_outside_tail)
         else:
             working_leaf_chunk_tokens = self._config.leaf_chunk_tokens
-        if raw_tokens_outside_tail < working_leaf_chunk_tokens:
+        if (
+            raw_tokens_outside_tail < working_leaf_chunk_tokens
+            and self._pressure_yield_tail_token_limit <= 0
+        ):
             return False, "raw backlog outside fresh tail is below leaf chunk threshold"
         return True, "eligible raw backlog outside fresh tail"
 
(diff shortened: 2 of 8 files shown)


stephenschoettler:
Thanks for the contribution. Please send a contributor-authored revision on current `main` before we review this further:

- Make preflight agree with compress under persisted ignored-placeholder/count-tail pressure, including threshold full-sweep paths.
- Make the pressure-yield bound invocation-scoped, and clear it through success, exception, nested/reentrant, and session-reset paths.
- Make the semantics explicit: with default-on yielding, a single over-threshold observation currently makes `fresh_tail_count` a soft suffix rather than requiring sustained pressure.

Please rebase or restack on current `main`, add focused regressions for those paths, and include deterministic combined PR #440 rollup/compaction validation.

Milgauss:
Revised as requested; the branch is restacked on current `main`.

**Sustained semantics.** You were right that a single over-threshold observation made `fresh_tail_count` a soft suffix. The yield now arms only after `fresh_tail_pressure_yield_min_observations` (default 3, env `LCM_FRESH_TAIL_PRESSURE_YIELD_MIN_OBSERVATIONS`) consecutive entry-point invocations were blocked by the tail under host-observed over-threshold pressure. The evidence streak resets when a turn is observed under threshold, when a compaction pass succeeds, and on session reset. Setting the knob to 1 restores first-observation yielding.

**Invocation scoping.** `compress` and `should_compress_preflight` enter `_fresh_tail_pressure_yield_invocation`, which clears the armed bound on every exit path (success and exception), gives a nested or reentrant invocation its own clean scope while restoring the outer bound on exit, and is cleared together with the streak by session reset. Nothing outside a scope observes a bounded tail.

**Preflight/compress agreement.** The pre-ingest persisted-placeholder candidate check now passes observed tokens, so a placeholder session whose only blocker is the tail no longer parks in the ambiguous-noop state while compress would progress. The threshold full-sweep sites in `_compress_impl` engage the yield instead of reporting `raw_prefix_drained` when the "drained" prefix is really the tail covering the whole session. The yield gate also judges blockage from the same filtered backlog view the candidate status and compress use (`eligible_tokens` threaded from the call sites).

**Regressions.** `tests/test_fresh_tail_pressure_yield.py` now carries 17 tests: the sustained gate (single observation preserves the classic noop, the Nth consecutive blocked turn yields, relief resets the streak), scoping (cleared after success, after a summarizer exception, nested isolation, session reset), agreement (persisted-placeholder pressure with the yield on and off, full-sweep deadlock compacts), and the original deadlock shapes at `min_observations=1`.

**Combined #440 validation.** Added a deterministic rollup/compaction interplay test: a yield-driven compaction stales the covered day on publication and a synchronous `run_rollup_maintenance` pass rebuilds it, with no scheduler timing involved. I also merged #440 into this branch locally (clean merge) and ran the combined suite: 2322 passed, with no failures beyond two pre-existing macOS-only path-test artifacts that fail identically on stock `main`.

Milgauss:
Restacked on current main now that #440 is merged. The only conflict was the import block in engine.py (main added asyncio, this branch adds contextlib); no logic changed. Full suite on the rebased branch: 2325 passed, 12 xfailed, with the same two macOS-only path-test failures that reproduce identically on stock main. The deterministic rollup/compaction interplay test now runs against the merged #440 rather than a local merge.

Tosko4 (changes requested):
The core design is heading in the right direction and the test coverage is strong, but I can still reproduce two state bugs that violate the safety contract this PR describes.

1. **An intervening eligible invocation does not break the blocked streak.**

   With `fresh_tail_pressure_yield_min_observations=2`, this public-entrypoint sequence currently produces:

   ```text
   blocked tail-only preflight  -> False, streak 1
   ordinary eligible preflight -> True,  streak still 1
   blocked tail-only preflight  -> True,  streak 2 and yield engages
   ```

   These are two separated blockage events, not two consecutive blocked invocations. The stale evidence can also survive successful non-leaf progress paths such as sanitation-only or forced-overflow recovery. Please reset the streak whenever an invocation is not blocked specifically by fresh-tail eligibility, and after every successful context-reducing path. Add regressions for `blocked -> eligible -> blocked` and the non-leaf progress variants.

2. **A session reset inside a nested invocation can restore pre-reset tail state.**

   `_fresh_tail_pressure_yield_invocation()` saves the outer limit and restores it unconditionally in `finally`. A reset inside the nested scope clears the limit to `0`, but leaving that scope restores the saved pre-reset value. I reproduced `1234 -> reset to 0 -> inner exit restores 1234`. A session reset needs to remain authoritative after every enclosing scope exits. A reset generation/epoch guard around restoration would make that explicit; please add the composed nested-reset regression as well.

Please also document the two new default-on operator settings in the configuration docs:

- `LCM_FRESH_TAIL_PRESSURE_YIELD_ENABLED`
- `LCM_FRESH_TAIL_PRESSURE_YIELD_MIN_OBSERVATIONS`

And before re-review, please reshape the PR body to the repository headings (`Why`, `Validation`, `Risk / impact`, `Scope boundaries`, `Rollout / operator notes`, and `Reviewer focus`) so the default-on behavior and rollback switch are clear.

- focused engine/yield tests: `767 passed`
- full suite: `2328 passed, 12 xfailed`
- full suite with `ulimit -n 1024`: `2328 passed, 12 xfailed`
- diff and compile/smoke gates passed

So the broad suite is healthy; the blockers are missing state-composition cases rather than general regressions.

Milgauss:
Both reproductions confirmed; fixed in 6ede958.

**Strict-consecutive streak.** Each outermost invocation now settles a verdict at scope exit: only a final tail-blocked verdict keeps the streak, and anything else resets it, so your `blocked -> eligible -> blocked` sequence at `min_observations=2` now ends at streak 1 with the classic no-op preserved. The reset covers every non-blocked shape: an eligible preflight, a sanitation-only pass (the two "sanitized" completions now relieve pressure explicitly, matching leaf compaction), a forced-overflow preflight, and blockage attributed to anything other than fresh-tail eligibility (regression uses the boundary cooldown). Two deliberate carve-outs, both documented in the scope docstring: LCM-bypassed traffic is verdict-neutral because it observes nothing about the pressured session's tail, and an exception is not an observation in either direction (a summarizer failure mid-yield neither extends nor wipes the evidence). One subtlety worth your eye: a preflight that advertises work settles clear UNLESS the work exists only because the yield armed this invocation; without that exception the yield-engaging preflight would reset the streak and the follow-up `compress`, running as its own invocation, would no-op against the advertised compaction.

**Reset authority.** `_clear_fresh_tail_pressure_yield_state` now advances a reset epoch, and every scope exit restores its saved state only when no reset intervened; otherwise it leaves the cleared state in place. Your `1234 -> reset to 0 -> inner exit` shape is pinned as a composed regression (streak built up via real preflights, reset inside the nested scope, both exits observed at 0).

**Docs and body.** Both settings are in the README and operator-guide configuration tables with the default-on behavior and the `LCM_FRESH_TAIL_PRESSURE_YIELD_ENABLED=false` rollback called out, and the PR body is reshaped to the repository headings.

Validation on 6ede958: the yield file carries 23 tests; full suite 2331 passed, 12 xfailed, identical under `ulimit -n 1024`, with the same two macOS-only path-test failures that reproduce on stock main; ruff, compileall, script `bash -n`, and `git diff --check` clean.

PR #457, opened 2026-07-27
fix: require credential shape before sensitive-pattern redaction

Kyzcreig:
## Summary

- Require the captured value to look like a credential before `api_key`, `bearer_token` or `password_assignment` redaction fires.
- Reject environment-variable indirection, config/docs templating, and structured code references; then require a per-pattern length floor and an entropy floor.
- Leave `private_key` untouched: it is PEM-anchored, self-validating and already precise.
- Add `tests/test_sensitive_pattern_precision.py` (46 tests) covering both directions.

This is the first of two patches for the redaction precision and durability issue I raised with you privately by email earlier today. It stands alone and does not depend on the second one. Happy to close it if you would rather take a different approach; it is your architecture and your call.

## Why

The three keyword patterns anchor on a credential-ish name plus a separator, then accept whatever follows with no test on the captured value:

```python
"api_key": re.compile(
    r"(?P<prefix>...(?:api[_-]?key|...)\b\s*...[:=]\s*...)"
    r"(?P<secret>[A-Za-z0-9._~+/=-]{12,})"   # <- no shape test
    ...
```

So they match the code that proves a secret was *not* hardcoded:

```
client = OpenAI(api_key=os.environ.get("OPENAI_API_KEY"))
  -> client = OpenAI(api_key=[LCM sensitive redaction: name=api_key; chars=14; ...]("OPENAI_API_KEY"))

password = os.environ["POSTGRES_PASSWORD"]
  -> password = [LCM sensitive redaction: name=password_assignment; chars=11; ...]"POSTGRES_PASSWORD"]

client = OpenAI(api_key=settings.openai_api_key)
  -> client = OpenAI(api_key=[LCM sensitive redaction: name=api_key; chars=23; ...])

access_token = response.json()["access_token"]
  -> access_token = [LCM sensitive redaction: name=api_key; chars=13; ...]()["access_token"]
```

The destroyed span is `os.environ.get`, `settings.openai_api_key`, `response.json()`. The secret name in the string literal survives; the surrounding source is what gets eaten. Since redaction runs on the ingest write path, that lands in SQLite and the original characters are unrecoverable.

### What the gate does

`_looks_like_credential(value, pattern_name)` is applied to the captured `secret` group before substitution, on both the regex-span path (`_redact_match`) and the JSON-key-context path (`_redact_entire_sensitive_string`). It rejects, in order:

1. **Environment-variable indirection**: `os.environ`, `os.getenv`, `environ.get`, `process.env`, `import.meta.env`, `Deno.env`, `ENV[`, `System.getenv`, `Environment.GetEnvironmentVariable`, `$VAR`, `${VAR}`, `%VAR%`.
2. **Config/CI templating and doc placeholders**: `${{ secrets.X }}`, `{{ vault_x }}`, `${...}`, `$(...)`, `<anything>`, `!secret x`, `YOUR_*`/`DUMMY_*`/`TEST_*`, `REPLACE_ME`, `CHANGEME`, `TODO`, `xxx`, `...`.
3. **Structured code references** carrying no digits: dotted paths, snake_case names, subscripts, calls, camelCase. The structure requirement matters, because an unbroken single-case run like `OLDSECRET` is a literal rather than a reference and is still redacted (your `test_replay_*_lossy_redaction` tests depend on exactly that, and they stay green).

Values that clear those must then meet a length floor (`api_key` 12, `bearer_token` 16, `password_assignment` 8) and an entropy floor: mixed letters and digits, an unbroken single-case literal, a multi-word passphrase, or credential punctuation an identifier cannot carry.

`private_key` is exempt via `_VALUE_SHAPE_GATED_SENSITIVE_PATTERNS`. It produced no false positives in anything I measured, so there is nothing to fix there.

### Measured effect

110-line corpus of realistic code-review traffic: 90 non-secret lines (environment indirection, shell/YAML templating, identifier references, prose and type signatures, test fixtures) and 20 synthetic hardcoded credentials shaped like real provider keys.

| | before | after |
|---|---|---|
| spans redacted | 55 | 19 |
| false positives | **36** / 90 non-secret lines | **0** / 90 |
| FP rate (FP / redactions) | **65.5%** | **0.0%** |
| detection of real secrets | 19 / 20 (95.0%) | 19 / 20 (95.0%) |

Detection is unchanged. The one miss (`DB_PASSWORD = "..."`) fails on both versions because `\b(?:password|passwd|pwd|passphrase)\b` does not match the `DB_PASSWORD` prefix. That is a pre-existing recall gap in the shipped pattern and I have deliberately left it alone rather than widen scope here.

Per family, false positives before -> after: environment indirection 14 -> 0, templating 4 -> 0, identifier references 15 -> 0, prose 1 -> 0, test fixtures 2 -> 0, real secrets correctly redacted 19 -> 19.

## Validation

- [x] Focused validation: `pytest tests/test_sensitive_pattern_precision.py tests/test_ingest_protection.py tests/test_lcm_core.py -q` -> `481 passed`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> pass except one pre-existing wall-clock flake, see Notes
  - [x] `pytest -q` -> `2348 passed, 12 failed`, every failure pre-existing on clean `main` on this machine, see Notes
  - [x] `ruff check .` -> `All checks passed!`
- [ ] Workflow validation: no workflow changes.

## Notes

**On the failures.** I ran `pytest -q` on a pristine `upstream/main` worktree at `424a6b2` before touching anything, and got the same failure set. They are all environment-dependent on this machine (macOS, Python 3.14) and none of them touch redaction:

- Wall-clock deadline assertions in `test_embedding_provider.py` and `test_embedding_search_modes.py`, e.g. `assert (1124192.069814166 - 1124191.999403458) < 0.03`.
- `test_on_session_end_returns_quickly_under_real_sqlite_writer_lock`, a timing flake. On clean `main` it failed 4 of 5 consecutive runs and passed once.
- 2 macOS `/private` symlink path assertions in `test_path_containment.py` and `test_path_security.py`, where `TemporaryDirectory()` returns `/var/...` but the resolved path is `/private/var/...`.

Exact membership of the timing family varies run to run (I saw 12 and 13 on two runs of the same unchanged tree, with reds appearing and disappearing in both directions), so I checked it as a set rather than a list: **zero failures fall outside those files**, on either the baseline or my branch. Any test I saw fail on my branch but not in the baseline run, I re-ran 5x on a pristine `upstream/main` worktree and it failed 5/5 there too.

I am claiming "your suite minus that known-red set", not full green, and I have not touched any of them. Happy to send the pristine-baseline log if useful.

**Scope.** This does not change the write-path architecture. A false positive that still slips through is still permanent, because redaction rewrites `content` before the `INSERT`. That is the second patch, and it is deliberately separate so this one can land or not on its own merits.

**Testing style.** I followed the shape of `tests/test_ingest_protection.py` (real `LCMEngine`, real store, assert on the actual SQLite row). All credential values in the tests and the corpus are synthetic and were generated for this patch; no real key, token or transcript is included anywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,11 +447,23 @@ Supported named catalog entries are:
   assignments or JSON keys, including quoted values with spaces.
 - `private_key`: PEM private-key blocks.
 
+The three keyword patterns anchor on a credential-ish name plus a separator, so
+the matched value is additionally checked for credential shape before it is
+replaced. A value is left alone when it is environment-variable indirection
+(`os.environ.get(...)`, `process.env.X`, `$VAR`, `${VAR}`), a config or docs
+template (`${{ secrets.X }}`, `{{ vault_x }}`, `<your-key-here>`, `!secret x`,
+`YOUR_*`, `CHANGEME`), or a plain code reference (`settings.openai_api_key`,
+`user_password`, `apiKeyFromContext`). Values that clear the check need real
+entropy: mixed letters and digits, a multi-word passphrase, or credential
+punctuation an identifier cannot carry. `private_key` is PEM-anchored and
+self-validating, so it is not shape-checked.
+
 Redaction is forward-only. Enabling it does not rewrite existing SQLite rows,
 FTS shadow tables, DAG summaries, or externalized payload JSON that were written
 before the setting was enabled. Non-password placeholders include a short
 truncated SHA-256 digest for correlation. `password_assignment` placeholders omit
 the digest to avoid making password-like values easier to dictionary-check.
+
 `lcm_status`, `lcm_inspect`, and `lcm_doctor` expose the enabled state, configured pattern names,
 unknown names, source, and placeholder format without exposing raw secret values.
 
--- a/ingest_protection.py
+++ b/ingest_protection.py
@@ -167,6 +167,104 @@ _SENSITIVE_PATTERN_CATALOG: dict[str, re.Pattern[str]] = {
     ),
 }
 
+# --- Value-shape gate for the keyword patterns -----------------------------
+#
+# The three keyword patterns above anchor on a credential-ish name plus a
+# separator, then accept whatever follows. Without a shape test on the captured
+# value they match the code that proves a secret was NOT hardcoded (environment
+# lookups, config templating, plain identifier references) far more often than
+# they match a secret. Because redaction runs on the ingest write path, each
+# such match destroys non-secret text irreversibly.
+#
+# `_looks_like_credential` (below) gates every keyword-pattern match on these.
+# `private_key` is PEM-anchored, self-validating and produces no false
+# positives, so it is deliberately exempt.
+
+# Environment-variable indirection: os.environ[...] / os.getenv(...) /
+# process.env.X / $VAR / ${VAR} / %VAR% and friends. Anchored at the start of
+# the captured value.
+_ENV_INDIRECTION_RE = re.compile(
+    r"""^(?:
+          os\.environ\b
+        | os\.getenv\b
+        | environ\.get\b
+        | getenv\b
+        | process\.env\b
+        | import\.meta\.env\b
+        | Deno\.env\b
+        | ENV\[
+        | System\.getenv\b
+        | Environment\.GetEnvironmentVariable\b
+        | \$\{?[A-Za-z_][A-Za-z0-9_]*\}?
+        | %[A-Za-z_][A-Za-z0-9_]*%
+    )""",
+    re.VERBOSE,
+)
+
+# Config/CI templating and documentation placeholders that stand in for a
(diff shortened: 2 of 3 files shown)


Kyzcreig:
Hi — this PR has been open for about a month without anyone taking a look yet. No decision needed from me at this stage; I'm just hoping to get it on someone's radar. It's open, not a draft, and currently mergeable, so it should be straightforward to review whenever there's a spare moment. If someone can give it a first pass — or point me to the right person to tag — that would be a big help. Happy to answer questions or make changes if anything needs adjusting.

PR #458, opened 2026-07-27
feat: optionally retain pre-redaction content so redaction is not data loss

Kyzcreig:
## Summary

- Add `LCM_SENSITIVE_RETAIN_RAW_CONTENT` (default `false`). When enabled alongside sensitive patterns, the pre-redaction text is kept in a separate **non-indexed** `messages.content_raw` column.
- `messages.content` is unchanged: it still holds only the redacted form, so FTS, search, summarization, active replay and externalization behave exactly as today.
- Lazy `ALTER` recorded with the named `sensitive_raw_content_v1` migration marker. `SCHEMA_VERSION` stays at **5**.
- `MessageStore.get_raw_content(store_id)` for read-back; `sensitive_pattern_status` reports `retain_raw_content`.
- Add `tests/test_sensitive_raw_content_retention.py` (17 tests).

Second of two patches for the redaction issue I raised with you privately by email earlier today. **Stacked on #457** (that PR's commit is the base here), so please take that one first. This is the bigger and more opinionated of the two, and I have kept it entirely behind a default-off setting so it changes nothing for existing installs.

## Why

`protect_message_for_ingest` rewrites `msg["content"]` before the `INSERT INTO messages` at `store.py`, so the original characters never reach SQLite. `_sensitive_placeholder()` keeps name, char count, byte count and a truncated SHA-256, which is verifiable but unreconstructable. There is no raw column, no sidecar and no audit log.

The README already documents that redaction is intentionally lossy for matched values, and that is a legitimate design choice. The part worth revisiting is what happens when the match is wrong: with no second copy anywhere, a pattern false positive is permanent data loss in a transcript store, and the operator has no way to tell it happened or to get the text back.

The companion PR takes the measured false-positive rate on a realistic corpus from 65.5% to 0.0%. But no pattern set is perfect. A residual false positive still exists, for instance a filesystem path to where a credential lives rather than the credential:

```
api_key: /etc/secrets/openai_v2_key.txt
```

This patch makes that a rendering annoyance instead of a deletion.

## Design

I kept your architecture rather than moving redaction to display time, which would have meant touching every reader.

- **`content` semantics are untouched.** Every existing consumer keeps seeing only the redacted form. Nothing downstream needs to know this feature exists.
- **`content_raw` is never indexed.** It is not in the `messages_fts` spec (which indexes `content` only) and has no index of its own, so retained text is never searchable and can never surface through a query path that only knows about `content`.
- **Written only for rows a sensitive pattern actually rewrote.** Externalization and truncation markers are already recoverable through their own refs, so their payloads are not duplicated here. An unmatched message costs nothing.
- **Named marker, not a schema bump.** The column is added lazily on first use via `add_column_if_missing` and recorded as `sensitive_raw_content_v1`, following the `temporal_rollups_v1` / `embeddings_v1` idiom. `SCHEMA_VERSION` stays 5, so an install with the setting off keeps the v5 shape and stays readable by a base build.
- **Separate `UPDATE`, not a widened `INSERT`.** The `INSERT` column list and therefore the `msg_fts_insert` trigger are untouched, and `msg_fts_update` fires only on `UPDATE OF content`, which this is not.
- **Off by default.** Retaining raw text is the right trade when transcript fidelity matters more than the store never holding a matched secret, and the wrong one otherwise. That is the operator's call, not mine.

### One thing worth flagging

`LCMEngine._ingest_messages` calls `_redact_active_replay_messages` **before** `protect_messages_for_ingest`, so by the time ingest protection runs, its input is already redacted and the original characters are gone. The engine therefore re-supplies them from the untouched caller-provided `messages` list by absolute index, via a small public `attach_retained_raw_content()`.

Without that, the feature silently no-ops on the engine path while every direct-store test passes, which is exactly the sort of false green I would rather point at than leave for you to find.

## Validation

Proven end to end through the real `LCMEngine._ingest_messages` path, then reading SQLite directly:

```
===== RETENTION OFF (current behaviour) =====
content_raw column present : False
  content     : api_key: [LCM sensitive redaction: name=api_key; chars=30; ...]
  content_raw : None

===== RETENTION ON =====
content_raw column present : True
  content     : api_key: [LCM sensitive redaction: name=api_key; chars=30; ...]
  content_raw : 'api_key: /etc/secrets/openai_v2_key.txt'

  false positive destroyed in `content`      : True
  false positive RECOVERABLE via content_raw : True
  real secret still redacted in `content`    : True
  FTS matches for a term only in raw text    : 0
```

Same redacted `content` in both runs (identical sha256 in the placeholder), a real synthetic secret still redacted, and `SELECT COUNT(*) FROM messages_fts WHERE messages_fts MATCH ...` returns 0 for a term that appears only in the retained text.

- [x] Focused validation: `pytest tests/test_sensitive_raw_content_retention.py tests/test_sensitive_pattern_precision.py tests/test_ingest_protection.py tests/test_lcm_core.py tests/test_db_bootstrap_fts.py tests/test_rollup_store.py tests/test_schema_stamp_remediation.py -q` -> `590 passed`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> pass except one pre-existing wall-clock flake, see Notes
  - [x] `pytest -q` -> `2348 passed, 12 failed`, every failure pre-existing on clean `main` on this machine, see Notes
  - [x] `ruff check .` -> `All checks passed!`
- [ ] Workflow validation: no workflow changes.

## Notes

**On the failures.** I ran `pytest -q` on a pristine `upstream/main` worktree at `424a6b2` before touching anything, and got the same failure set. They are all environment-dependent on this machine (macOS, Python 3.14) and none of them touch redaction:

- Wall-clock deadline assertions in `test_embedding_provider.py` and `test_embedding_search_modes.py`, e.g. `assert (1124192.069814166 - 1124191.999403458) < 0.03`.
- `test_on_session_end_returns_quickly_under_real_sqlite_writer_lock`, a timing flake. On clean `main` it failed 4 of 5 consecutive runs and passed once.
- 2 macOS `/private` symlink path assertions in `test_path_containment.py` and `test_path_security.py`, where `TemporaryDirectory()` returns `/var/...` but the resolved path is `/private/var/...`.

Exact membership of the timing family varies run to run (I saw 12 and 13 on two runs of the same unchanged tree, with reds appearing and disappearing in both directions), so I checked it as a set rather than a list: **zero failures fall outside those files**, on either the baseline or my branch. Any test I saw fail on my branch but not in the baseline run, I re-ran 5x on a pristine `upstream/main` worktree and it failed 5/5 there too.

I am claiming "your suite minus that known-red set", not full green, and I have not touched any of them. Happy to send the pristine-baseline log if useful.

**Migration behaviour.** Enabling the setting on an existing store migrates the column lazily. Rows written before the flip stay lossy, matching the documented forward-only contract for redaction. `test_retention_can_be_enabled_on_an_existing_store` locks that.

**Alternative I did not take.** Redacting at display time instead of write time is arguably cleaner, but it means auditing every reader and it changes what `content` means for consumers outside this repo. The sidecar column gets the same durability guarantee with a much smaller blast radius. If you would rather have the display-time version, I am happy to write it instead.

All credential values in tests are synthetic and were generated for this patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,11 +447,23 @@ Supported named catalog entries are:
   assignments or JSON keys, including quoted values with spaces.
 - `private_key`: PEM private-key blocks.
 
+The three keyword patterns anchor on a credential-ish name plus a separator, so
+the matched value is additionally checked for credential shape before it is
+replaced. A value is left alone when it is environment-variable indirection
+(`os.environ.get(...)`, `process.env.X`, `$VAR`, `${VAR}`), a config or docs
+template (`${{ secrets.X }}`, `{{ vault_x }}`, `<your-key-here>`, `!secret x`,
+`YOUR_*`, `CHANGEME`), or a plain code reference (`settings.openai_api_key`,
+`user_password`, `apiKeyFromContext`). Values that clear the check need real
+entropy: mixed letters and digits, a multi-word passphrase, or credential
+punctuation an identifier cannot carry. `private_key` is PEM-anchored and
+self-validating, so it is not shape-checked.
+
 Redaction is forward-only. Enabling it does not rewrite existing SQLite rows,
 FTS shadow tables, DAG summaries, or externalized payload JSON that were written
 before the setting was enabled. Non-password placeholders include a short
 truncated SHA-256 digest for correlation. `password_assignment` placeholders omit
 the digest to avoid making password-like values easier to dictionary-check.
+
 `lcm_status`, `lcm_inspect`, and `lcm_doctor` expose the enabled state, configured pattern names,
 unknown names, source, and placeholder format without exposing raw secret values.
 
--- a/README.md
+++ b/README.md
@@ -403,6 +403,7 @@ moved back to that assistant even when doing so exceeds a configured bound.
 | `LCM_IGNORE_MESSAGE_PATTERNS` | empty | Comma-separated regex patterns; matching message content is excluded from LCM storage |
 | `LCM_SENSITIVE_PATTERNS_ENABLED` | `false` | Opt in to deterministic redaction before LCM storage, FTS indexing, summarization, active replay, and externalized ingest payloads |
 | `LCM_SENSITIVE_PATTERNS` | `api_key,bearer_token,password_assignment,private_key` | Comma-separated named sensitive pattern catalog entries to apply when redaction is enabled |
+| `LCM_SENSITIVE_RETAIN_RAW_CONTENT` | `false` | Retain the pre-redaction message text in a separate non-indexed `messages.content_raw` column so a pattern false positive is recoverable |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_ENABLED` | `false` | Store oversized ingest payloads, including tool results, media blocks, and generic raw content, in plugin-managed JSON files |
 | `LCM_LARGE_OUTPUT_EXTERNALIZATION_THRESHOLD_CHARS` | `12000` | Externalization threshold for normalized payload text |
 | `LCM_LARGE_OUTPUT_ACTIVE_REPLAY_STUBBING_ENABLED` | `false` | Replace token-heavy textual tool results with recoverable externalized refs in active replay; current-turn ingest is immediate and historical assembly respects the protected fresh tail; requires large-output externalization |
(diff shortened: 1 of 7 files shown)


Kyzcreig:
Hi — this PR has been open for about 34 days without anyone taking a look yet. No decision needed right now; I'd just appreciate a quick set of eyes on it to confirm it's on the right track and pointed at the right reviewer. It's not a draft and currently shows as mergeable. If there's a better person or channel for getting it triaged, happy to redirect. Thanks!

PR #462, opened 2026-07-30
fix(reconcile): strip volatile todo annotation from replay identity to prevent post-compaction duplication

TurgutKural:
## Problem

After context compression, the host appends a task-list annotation (`_PRESERVED_TODO_CONTEXT_PREFIX`) to the last user message. This annotation changes on every compression cycle as task statuses update (e.g. `in_progress` → `completed`), so including it in the replay identity causes `_find_reconciled_cursor_for_store_tail` to fail suffix matching against the stored tail.

When reconciliation fails, the engine falls through to `cursor=0` and re-persists **every** message in the incoming batch, creating duplicate rows with distinct `store_id`s and timestamps.

### Production evidence

In a long-running session (`20260729_093838_5d68d0`, reverse-engineer profile):
- **22 compaction cycles** over ~4 hours
- **120 messages** in the last 4h window, only **38 unique** → **82% duplication rate**
- Duplicate rows have distinct `store_id`s and timestamps that correlate exactly with compaction timestamps
- Root cause chain: `engine.py:5551` sets `_ingest_cursor = len(compressed)` → next turn `_reconcile_ingest_cursor_from_store()` (`reconcile.py:804`) compares stored tail with incoming messages → `_message_replay_identity()` (`reconcile.py:127`) includes the changed todo annotation → identity match fails → `reconcile.py:940` returns `cursor=0` → full re-ingest

### Additional symptom

The stale todo annotation from a previous session (with outdated task lists) can leak into the active context as a `tool`-role message, confusing the agent with irrelevant task state from a different session.

### `reconcile.py`
- `_message_replay_identity` now strips the `_PRESERVED_TODO_CONTEXT_PREFIX` suffix (and everything after it) before computing the identity tuple. This makes identity matching stable across compression cycles regardless of task-list changes.

### `engine.py`
- `_is_replayed_context_scaffold_message` now recognises standalone todo annotation messages (content starting with the prefix) as scaffolding, preventing them from being re-ingested as durable user content.

## Tests

8 regression tests in `tests/test_reconcile_todo_annotation.py`:
- **Identity stability**: plain content, annotated v1, and annotated v2 all produce the same identity
- **Scaffold detection**: standalone todo messages recognised as scaffolding; normal messages are not
- **End-to-end no-duplication**: ingest → compress (annotation changes) → re-ingest produces exactly N+1 messages, not 2N+1
- **Control**: identical annotation also produces no duplication

All 8 new tests pass. 25 existing reconcile/replay/duplicate tests in `test_lcm_core.py` pass with no regressions.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -700,6 +700,13 @@ class LCMEngine(CompactionMixin, ResetStateMixin, ReconcileMixin, AuxiliarySessi
         """
         return self._last_compression_status
 
+    @last_compression_status.setter
+    def last_compression_status(self, value: str) -> None:
+        # ponytail: host (conversation_compression.py) resets this via setattr
+        # before each compress() pass; without a setter the property raises
+        # AttributeError.  Backing field is _last_compression_status.
+        self._last_compression_status = value
+
     @property
     def last_compression_noop_reason(self) -> str:
         """Human-readable reason for the latest no-op compression decision."""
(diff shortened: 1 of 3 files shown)


TurgutKural:
Superseded by https://github.com/stephenschoettler/hermes-lcm/pull/507, which consolidates the replay-identity fixes into a clean branch based directly on current main with focused regression coverage.

TurgutKural:

PR #464, opened 2026-07-30
fix(reconcile): skip fork/side-channel ingest that corrupts stored tail

TurgutKural:
## Problem

A forked agent (background review, cron side-channel) shares the parent's `session_id` but carries a shorter, divergent message list. When its ingest reaches the reconcile fallback in `_reconcile_ingest_cursor_from_store`, persisting it corrupts the stored tail. The parent's next reconcile then fails suffix matching, falls through to `cursor=0`, and re-persists every message — creating duplicates with distinct `store_id`s and timestamps.

### Production evidence

A long-running CLI session accumulated **1512 messages with only 356 unique (76% duplication)** and **zero compression cycles** (no DAG nodes). Three re-ingest events correlated exactly with background review and model switch events:

| Timestamp | Messages written | Trigger |
|---|---|---|
| T+0 | 73 | Background review ingest (26-msg fork list) |
| T+45s | 474 | Parent reconcile fails → full re-ingest |
| T+6min | 444 | Model switch → system prompt rebuild → re-ingest |

The existing `_is_suspicious_stale_no_overlap_snapshot` check did not catch this because it requires the incoming batch to be an exact prefix of the stored head — fork snapshots share the system prompt but diverge immediately after.

## Fix

Before the final `cursor=0` fallback ("persisted ambiguous delta"), add a guard that skips the batch when:

1. The incoming list is **strictly shorter** than the durable session count, AND
2. Its **last 5 identities have zero overlap** with the stored tail

A legitimate restart always shares tail messages with the durable store; a fork does not. This complements the existing stale-snapshot check without overlapping with it.

### Why this is safe

- **Legitimate restarts**: incoming ≥ session_count (full replay) or tail overlaps → guard skipped
- **Short deltas**: tail overlaps with stored tail → guard skipped
- **Very short sessions**: `len < session_count` is false when counts match → guard skipped
- **Empty sessions**: early return before guard is reached
- **Performance**: O(n) set construction on ≤5 incoming + existing stored_tail, negligible vs the O(n²) suffix scan

### Accepted limitation

A fork that carries the parent's last 5+ messages verbatim will not be caught (tail overlap exists). This is acceptable: if the fork replays parent context, re-ingest is less harmful since the content is already stored.

## Tests

10 regression tests in `tests/test_reconcile_fork_guard.py`:

| Test | Scenario | Expected |
|---|---|---|
| T1 | Fork, no tail overlap | Skipped, no new rows |
| T2 | Fork, head overlap only | Skipped |
| T3 | Legitimate restart, full replay + new | Cursor advances, only new persisted |
| T4 | Short delta, partial tail match | Guard skipped, existing logic persists (pre-existing behavior documented) |
| T5 | incoming == session_count | Guard skipped |
| T6 | Very short session (3 msgs) | Guard skipped |
| T7 | Empty session | Early return, guard never reached |
| T8 | Fork with tail overlap | Guard skipped (accepted false negative) |
| T9 | Fork skipped → parent continues | Parent ingest unaffected |
| T10 | End-to-end: fork + restart | Zero duplication |

All 10 new tests pass. 24 existing reconcile/replay/duplicate tests pass with no regressions (1 pre-existing failure unrelated to this change).

### DIFF
--- a/docs/fork-ingest-guard-analysis.md
+++ b/docs/fork-ingest-guard-analysis.md
@@ -1,0 +1,111 @@
+# Fork/Side-Channel Ingest Guard — Risk Analysis & Test Plan
+
+## Problem
+
+A forked agent (background review, cron side-channel) shares the parent's
+session_id but carries a different, shorter message list. When it calls
+`should_compress_preflight()` → `_ingest_messages()`, the short list
+overwrites the stored tail. The parent's next reconcile fails suffix
+matching → cursor=0 → full re-ingest → duplication.
+
+## Proposed Fix
+
+In `_reconcile_ingest_cursor_from_store`, before the final `cursor=0`
+fallback ("persisted ambiguous delta"), add a guard:
+
+```python
+if (
+    cursor is None
+    and len(incoming_identities) < session_count
+    and not set(
+        incoming_identities[-min(5, len(incoming_identities)):]
+    ).intersection(set(stored_tail))
+):
+    # Short batch with no tail overlap: fork or side-channel.
+    # A legitimate restart always shares tail messages with the
+    # durable store. Skip the batch to protect the stored tail.
+    return len(messages)
+```
+
+## Risk Assessment
+
+### False positive: legitimate ingest skipped
+
+| Scenario | Risk | Why safe |
+|---|---|---|
+| Normal restart, full replay | None | incoming >= session_count → guard skipped |
+| Normal restart, short delta | None | delta tail overlaps stored tail → intersection non-empty |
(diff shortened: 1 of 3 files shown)


TurgutKural:
Superseded by https://github.com/stephenschoettler/hermes-lcm/pull/508, rebuilt directly from current main with only the narrow fork/tail-continuity guard and its regression tests.

TurgutKural:
Superseded by https://github.com/stephenschoettler/hermes-lcm/pull/508.

PR #465, opened 2026-07-31
fix(reconcile): treat model-switch notifications as ephemeral scaffolding

TurgutKural:
## Problem

The host injects `[Note: model was just switched from X to Y...]` into the active message list when the model changes. LCM persists it as a normal user message, but the host **removes it after processing**. The next turn's incoming list no longer contains it, so the stored tail can never match — every cursor position fails suffix matching, reconcile falls through to `cursor=0`, and the entire message list is re-persisted.

### Production evidence

Session with 3 model switches, each causing a **full re-ingest**:

| Event | Messages written | Trigger |
|---|---|---|
| 08:36:47 | 549 | Compression boundary ingest |
| 08:40:19 | 549 | **Re-ingest** (model switch notification in stored tail) |
| 11:28:13 | 641 | Background review + model switch |
| 11:28:44 | 641 | **Re-ingest** (same pattern) |

Content comparison between original and re-ingested batches: **548/549 identical**, only the last message differed:

```
Stored tail:  "[Note: model was just switched from qwen3.8 to kimi-k3...]" (242 chars)
Incoming:     "kaldığın yerden devam et..."                                (95 chars)
```

The model-switch notification is a **ghost message** — it exists in the store but will never appear in any future incoming list.

### 1. `_message_replay_identity` (reconcile.py)
Model-switch notification content → empty identity string. Prevents it from participating in suffix matching.

### 2. `_is_replayed_context_scaffold_message` (engine.py)
Recognise `[Note: model was just switched from ` prefix as scaffolding. In most cases this prevents the notification from being persisted at all.

### 3. Stored tail construction (reconcile.py)
Filter model-switch rows from `stored_tail_rows` before building the identity list. This handles the case where the notification was already persisted before the fix was deployed.

## Tests

7 regression tests in `tests/test_reconcile_model_switch.py`:

| Test | Scenario | Expected |
|---|---|---|
| Identity ignores switch | Switch message → empty identity | Same as empty content |
| Stable across switches | Different switch messages | Same identity |
| Normal note unaffected | `[Note: ...]` without model switch | Different identity |
| Switch is scaffold | `_is_replayed_context_scaffold_message` | True |
| Regular note not scaffold | `[Note: remember...]` | False |
| No duplication after switch | Ingest → switch → remove → reconcile | 6 messages, not 12 |
| Tail switch does not block | Switch at stored tail end | Reconcile succeeds |

All 7 new tests pass. 51 total reconcile/replay/duplicate tests pass with no regressions.

## Relationship to other PRs

- Builds on #462 (todo annotation identity stripping) — shares the same `_message_replay_identity` pattern
- Complementary to #464 (fork/side-channel guard) — different trigger, same symptom

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -700,6 +700,13 @@ class LCMEngine(CompactionMixin, ResetStateMixin, ReconcileMixin, AuxiliarySessi
         """
         return self._last_compression_status
 
+    @last_compression_status.setter
+    def last_compression_status(self, value: str) -> None:
+        # ponytail: host (conversation_compression.py) resets this via setattr
+        # before each compress() pass; without a setter the property raises
+        # AttributeError.  Backing field is _last_compression_status.
+        self._last_compression_status = value
+
     @property
     def last_compression_noop_reason(self) -> str:
         """Human-readable reason for the latest no-op compression decision."""
@@ -3786,6 +3793,8 @@ class LCMEngine(CompactionMixin, ResetStateMixin, ReconcileMixin, AuxiliarySessi
                 and "Earlier turns have been compacted into hierarchical summaries below." in content
             )
         if content.lstrip().startswith(_PRESERVED_OBJECTIVE_CONTEXT_PREFIX):
+            return True
+        if content.lstrip().startswith(_PRESERVED_TODO_CONTEXT_PREFIX):
             return True
         if "[Expand for details:" not in content:
             return False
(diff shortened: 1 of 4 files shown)


TurgutKural:
Superseded by https://github.com/stephenschoettler/hermes-lcm/pull/507. The model-switch replay normalization is now part of the consolidated, clean replay-scaffolding PR.

TurgutKural:

PR #467, opened 2026-07-31
fix: merge adjacent assistant rows in active-context assembly

Kyzcreig:
## Problem

`_sanitize_active_context_messages` -> `_sanitize_tool_pairs` drops `tool` rows and strips `tool_calls` off assistant turns to shed token-heavy scaffolding, but **never merges the assistant narration rows left adjacent** once the tool results between them are gone.

A long agentic turn (N tool steps, each with its own narration preamble) therefore emits — and the host persists — ~N consecutive bare assistant rows. That is a strict message-role-alternation violation: the host has to repair it on every load, and the persisted message count ends up inflated above the count the model actually replays. In our deployment this surfaced as a runtime footer (`X/Ymsgs`, from the raw persisted count) over-reading the repaired `len(history)` a session-hygiene valve enforces (1463 shown vs 1003 checked), making a working valve look broken.

## Fix

Add `_merge_adjacent_assistant_messages` (in `message_analysis.py`) and call it as the final step of `_sanitize_active_context_messages`:
- union `tool_calls` (order-preserving), concatenate content, carry first non-empty `reasoning_content`
- **only merge plain-string turns** — multimodal / structured (list/dict) content is left as its own turn so attachment/thinking blocks are never mangled
- **exempt Codex Responses interim turns** (encrypted continuation state must replay verbatim)
- the merge is turned **off** (`merge_adjacent_assistants=False`) for the intermediate token-budget selection pass in `_assemble_context`, where each turn must be weighed and kept/dropped on its own — merging there would force an all-or-nothing decision on a combined row and could drop a small tail turn glued to an oversized one

The raw store and DAG are untouched (active-context-only), so `lcm_grep` / `lcm_expand` recovery granularity is fully preserved.

## Tests

3 new tests: no-consecutive-assistants, narration-text preservation, codex-interim exemption. Full suite green except pre-existing environmental flakes (wall-clock absolute-deadline tests in `test_embedding_*`; macOS `/var`->`/private/var` in `test_path_*`) which fail identically on unpatched `main`.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -117,6 +117,7 @@ from .session_patterns import (
 from .message_analysis import (
     _is_synthetic_assistant_noise,
     _matched_tool_call_ids,
+    _merge_adjacent_assistant_messages,
     _tool_call_id,
 )
 from .fresh_tail import FreshTailBoundary, resolve_fresh_tail_boundary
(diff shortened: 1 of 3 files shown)


Kyzcreig:
Hi all — this PR has been open for about a month without anyone taking a look yet. It's not a draft and currently shows as mergeable, so it should be straightforward to review. Could someone spare a few minutes to give it a first pass? Happy to add context, split it up, or rebase if that would make it easier to review.

PR #470, opened 2026-08-02
fix: close finalized LCM engine clones

bennybuoy:
## Summary

- register a supported `on_session_finalize` plugin hook
- resolve the exact per-agent LCM clone by durable session ID
- shut down only that finalized clone, leaving other active clones and the plugin prototype untouched
- preserve interactive CLI `/new`, which reuses its agent/context engine

## Why

Hermes deep-copies the registered context-engine prototype for each `AIAgent`. In long-lived Desktop and gateway processes, retiring an agent does not currently guarantee that the clone's SQLite-backed LCM helpers are closed. Finalized sessions can therefore leave Store, DAG, and lifecycle descriptors open.

The plugin already owns an exact weak session-to-engine registry and an idempotent `shutdown()` boundary. This change connects those existing pieces through Hermes' supported hard session-boundary hook; it does not patch Hermes core or sweep unrelated engines.

## Verification

- focused plugin lifecycle, hook registration, and packaging matrix: 33 passed
- 12-cycle Linux FD soak while retaining references to every finalized clone: post-warm-up LCM FD count returned from 13 to the steady baseline of 10 after every cycle
- Ruff: passed
- Python compilation: passed
- `git diff --check`: passed

## Scope

This closes plugin-owned LCM SQLite resources. It does not claim to close Hermes-owned `state.db` connections.

## Automated review

GitHub Copilot reviewed the identical diff in [bennybuoy/hermes-lcm#24](https://github.com/bennybuoy/hermes-lcm/pull/24), because fork contributors cannot request reviewers on the upstream repository. Its initial test-helper suggestion was addressed in `085d6ee`; a second review of that commit covered both changed files and generated no new comments.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -78,6 +78,33 @@ def _ensure_engine_bound_to_session(
         )
 
 
+def _make_session_finalize_handler(prototype_engine, resolve_active_lcm_engine):
+    """Close the exact per-agent LCM clone retired by a host session boundary."""
+
+    def _on_session_finalize(**payload) -> None:
+        session_id = str(payload.get("session_id") or "")
+        platform = str(payload.get("platform") or "").strip().lower()
+        if not session_id or platform == "cli":
+            # Interactive CLI /new finalizes the old session while reusing the
+            # same agent and context engine. Process exit closes its resources.
+            return
+
+        active_engine = resolve_active_lcm_engine(session_id=session_id)
+        if (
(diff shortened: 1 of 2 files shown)


bennybuoy:
Superseded by #528, refreshed onto current main with updated lifecycle/packaging coverage and explicit compatibility validation against #501.

PR #471, opened 2026-08-02
fix(reconcile): volatile system prompt, compacted tools, tool scaffold false-positive

TurgutKural:
## Problem

Three reconcile bugs cause full-session re-ingestion (duplication) in production.

### Bug 1: System prompt identity is volatile
The host rebuilds the system prompt on every model/provider switch. Because `_message_replay_identity` included the full system prompt content, any switch broke suffix matching at position 0 → cursor=0 → full re-ingest.

### Bug 2: Compacted-out tool rows break suffix matching
Leaf compaction summarizes tool results into DAG nodes and removes them from the active context, but the durable store keeps the rows (lossless design). After a compaction the incoming list has no tool rows at all, while the stored tail still contains them. Exact suffix matching fails at the first missing tool row → cursor=0 → the whole session is re-ingested.

### Bug 3: Tool rows quoting summary excerpts are mistaken for scaffolding
A tool result can legitimately quote a summary excerpt (delegation receipts, session links, embedded "[Expand for details:" blocks). `_is_replayed_context_scaffold_message` matched the summary regex on tool content, so such a row was dropped from the replay identity list during restart reconciliation. The incoming list became one row shorter than the stored tail, suffix matching shifted, cursor fell to 0 and the whole session was re-ingested.

Observed: a tool receipt containing a quoted summary excerpt caused effective_incoming=37 for a 38-message replay; cursor=0 → all 38 rows re-persisted. Each compression re-writes exactly the compacted message count, producing an ever-growing duplication loop on every restart.

### Bug 1
`_message_replay_identity` returns a role-only identity `("system", "", "", "")` for system messages.

### Bug 2
- `_matches_store_tail_suffix` stays exact (persisted-output marker identity depends on verbatim tool replay).
- New `_matches_store_tail_suffix_tool_tolerant`: two-pointer suffix match that skips stored tool rows, engaged ONLY when the full incoming list is genuinely tool-free (the compaction outcome). A normal restart replays tool rows verbatim and keeps exact matching, so changed/retried tool markers are still detected as new rows.
- Full-replay length checks use non-tool counts when the tolerant matcher is active, exact counts otherwise.

### Bug 3
Tool rows are never scaffolding — they carry real durable content. Scaffold detection still applies to system/assistant/user rows (summary nodes, preserved objective/todo context).

## Tests

- `tests/test_compression_reconcile_reingest.py` (new): compression + rebind must not re-ingest — reproduces the production loop, fails on main.
- `tests/test_reconcile_tool_scaffold_excerpt.py` (new): tool row with summary excerpt keeps identity; restart does not re-ingest; assistant summary rows still detected as scaffold.
- `tests/test_reconcile_externalized_placeholder.py`: system prompt + tool rows coexist correctly; divergent case updated for the tolerant matcher.
- 346 tests pass; `test_lcm_core` result is identical to clean main (291 passed).

### DIFF
--- a/compaction.py
+++ b/compaction.py
@@ -318,15 +318,36 @@ class CompactionMixin:
             working_leaf_chunk_tokens = self._working_leaf_chunk_tokens(raw_tokens_outside_tail)
         else:
             working_leaf_chunk_tokens = self._config.leaf_chunk_tokens
+            ctx_cap = self._context_aware_leaf_cap()
+            if ctx_cap is not None and working_leaf_chunk_tokens > ctx_cap:
+                working_leaf_chunk_tokens = ctx_cap
         if raw_tokens_outside_tail < working_leaf_chunk_tokens:
             return False, "raw backlog outside fresh tail is below leaf chunk threshold"
         return True, "eligible raw backlog outside fresh tail"
 
+    def _context_aware_leaf_cap(self) -> int | None:
+        """Return a context-proportional cap for leaf chunk sizing.
+
+        When context_length is known, leaf chunks should never exceed
+        ~40% of the model window — otherwise the fresh tail alone can
+        consume the entire context and compression becomes impossible.
+        Returns None when context_length is unknown (no clamping).
+        """
+        ctx = getattr(self, "context_length", 0) or 0
+        if ctx <= 0:
+            return None
+        return max(1, int(ctx * 0.4))
+
     def _working_leaf_chunk_tokens(self, raw_tokens_outside_tail: int) -> int:
         base = max(1, self._config.leaf_chunk_tokens)
+        ctx_cap = self._context_aware_leaf_cap()
+        if ctx_cap is not None and base > ctx_cap:
+            base = ctx_cap
         if not self._config.dynamic_leaf_chunk_enabled:
             return base
         ceiling = max(base, self._config.dynamic_leaf_chunk_max)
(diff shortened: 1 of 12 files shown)


TurgutKural:
Note for reviewers: this PR is stacked on the reconcile fix family and its diff currently includes commits from #462 (todo-annotation identity), #464 (fork/side-channel guard), #465 (model-switch prefix), #468 (context clamping) and #469 (assistant externalization) because all six touch the same identity/suffix-matching code paths and are not independently testable (verified: cherry-picking only the three #471 commits onto main fails 63+12 tests). Merge order: #462 → #465 → #464 → #468 → #469 → #471. After the earlier PRs merge, this PR's diff shrinks automatically on rebase.

TurgutKural:
Superseded by two focused PRs:

- replay/tool-scaffold normalization: https://github.com/stephenschoettler/hermes-lcm/pull/507
- compacted-out tool suffix reconciliation: https://github.com/stephenschoettler/hermes-lcm/pull/509

The volatile-system-prompt portion was intentionally dropped because no durable system-row ingest reproducer was found. Splitting the changes keeps each bug fix independently reviewable.

TurgutKural:
Superseded by https://github.com/stephenschoettler/hermes-lcm/pull/507 and https://github.com/stephenschoettler/hermes-lcm/pull/509.

PR #486, opened 2026-08-03
fix: serialize concurrent FTS bootstrap repair

100yenadmin:
## Summary
- serialize constructor-time external-content FTS repair with one short `BEGIN IMMEDIATE` ownership transaction;
- re-inspect FTS table, shadow-table, and trigger state after ownership is acquired so losing processes accept the winner's valid schema instead of recreating it;
- preserve SQLite lock provenance rather than converting contention into corruption or destructive repair;
- retain caller-owned transaction compatibility through a savepoint path;
- ensure a trigger disappearance observed on the healthy fast path re-enters ownership before any trigger installation;
- add deterministic trigger-race and six-process `spawn` regressions with message conservation, token-level search, FTS5 integrity, shadow-table parity, clean reopen, and bounded lock-contention coverage.

## Why
Two independent processes could both observe an incomplete fresh-database FTS schema and then race through unconditional virtual-table creation. The loser failed startup with:

```text
sqlite3.OperationalError: table "messages_fts" already exists
```

The existing WAL conversion retry happens earlier and does not own the later FTS repair boundary.

`repair_external_content_fts()` now separates cheap preflight detection from the destructive repair boundary:

1. detect structural or deep repair need;
2. acquire `BEGIN IMMEDIATE` when the connection does not already own a transaction;
3. re-check complete FTS/shadow/trigger state under ownership;
4. let only the owner drop, create, rebuild, and install triggers;
5. commit atomically;
6. let concurrent losers acquire the transaction after the winner, re-inspect, and accept the complete winner state.

If SQLite reports lock/busy contention on the repair-needed path, the exception remains an availability failure. It is not classified as FTS corruption and does not authorize destructive repair. Classification uses SQLite `BUSY`/`LOCKED` result codes plus genuine lock-message fallbacks; unrelated error text containing `timeout` or `busy` is not treated as lock provenance.

## Validation
- [x] Focused validation: `python -m pytest -q -p no:cacheprovider tests/test_db_bootstrap_fts.py` -> `35 passed`
- [ ] Default validation:
  - [ ] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` — exact grouped command not run; `tests/test_lcm_core.py` passed `292`, and the release validator's broader focused gate passed.
  - [ ] `pytest -q` — run; `2311 passed, 1 skipped, 12 xfailed, 4 failed` on the known macOS `/var` versus `/private/var` externalization-path baseline.
[...]

Issue #476, opened 2026-08-03
feat: define versioned provenance-bound retrieval handles and typed stale-reference failures


100yenadmin:
## Problem

Several public LCM tools expose raw `store_id`, `node_id`, externalized sidecar references, and positional offsets that callers may replay as expansion or continuation inputs. These values are database-local or mutation-sensitive locators; they do not carry logical database identity, target revision, caller scope, or a reference contract version. `expand_hint` is heterogeneous advisory text and is not a validity-bearing reference.

A replacement/restored database may reuse the same integer for a different object. A character offset may still parse after the underlying content changed. A reference may be replayed against another profile/database or under another authorization scope. A successful scalar lookup therefore does not prove the caller reached the object that originally emitted the ID.

Wrong-content resolution is already a correctness defect. It is also a conditional confidentiality risk today when one profile/database serves mutually untrusted principals. Issue #473 owns authenticated principal/tenant authorization; this issue owns durable reference identity, revision binding, and stale-reference classification and must consume the host-derived authority model rather than define one.

Raw IDs are locators, not capabilities. Some current raw lookups are intentionally cross-session, so hardening must preserve those documented semantics rather than silently convert every lookup to current-session-only behavior.

## User or operator impact

Agents and operators cannot reliably distinguish:

- current valid reference;
- logical database replacement or clone;
- deleted/missing target;
- changed target content/revision;
- wrong expected collection/session/filter scope;
- authorization denial;
- intentional cross-session lookup.

This affects saved tool results, citation/evidence drill-down, restore/import, externalized payloads, character-offset continuation, and future multi-principal collection routing. A bare legacy integer contains no origin, so if a new database reuses it the resolver cannot truthfully diagnose the collision after the fact.

## Proposed direction

Introduce additive versioned opaque references for new responses while retaining legacy scalar inputs as explicitly best-effort compatibility locators.

A new reference should bind, as applicable:

```text
reference version
logical database UUID
target kind + target identity
target revision/content digest
expected session/conversation/collection lineage
expected principal/policy/authorized-collection scope claim
issued-at and optional expiry
integrity MAC or equivalent server-side opaque identity
```

The reference’s principal/session/collection fields are provenance and expected-scope claims only. A MAC proves issuer/integrity; it is not caller authentication or authorization. Resolution must compare the token with authenticated host-derived caller scope before ranking, pagination, hydration, or expansion. User-supplied principal arguments are never authoritative.

New opaque references may return structured failures such as:

[...]

PR #487, opened 2026-08-03
feat: add provenance-bound retrieval reference foundation

100yenadmin:
## Summary
- add named, provenance-checked migrations for a durable database authority UUID and the `retrieval_references_v1` server-side registry without incrementing numeric schema version 5;
- add a strict V1 opaque `rh1_` envelope codec plus issue, resolve, revoke, and explicit-clone-rotation primitives that store only SHA-256 token digests;
- add host-supplied authorization seams with pre-lookup authorization and stored-scope authorization before target/revision disclosure;
- fail closed on incomplete migration provenance, malformed registry schema/state, database replacement, malformed canonical JSON, non-finite timestamps, revocation, expiry, and unsupported envelopes;
- document the frozen foundation contract and add deterministic migration, transaction, authorization, disclosure, restore, clone, schema, and codec coverage.

## Why
Raw database-local IDs and positional cursors are locators, not durable references. They cannot distinguish an ordinary restart from database replacement, detect target revision drift, or prove that the current trusted host context is authorized for the stored scope.

This PR implements only the additive V1 foundation needed by #476: durable logical-database identity, a digest-only opaque-reference registry, strict serialization and typed failures, and host authorization boundaries. It intentionally does not emit references from public tools, add pagination/adapters, define principal or tenant schemas, or add MAC/signing/key management.

The registry and database UUID are governed by named migration provenance while the existing numeric schema version remains 5. Completed healthy migrations are read-only; incomplete, damaged, or exact-shape-but-unmarked authority state is rejected rather than silently adopted or regenerated.

## Validation
- [x] Focused validation: `python -m pytest tests/test_retrieval_reference_contract.py tests/test_retrieval_reference_safety.py tests/test_db_bootstrap_fts.py tests/test_schema_stamp_remediation.py -q -o addopts=` -> `93 passed, 4 xfailed`
- [ ] Default validation:
  - [x] `python -m pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py tests/test_tool_contracts.py tests/test_lcm_command.py -q -o addopts=` -> `1148 passed, 1 skipped`
  - [ ] `pytest -q` -> broad run remains red only on four failures independently reproduced on accepted `main`: two import-externalization assertions and two macOS `/var` versus `/private/var` path-alias assertions; no candidate-path regression was found
  - [ ] `scripts/validate_release.sh --full --keep-going --output /tmp/hermes-lcm-release-validation-retrieval-references-v1` -> not rerun for this additive foundation; stacked prerequisite PR #486 completed the release validator, while this exact candidate completed the focused and neighboring matrices above
  - [x] `git diff --check upstream/main...HEAD && git diff --check && git diff --cached --check`
- [ ] Workflow validation, if workflows changed: `actionlint` -> not applicable; no workflow files changed
- [x] Static validation: `python -m ruff check db_bootstrap.py retrieval_references.py tests/test_retrieval_reference_contract.py` and Python compilation -> passed
- [x] Fresh ad-hoc verifier -> operation provenance, `NULL` digest rejection, and DDL-free healthy named-migration behavior passed; temporary source and bytecode were removed
- [x] Exact-current independent adversarial review -> approved immutable file hashes after reproducing every prior blocking mechanism

## Notes
- This PR is intentionally stacked on #486 (`f0fd3608`) because it extends the same bootstrap surface. Review the retrieval-reference commit `bf44dc8` independently; after #486 merges, the base range should collapse to this commit.
- Existing scalar IDs and public tool schemas are unchanged. No reference adapters or public tool emission are included.
- Authorization context is trusted caller-supplied host context. Token provenance is never treated as caller authentication or authorization.
- A byte-for-byte restore preserves authority and valid registry entries. Explicit clone rotation requires trusted authorization and revokes existing entries atomically.
- The preserved characterization file `tests/test_retrieval_reference_safety.py` remains outside this commit because it predates the implementation candidate.

Refs #476

### DIFF
--- a/db_bootstrap.py
+++ b/db_bootstrap.py
@@ -15,8 +15,9 @@ import shutil
 import sqlite3
 import threading
 import time
+import uuid
 from contextlib import contextmanager
-from typing import Iterable, Sequence
+from typing import Callable, Iterable, Sequence
 
 logger = logging.getLogger(__name__)
 
@@ -49,6 +50,28 @@ REQUIRED_CORE_TABLES = (
     "messages_fts",
     "nodes_fts",
 )
+
+DATABASE_UUID_METADATA_KEY = "database_uuid"
+DATABASE_UUID_MIGRATION = "database_uuid_v1"
+RETRIEVAL_REFERENCES_MIGRATION = "retrieval_references_v1"
+RETRIEVAL_REFERENCES_TABLE = "retrieval_references_v1"
+RETRIEVAL_REFERENCE_KINDS = (
+    "message",
+    "summary_node",
+    "externalized_payload",
+    "source_cursor",
+    "content_cursor",
+    "session_cursor",
+)
+RETRIEVAL_REFERENCE_CONSISTENCIES = ("live", "high_watermark", "snapshot")
+
+
+class DatabaseIdentityError(RuntimeError):
+    """Raised when a migrated database no longer has a valid authority UUID."""
+
+
+class RetrievalReferenceSchemaError(RuntimeError):
+    """Raised when the named V1 registry exists with an incompatible shape."""
 
 
 class ExternalContentFtsSpec:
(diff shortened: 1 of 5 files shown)


100yenadmin:
Closing in favour of **#505**, the maintenance roll-up — this is commit 2 of that stack.

Issue #477, opened 2026-08-03
feat: define mutation-safe continuation semantics for LCM pagination


100yenadmin:
## Problem

Existing caller-visible continuation paths are positional and incompletely bound to the query/target state that produced them:

- `lcm_load_session` uses `after_store_id` / integer `next_cursor`;
- node expansion slices a live source list with `source_offset` / `source_limit`;
- raw-message and externalized-payload expansion use `content_offset`.

Between page N and page N+1, append, deletion, source reassignment, reset/import/repair rewrite, content change, filter change, or database replacement can cause duplicated/skipped rows, silent missing sources, offsets applied to revised content, or an empty page that looks like end-of-data.

The issue is not whether a cursor is human-readable. A hidden/signed offset still lacks a consistency contract. Each public continuation path must say whether it is live, high-watermark, or snapshot and define how mutation invalidates it.

Bounded non-resumable tools such as `lcm_grep`, `lcm_recall`, `lcm_recent`, `lcm_describe`, and `lcm_inspect` are out of scope unless they later expose a continuation contract.

## User or operator impact

Agents can silently miss or repeat history during export, audit, backfill, citation drill-down, and memory reconstruction. A filter or target mismatch can look like legitimate exhaustion. Character offsets can return text from a newer revision.

`lcm_expand(store_id=...)` is intentionally cross-session. Strong raw-content continuation must therefore bind target/database/revision/current authorization—not implicitly require the current session.

### Preserve legacy live paths

Keep existing `after_store_id`, `source_offset`, `source_limit`, `content_offset`, and integer/null `next_cursor` fields with their current documented live behavior.

Add opt-in `page_token` input and `next_page_token` output for strong continuation. A strong token is never silently downgraded to legacy mode. Supplying it with conflicting legacy offsets or semantic filters returns `cursor_conflict`.

Typed errors apply only to strong-token mode. Preserve the existing human-readable `error` field and add structured `error_code`.

Shared taxonomy:

- `cursor_invalid`;
- `cursor_unsupported_version`;
- `cursor_scope_mismatch`;
- `cursor_stale`;
- `cursor_target_missing`;
- `cursor_conflict`.

Map database identity, target identity, and content-revision failures consistently with stable references #476 rather than creating a second vocabulary. Authorization is checked before target/revision disclosure; token fields are expected-scope claims, never authority grants.

### Strong `lcm_load_session` contract

Use bounded high-watermark semantics, not a full materialized snapshot:

[...]

PR #489, opened 2026-08-03
feat: add durable message retrieval-reference adapter

100yenadmin:
## Summary

- add one internal `message` retrieval-reference adapter over the V1 authority foundation in #487;
- bind each opaque reference to exact positive `messages.store_id`, caller-supplied authorization scope, and a SHA-256 semantic revision;
- normalize legacy blank source/conversation fields, content, valid JSON `tool_calls`, and finite timestamps before hashing;
- treat content/session/source/conversation/role/tool/timestamp changes, GC tombstones, malformed rows, and malformed registry bindings as typed stale/not-found failures;
- exclude operational `pinned` and `token_estimate` fields from semantic staleness;
- require host target preauthorization before any transaction/message lookup, making denied existing and missing IDs indistinguishable;
- preserve authorization-before-registry/message disclosure and the foundation's kind/scope/authority/expiry/revocation taxonomy;
- issue atomically under `BEGIN IMMEDIATE` or a caller-owned transaction;
- resolve under one stable read snapshot and hold `MessageStore`'s re-entrant write lock so shared-connection writers cannot interleave; raw SQLite callers retain caller-managed concurrency;
- document the adapter boundary without changing public tools, schemas, pagination, principals, or tenants.

## Why

#487 establishes a product-neutral opaque authority but intentionally does not load target rows. Before strong mutation-safe pagination (#477) can reuse that authority, one target adapter must prove exact target/revision binding, authorization-before-disclosure, transactional rollback, and mutation behavior end to end.

This PR is that single vertical. It does **not** change `tools.py`, public schemas, or numeric schema version 5.

## Verification

Sandboxed root matrix with both `HERMES_HOME` and `LCM_DATABASE_PATH` isolated:

```bash
python -m pytest -q -p no:cacheprovider \
  tests/test_message_references.py \
  tests/test_retrieval_reference_contract.py \
  tests/test_db_bootstrap_fts.py \
  tests/test_schema_stamp_remediation.py \
  tests/test_packaging_install.py
```

Result: **147 passed** (including packaging/install coverage).

Focused adapter suite: **30 passed**, including:

- round trip and exact normalized revision;
- digest-only registry persistence;
- authorization denial before registry/message SELECT;
- equal no-I/O forbidden behavior for denied existing/missing issuance targets;
- content, scope, role, tool, timestamp, GC, delete, expiry, revoke, clone, kind, and malformed-binding behavior;
- caller rollback and callback-denial behavior;
- WAL mutation snapshot stability;
- same-connection MessageStore writer exclusion;
- raw SQLite connection facade.

```bash
python -m ruff check message_references.py retrieval_references.py db_bootstrap.py \
  tests/test_message_references.py tests/test_retrieval_reference_contract.py
python -m py_compile message_references.py retrieval_references.py db_bootstrap.py \
  tests/test_message_references.py tests/test_retrieval_reference_contract.py
git diff --check
git diff --cached --check
```

All clean.

## Stack and scope

- Depends on #487 (`bf44dc81982ff507ce86a8f46fb96077a6394b5f`), which is stacked on #486.
- No public-tool or schema fan-out.
- No pagination/cursor implementation.
- No principal/tenant policy.

Refs #476.

### DIFF
--- a/db_bootstrap.py
+++ b/db_bootstrap.py
@@ -15,8 +15,9 @@ import shutil
 import sqlite3
 import threading
 import time
+import uuid
 from contextlib import contextmanager
-from typing import Iterable, Sequence
+from typing import Callable, Iterable, Sequence
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


100yenadmin:
Closing in favour of **#505**, the maintenance roll-up — this is commit 3 of that stack.

PR #490, opened 2026-08-03
fix(reconcile): strip out-of-band user message blocks from replay identity

TurgutKural:
## Problem

When the user sends a message while the agent is mid-turn, the host appends it to the tool output currently being delivered, wrapped in an `[OUT-OF-BAND USER MESSAGE ...] ... [/OUT-OF-BAND USER MESSAGE]` block. LCM persists the tool row WITH the block. On resume/restart the host replays the same tool result WITHOUT it — the user message was already delivered separately — so the stored and incoming replay identities differ by exactly that block, suffix matching fails, cursor falls to 0 and the whole session is re-ingested (duplication).

Reproduced from production forensics: a resume re-ingest batch of 272 rows where 220 were already stored; the first identity mismatch was a tool row whose stored content (3263 chars) and replayed content (3103 chars) differed only by a 160-char out-of-band block. Offline reproduction against a truncated copy of the store confirmed `cursor=0`, and stripping the block made the identities match exactly.

Sessions with many mid-turn interruptions accumulate one such row per interruption, so every resume/restart full-re-ingests the session.

## Fix

`_message_replay_identity` removes every out-of-band block from the identity content (regex over the wrapped block, DOTALL). Both directions are covered:
- stored WITH block, replayed WITHOUT (the production case)
- stored WITHOUT block, replayed WITH

Plain text that merely mentions the phrase without the marker pair is not affected.

## Tests

- `tests/test_reconcile_oob_message_block.py` (new): identity stripping (single + multiple blocks), no false positives on marker mentions, restart reconciliation in both directions. **Verified failing (4 of 5) on the unfixed tree** before the fix was applied.
- `test_lcm_core`: 292 passed — identical to clean main.

### DIFF
--- a/reconcile.py
+++ b/reconcile.py
@@ -50,6 +50,15 @@ import logging
 logger = logging.getLogger(__name__)
 
 _PRESERVED_OBJECTIVE_CONTEXT_PREFIX = "[Current user objective preserved from compacted history]"
+# When the user sends a message mid-turn, the host appends it to the tool
+# output currently being delivered, wrapped in this block.  LCM persists the
+# row WITH the block; on resume/restart the host replays the tool result
+# WITHOUT it (the user message was already delivered separately).  Strip the
+# block from the replay identity so matching survives the delivery split.
+_OOB_MESSAGE_BLOCK_RE = re.compile(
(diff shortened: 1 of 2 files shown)


TurgutKural:
Superseded by https://github.com/stephenschoettler/hermes-lcm/pull/507. The out-of-band replay normalization is preserved there together with the related identity fixes and negative coverage.

TurgutKural:

PR #492, opened 2026-08-04
fix: preserve Unicode boundaries in tiktoken state chunks

stephenschoettler:
## Summary
- Preserve UTF-8 character boundaries when the optional `tiktoken` path splits oversized trajectory-state documents.
- Re-encode each boundary-safe chunk so the provider-facing token budget remains authoritative.
- Add a real-`tiktoken` CJK regression and correct the exact-budget versus oversized semantic-index test.

## Why
- Fixed-width token-ID slices were decoded independently. A slice can end inside a multibyte character, which introduced U+FFFD replacement characters and made the concatenated chunks differ from the source text.
- The correction assembles exact token bytes, shrinks to a strict UTF-8 boundary, and verifies the resulting text still fits the requested token budget. If the budget cannot contain one complete Unicode character, the path fails explicitly instead of emitting corrupt or over-budget text.

## Validation
- [x] Focused validation: `pytest tests/test_trajectory_state_semantic_expansion.py -q` -> `25 passed`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> included in the repository release validator; focused release gate `476 passed`
  - [x] `pytest -q` -> `2793 passed, 12 xfailed`
  - [x] `bash -lc 'ulimit -n 1024 && pytest -q'` -> `2793 passed, 12 xfailed`
- [x] Independent exact-candidate review: PASS for `b3cb1da33b33a125f8bef715d20f590bc5e2f035`, including real `tiktoken 0.13.0` CJK, emoji, combining-mark, and mixed-script probes.

## Notes
- Scope is limited to trajectory-state token chunking and its regression tests: `trajectory_store.py` and `tests/test_trajectory_state_semantic_expansion.py`.
- This establishes exact source rejoin and truthful per-chunk budgets for this optional-`tiktoken` chunking path; it does not make a broader losslessness claim.

### DIFF
--- a/tests/test_trajectory_state_semantic_expansion.py
+++ b/tests/test_trajectory_state_semantic_expansion.py
@@ -650,9 +650,32 @@ def test_fallback_chunks_obey_the_shared_token_estimator(tmp_path, monkeypatch):
     assert all(token_module._fallback_token_estimate(chunk) <= 5 for chunk in chunks)
 
 
-def test_chunked_path_pools_oversize_documents(tmp_path):
-    """Every document over the test-lowered token budget takes the chunk path
-    and each state still yields exactly one usable vector."""
+def test_tiktoken_chunks_preserve_multibyte_boundaries(monkeypatch):
+    tiktoken = pytest.importorskip("tiktoken")
+    encoder = tiktoken.get_encoding("cl100k_base")
+    monkeypatch.setattr(token_module, "_get_encoder", lambda: encoder)
+    store = object.__new__(TrajectoryStore)
+    document = "漢字かな交じり文" * 20
+    token_budget = 7
+
+    token_ids = encoder.encode(document)
+    naive_rejoin = "".join(
+        encoder.decode(token_ids[start:start + token_budget])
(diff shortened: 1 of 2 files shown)


stephenschoettler:
@Tosko4 Could you please review PR #492? It is the small CJK-safe tiktoken chunking fix; the current-head CI checks are green. Thank you.

Tosko4 (approved):
Reviewed `a071fff`.

The fix preserves the exact UTF-8 content at tiktoken boundaries and rechecks every emitted chunk against the requested token budget. I reproduced the replacement-character corruption on the base commit, then verified this head with:

- `pytest tests/test_trajectory_state_semantic_expansion.py -q -o addopts=`: 25 passed with tiktoken 0.13.0
- full suite: 2795 passed, 12 xfailed
- 1,500 deterministic mixed-Unicode chunk cases plus an end-to-end CJK state-index probe
- Ruff, `py_compile`, and `git diff --check`

No blocking findings from me.

PR #494, opened 2026-08-04
fix(reconcile): filter standalone model-switch notes from stored-tail replay

TurgutKural:
## Problem

The host injects `[Note: model was just switched from X to Y ...]` as a **standalone user message** when the model changes mid-session. LCM persists that row; on the next turn the host removes the note from its active list entirely. The stored tail then contains a row the incoming replay lacks — suffix matching shifts, cursor falls to 0 and the whole session is re-ingested (duplication).

Reproduced from production forensics: a session with a model switch accumulated five copies of the standalone note and re-ingested hundreds of rows every turn; the loop only stopped when the notes fell out of the stored-tail window. A fresh resume after the switch re-ingested the entire session (299 rows, 272 already stored; the batch restarted from the session's first message).

## Fix

Two-part:

1. `_is_replayed_context_scaffold_message` recognises a model-switch note with **no user content after the closing bracket** as scaffolding. A note PREPENDED to a real user message is NOT scaffolding — the user's content after the bracket is durable.
2. `_reconcile_ingest_cursor_from_store` filters scaffold rows from the `stored_tail_rows` set, so the stored tail and the incoming replay align.

## Tests

- `tests/test_reconcile_standalone_model_switch.py` (new): standalone note is scaffold; note + user content is not; no re-ingest when the note is stripped from the replay; no re-ingest when present on both sides. **Verified failing (2 of 4) on the unfixed tree** before the fix was applied.
- `test_lcm_core`: 308 passed.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -123,7 +123,11 @@ from .fresh_tail import FreshTailBoundary, resolve_fresh_tail_boundary
 from .message_patterns import compile_message_patterns, matches_message_pattern
 from .aux_session import AuxiliarySessionMixin
 from .placeholder_ledger import PlaceholderLedgerMixin
-from .reconcile import ReconcileMixin, _PRESERVED_OBJECTIVE_CONTEXT_PREFIX
+from .reconcile import (
+    ReconcileMixin,
+    _MODEL_SWITCH_NOTIFICATION_PREFIX,
+    _PRESERVED_OBJECTIVE_CONTEXT_PREFIX,
+)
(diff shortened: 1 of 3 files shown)


TurgutKural:
Superseded by https://github.com/stephenschoettler/hermes-lcm/pull/507. The standalone model-switch-note fix is preserved in the consolidated replay-scaffolding PR.

TurgutKural:

Issue #483, opened 2026-08-03
feat: add an opt-in fail-closed LCM Teams policy adapter (multi-principal memory)


100yenadmin:
## Problem

Hermes-LCM standard mode assumes one trusted profile owns one memory domain. Teams mode needs an explicit, opt-in, default-deny policy/routing adapter that consumes authenticated `AccessContextV1` and prevents every query, expansion, write, import, maintenance, and admin path from widening authority.

The adapter must live inside the canonical Hermes-LCM repository/distribution as an isolated optional module. It is not a separate repository and must not change standard single-user behavior when disabled.

## Proposed implementation boundary

Add an in-repository Teams module with:

- explicit setup/preflight/enable/disable/status/doctor lifecycle;
- `AccessContextV1` consumer and typed denial mapping;
- tenant/profile/collection catalog interfaces;
- authorized collection resolver;
- default write-collection enforcement;
- bounded read-only federated reader;
- explicit admin-review scope handling;
- operator-only audited seed/import path;
- policy/membership/revocation revision checks;
- audit event interface;
- provider-neutral management connector interface consumed by #497.

Standard engine/store/tool handlers continue on their existing direct path unless trusted configuration explicitly enables Teams mode.

## Fail-closed rules

When Teams is enabled:

- every memory operation requires a valid current `AccessContextV1`;
- missing/malformed/stale/revoked context never falls back to standard mode;
- model/user inputs cannot select tenant, principal, grants, collection database, canonical path, or management capability;
- authorization happens before collection open, count, ranking, pagination, hydration, expansion, write, import, maintenance, doctor/debug detail, or admin review;
- denial is non-enumerating to the caller;
- collections not authorized by the resolver are never opened;
- ordinary writes target only the caller's configured employee collection;
- shared writes are unavailable in the first pilot except the trusted idempotent operator seed/import path.

## Setup and compatibility

- Installing/upgrading Hermes-LCM does not enable Teams.
- Standard setup creates no Teams catalog, policy, or storage side effects.
- Explicit enable preflights versions and creates/adopts a versioned Teams storage root; it never silently rewrites the existing single-user database.
- Disable stops Teams routing and external management authority while preserving reversible/auditable state.
- Rollback never silently reconnects old shared state after new Teams writes.
- Unknown or conflicting legacy ownership is quarantined or denied until explicitly mapped.

## Bypass matrix

The implementation must inventory and gate:

[...]

Issue #481, opened 2026-08-03
Roadmap: opt-in LCM Teams mode (multi-principal memory) and MVP handoff


100yenadmin:
## Status

**Living architecture anchor and canonical handoff tracker. Implementation is not complete.**

Target milestone: **Hermes Teams MVP**. The contributor account cannot create repository milestones, so this issue is the milestone-equivalent tracker until a maintainer creates that milestone and attaches the linked Hermes-LCM issues.

## Product decision

Hermes Teams is an **opt-in mode/module inside the canonical Hermes-LCM repository and distribution**.

It is not a separate repository and does not require a long-lived fork.

Four boundaries are independent:

1. **Repository:** source/review/release truth remains in `stephenschoettler/hermes-lcm`.
2. **Package/module:** Teams code is isolated inside the repository and activated explicitly.
3. **Runtime/data plane:** standard single-user LCM remains unchanged; enabled Teams mode is default-deny and never falls back to standard trusted-owner behavior.
4. **Management plane:** authenticated provider-neutral APIs support external control planes. ElectricSheep Dashboard + Supabase is the first connector client, not a hard-coded authority.

## Goal

Allow one managed Hermes deployment to serve multiple authenticated people and agents with:

- private employee memory;
- authorized team/organization memory;
- host-derived immutable authority;
- attenuated subagent/cron/service delegation;
- explicit tenant-admin review scopes;
- audit, revocation, migration, disable, and rollback;
- no cross-principal disclosure through guessed/replayed IDs, broad retrieval, maintenance paths, callbacks, or model-authored arguments.

## MVP shape

- one isolated tenant deployment;
- one Hermes profile per employee plus operator/admin profile;
- database/storage domain isolation for employee-private and shared collections;
- ordinary writes only to the caller's employee collection;
- operator-seeded shared collections and bounded read-only federation;
- explicit admin review, time-bounded and read-only;
- provider-neutral management connector with ElectricSheep/Supabase reference implementation;
- two-principal end-to-end pilot;
- standard single-user behavior unchanged when Teams is disabled.

The first pilot proves security, operations, and read-only federation. Employee-facing publication/collaboration UI, cross-company federation, generalized billing, and proven 100-employee capacity are post-MVP.

## Trust boundary

Authority comes from authenticated host state, never from model/tool arguments.

```text
authenticated human/service grants
∩ profile policy
∩ session ownership
∩ delegated audience/allowlist/expiry
[...]

Issue #473, opened 2026-08-03
feat: add a tenant-neutral fail-closed authorization seam for cross-session retrieval


100yenadmin:
## Summary

Hermes-LCM's cross-session retrieval surfaces currently treat the profile-local database as the trust boundary. That is correct for one trusted owner per profile, but an opt-in Teams mode needs a **tenant-neutral, fail-closed authorization seam** consumed by policy supplied outside ordinary model/tool arguments.

This issue owns the smallest reusable core enforcement contract. Product-specific roles, catalogs, admin workflows, and management connectors belong to the in-repository Teams module tracked by #481/#483.

## Current risk

In a shared/multi-principal deployment, broad retrieval or guessed/replayed IDs can cross a user boundary unless current authenticated authority is applied before query, ranking, pagination, hydration, expansion, and write/maintenance operations.

Session, conversation, profile, store, node, collection, reference, and cursor IDs are routing/lineage locators—not authorization tokens.

## Proposed neutral seam

Add a small policy protocol consumed by LCM operations:

```text
authorize_operation(context, operation, expected_scope)
resolve_authorized_targets(context, operation, requested_narrowing)
authorize_stored_scope(context, operation, stored_scope)
audit_decision(context, operation, internal_reason, public_result)
```

Exact interface shape is an implementation decision, but the ordering and failure behavior are binding:

1. validate current host context;
2. authorize operation and expected scope;
3. resolve only authorized collection/database targets;
4. inspect stored scope before content/revision disclosure;
5. perform query/ranking/hydration within authorized targets;
6. audit allow/deny without plaintext results.

The caller may narrow scope but cannot widen it. A permissive policy can preserve trusted-owner semantics; Teams mode supplies a default-deny policy.

## Fail-closed contract

- In Teams mode, missing/malformed/expired/revoked/scope-mismatched context returns typed denial.
- No fallback to ambient profile-wide behavior.
- Authorization precedes existence, count, ranking, revision, target-shape, and content disclosure.
- Missing/conflicting legacy ownership is denied or quarantined until mapped.
- Derived summaries/chunks/vectors/rollups cannot have broader scope than their sources.
- References/cursors carry expected scope but never grant authority; every use is reauthorized.
- Callback exceptions fail closed and do not leave a partially selected target or open transaction.

## Required bypass inventory

Cover every path capable of reading or writing memory:

- recall, grep, recent, load-session, expand, expand-query, exact hydration;
[...]

PR #499, opened 2026-08-04
feat: freeze the AccessContextV1 identity, delegation, lease and revocation contract

### LINKED ISSUE #482: Design: freeze AccessContextV1 identity, delegation, lease, and revocation contract
## Problem

Hermes Teams needs one immutable, host-derived authority contract shared by Hermes Agent ingress/session lifecycles and Hermes-LCM policy enforcement. Profile, session, conversation, collection, cursor, and reference IDs are routing/lineage identifiers—not credentials.

Without a frozen contract, host and storage implementations can each appear scoped while disagreeing about tenant ownership, delegation, revocation, lifetime, or denial behavior.

## Owned contract: `AccessContextV1`

Authority-bearing fields are unavailable to model/tool arguments and derived from authenticated host state.

```text
contract_revision
context_id
request_id
source_kind
authenticated_transport

deployment_id
tenant_id
principal_id
actor_type: human | agent | service
profile_id
profile_incarnation

session_id
session_owner_principal_id
conversation_id
conversation_lane

default_write_collection_id
read_policy_ref
grants
policy_revision
membership_revision
revocation_epoch

delegation_chain
delegated_by
audience
issued_at
expires_at
narrowing

ownership_generation
lease_id
lease_generation
```

The exact serialization may evolve; the semantic fields, derivation rules, validation order, and failure behavior are the contract.

### Derivation and immutability

- Host transport/session authentication derives deployment, tenant, principal, actor, profile, and ownership fields.
- Model/user arguments can only narrow an already authenticated context.
- Session/conversation/profile IDs never grant authority.
- Context is immutable within one operation; retries use a new request/lease while preserving traceable ancestry.

### Delegation

Child/subagent/service/cron authority is:

```text
parent authority
∩ explicit operation allowlist
∩ explicit collection allowlist
∩ audience/profile/session binding
∩ expiry
∩ current policy/membership/revocation state
```

Delegation never widens authority. Re-delegation preserves the full chain and all narrowing.

### Lifetime and revocation

- Resolve-and-use is protected by ownership/lease generations.
- Resume, branch, clone, reset, rebind, cron, and callbacks revalidate current ownership.
- Membership removal, profile suspension, review-scope close, or service revocation blocks the next operation.
- Stale leases, handles, references, and cursors return typed stale/revoked results without existence disclosure.

### Denial taxonomy

At minimum:

```text
context_missing
context_invalid
context_unsupported_version
context_expired
context_revoked
scope_forbidden
scope_mismatch
ownership_changed
lease_stale
target_not_found_or_forbidden
```

Public denial may intentionally blur existence. Internal audit records the exact reason without plaintext content.

### Authorization order

Validate current host context and expected scope before:

- selecting a collection/database/path;
- counting or ranking candidates;
- hydrating targets;
- revealing revision, existence, ownership, or target shape;
- issuing/revoking/using handles or cursors;
- write/import/maintenance/admin operations.

## Deliverables

1. Versioned data model and validation rules.
2. Canonical denial/result taxonomy.
3. Host-carrier and LCM-consumer protocol interfaces.
4. Positive/negative fixtures shared across repositories.
5. Delegation and revocation conformance vectors.
6. Source/bypass map covering ingress, sessions, cron, callbacks, retrieval, expansion, writes, maintenance, and admin paths.
7. Compatibility statement for standard single-user mode.

## Dependencies and consumers

- Umbrella/milestone tracker: #481
- Hermes Agent host carrier: NousResearch/hermes-agent#78802
- Routed-profile execution prerequisite: NousResearch/hermes-agent#68559 / PR #78030
- Neutral LCM authorization seam: #473
- Teams policy adapter: #483
- Durable retrieval authority: #476 / PR #487 and PR #489
- Strong continuation: #477

## Acceptance criteria

- The same fixture corpus validates the Hermes Agent producer and Hermes-LCM consumer.
- Missing context denies every Teams memory path with no unscoped fallback.
- Two principals cannot replay each other's context, lease, reference, cursor, review handle, or request ID.
- Child authority is provably a subset of parent authority.
- Revocation and ownership-generation changes invalidate the next use.
- Authorization precedes existence/count/ranking/hydration disclosure.
- Concurrent profile A/B context does not cross threads/tasks/callbacks.
- Standard mode remains unchanged when Teams is disabled.

## Non-goals

- Choosing ElectricSheep/Supabase identity schemas.
- Defining billing roles.
- Treating token integrity as caller authorization.
- Implementing the full policy adapter or collection catalog in this contract issue.

## Implementation status

Contract issue handoff ready; no production `AccessContextV1` implementation is claimed.

100yenadmin:
Closes #482. Part of the opt-in LCM Teams chain tracked by #481.

## What this is

A frozen, machine-checkable contract for the authenticated host context that opt-in LCM Teams mode will
consume — identity, delegation, lease and revocation — plus a shared fixture corpus and the
authority-path inventory.

**This PR is deliberately inert.** It adds an `access_context/` subpackage, a JSON fixture corpus, tests
and one doc. It **edits no existing file, and nothing in the repository imports it.** A reviewer should be
able to conclude "this cannot change behaviour" by inspection, then confirm it from the diff: every path
in the first commit is a new path.

The intent is that the contract lands and stops moving *before* anything enforces it, so the authorization
seam (#473) and the policy adapter (#483) implement against a fixed target instead of co-evolving with it.

## Why the fixtures are JSON rather than Python

The contract has two sides. `NousResearch/hermes-agent#78802` is the *producer* of the host context; LCM
is the *consumer*. Shipping the corpus as JSON lets the producer validate against the same vectors without
importing this package. It also keeps the PR dependency-free — **stdlib only** (PyYAML is not a dependency
here, so the inventory is JSON too).

## Fail-closed degradation is a 2×2, and the cells are kept distinct

| | Teams off | Teams on |
|---|---|---|
| **carrier absent** | standard mode; contract not consulted | **fail closed** — `context_missing`, no unscoped fallback |
| **carrier present** | **ignored** — carrying context must not by itself enable Teams | validate |

The third cell is worth calling out: a host that happens to propagate context must never self-enable
Teams. Enabling stays explicit and configuration-driven. This is also what keeps the chain safe if #78802
never lands — Teams simply stays off.

## The authority-path inventory, and why it is machine-checked

#473 asks that the inventory be machine-readable or test-enumerated **so new memory paths cannot silently
omit the hook**. That property is only real if a test fails when someone forgets. So:

- The inventory's `lcm_*` entries must equal `plugin.yaml`'s `provides_tools` **by set equality in both
  directions** — adding a tool without classifying it fails the suite.
- Every module referenced must exist on disk, and every non-tool `entry_point` must be **found in its
  module's source**, which is what stops the inventory drifting into fiction.
- The fixture-kind set is asserted **closed**, so a count assertion cannot quietly go stale.

The inventory covers **38 entries across 19 modules**: the 15 public `lcm_*` tool handlers plus **23
non-tool authority paths** — `store.py`, `compaction.py`, `rollup_store.py`, `rollup_builder.py`,
`externalize.py`, `placeholder_ledger.py`, `maintenance.py`, `reset_state.py`, `db_bootstrap.py`,
`diagnostics.py`, `scripts/import_lossless_claw.py`, `retrieval_core.py`, `vector_store.py`, `dag.py`,
`aux_session.py`, `lifecycle_state.py`, `engine_registry.py`, `engine.py`. An inventory listing only the
front door would not serve the purpose, since the point is precisely the paths that can bypass public tool
handlers.

On `cron`: this repo has **no separate OS-cron entry point**. The only cron-like path is `engine.py`'s
rollup maintenance scheduler, so that is what the inventory records, with a `notes` field saying exactly
that. Naming the absence seemed more useful than inventing an entry or leaving the category silently empty.

## Denials disclose nothing — including through the detail side-channel

Denials carry an exact **internal** reason and a blurred **public** result. `SCOPE_FORBIDDEN`,
`SCOPE_MISMATCH`, `OWNERSHIP_CHANGED` and `LEASE_STALE` all project to a single public
`TARGET_NOT_FOUND_OR_FORBIDDEN`, so a caller cannot tell "not found" from "forbidden".

Blurring the reason is not sufficient on its own, and this is worth stating because it is easy to get
wrong: if the accompanying `detail` map varies by reason, the **shape** of its keys re-identifies what the
reason field just hid. So the public projection emits a **fixed key set** (`context_id`, `request_id` —
both values the caller itself supplied), always present, carrying `None` when absent internally. Varying
*presence* leaks just as effectively as varying values.

The test asserts the public key set is identical for every `DenialReason`, that no discriminating internal
key survives, and that all reasons sharing a public bucket project to **literally equal** objects —
indistinguishable rather than merely similar.

## A note on the ordering test

`test_protocol_order_authorize_precedes_every_disclosure_primitive` freezes #473's six steps: authorize
operation → resolve authorized targets → open collection → re-authorize stored scope → count/rank →
hydrate → issue handle → audit. In particular **target resolution precedes counting and ranking**, because
counting or ranking over unresolved candidates discloses existence and count for targets the principal may
not be entitled to.

To be straightforward about what this proves: it is a **contract freeze against a reference fake, not a
behavioural proof** — nothing implements the protocol yet, by design of this inert PR. Enforcement, and
the tests binding real retrieval arms to this order, are #473.

## Naming note for reviewers

There is a `bypass.py` at the repository root; it is **LCM-bypass compaction** (sessions opting out of
context management) and is unrelated to authorization bypass. It is untouched here. To avoid the
collision, the artifact in this PR is called the **authority-path inventory** rather than a "bypass map".

## Verification

- **CI is green on this branch across 3.11 / 3.12 / 3.13 / 3.14**, plus lint and workflow-lint.
- Contract tests: `pytest tests/test_access_context_v1_{model,validation,fixtures,inventory}.py -q` →
  **77 passed**.
- Locally this checkout shows 35 failures in `test_packaging_install.py`, `test_host_capability.py` and
  `test_auto_focus_topic.py`. Those are **environment artifacts of my checkout**, not repository defects —
  they are the host-integration surface that needs a hermes-agent stub CI provides, they reproduce on
  pristine `main` @ `6b7dbb1` before any change here, and the before/after failure sets are compared by
  diff rather than by eye: **identical, no new failures**. CI is the authoritative signal, and it is clean.
- Additivity: every path in the first commit is an addition; grep for `access_context` outside its own
  tree returns only its own tests; no `pyproject.toml` change.
- The 23 non-tool entry points were each checked to exist in their module's source, so the inventory
  describes the repository rather than an aspiration.

## Review

This branch was reviewed across two different models before being opened, on the principle that a model
reviewing its own output finds less than a different one does. Four defects were found and fixed that way,
including two in the denial projection. The full review trail is in a comment below.

## What is deliberately not here

Enforcement (#473) and policy (#483). This PR freezes the shape only. It creates no Teams config, no
storage, and no behaviour change in standard single-user mode.


100yenadmin:
## Cross-model review trail

Posting this on the PR rather than keeping it internal, because two of the four defects are the kind a
reader of this diff would reasonably want to know were considered.

The working rule was that **the reviewer must not be the author** — a model reviewing its own output finds
less than a different one does. The contract corpus was written by one model, reviewed by a second, and my
own fixes to it were then reviewed by a third. Every defect below was found by a reviewer, not the author.

### The two that matter — both in `project_public()`

**1. Denial `detail` was forwarded verbatim, so the key shape re-identified the blurred reason.**
`SCOPE_FORBIDDEN`, `SCOPE_MISMATCH`, `OWNERSHIP_CHANGED`, `LEASE_STALE` and
`TARGET_NOT_FOUND_OR_FORBIDDEN` all project to one public reason — but each carried a different detail key
set (`ownership_generation` vs `lease_generation` vs `policy_revision` vs nothing), so a caller could
recover the exact internal reason from the keys alone. The projection table was being defeated through a
side channel.

**2. The first fix was incomplete — and this is the interesting one.** Emitting a *fixed key set* stopped
the keys from varying, but not the **values**. `SCOPE_FORBIDDEN` supplies a real `context_id`;
`TARGET_NOT_FOUND_OR_FORBIDDEN` supplies none. So the "safe" echoed `context_id` was a real ID for one
member of the bucket and `None` for another — the same defect one level down, in a fix specifically
written to close it.

The resolution is that **the public projection now carries no detail at all**. Echoing anything requires
proving it cannot co-vary with the reason, and that argument was got wrong twice by two different models.
Emitting nothing requires no argument. Nothing consumes this field, so the cost is zero.

The test was also wrong in an instructive way: it passed *identical* detail to every reason, which
manufactured the very uniformity it was meant to be checking. It now gives each reason different detail
mirroring what `validate()` actually produces, freezes the five members of the
`TARGET_NOT_FOUND_OR_FORBIDDEN` bucket, and asserts they project to **literally equal** objects rather
than merely similar ones.

Both fixes are mutation-tested: reverting to the original bug fails the test, and so does reverting to the
incomplete first fix. A test that cannot fail proves nothing.

### The other two

**3. The ordering test froze an order that contradicted #473.** It had `count_candidates` and
`rank_candidates` running *before* `resolve_authorized_targets`. Since #473 places target resolution at
step 3 and query/ranking at step 5 — and its acceptance says filtering happens before ranking and limits
in every retrieval arm — counting or ranking unresolved candidates discloses existence and count for
targets the principal may not be entitled to. Given that this corpus exists so that #473 implements against
a fixed target, freezing the wrong order would have propagated the leak into the enforcement layer.

It now freezes: authorize operation → resolve authorized targets → open collection → re-authorize stored
scope → count/rank → hydrate → issue handle → audit. The test is scoped and named for #473 steps 2–6,
because step 1 is `validate()` — a module function, not a consumer method — which is asserted as a
precondition rather than implied by the test's title.

**4. `Decision` and `PublicDecision` were `frozen=True` but unhashable**, because `detail` is a
`mappingproxy`. Frozen dataclasses advertise hashability, so an implementer deduping or memoizing decisions
would have hit a runtime crash on a contract described as frozen. Both now hash explicitly, detail values
are normalized to exact primitives (`isinstance` admits subclasses, and a subclass can set
`__hash__ = None` or redefine it), and `PublicDecision` sanitizes in `__post_init__` since it is exported
and can be built from a mutable dict.

### Acknowledged and deliberately not addressed

- Naive datetimes are assumed UTC rather than rejected (`model.py:_as_datetime`, `validation.py:_utc`).
  Flagged, not changed — nothing in the contract requires rejection.
- The ordering test freezes a **reference fake, not production code**. That is inherent to this PR being
  inert and is stated plainly in the description; binding real retrieval arms to the order is #473's job,
  not something this PR can honestly claim.

100yenadmin:
Closing in favour of **#504**, LCM Teams v1 mono-PR — its three commits are commits 1-3 of that stack.

PR #500, opened 2026-08-05
fix: share SQLite storage across LCM engine clones

### LINKED ISSUE #463: fix: agent clones multiply SQLite file descriptors
## Expected behavior

`LCMEngine.clone_for_agent()` should create independent per-agent/session runtime state without opening a new physical set of SQLite connections for every clone. Shutting down one engine should not invalidate storage still used by live clones, and the physical helpers should close after the final engine releases them.

## Actual behavior

Each clone currently deep-copies the engine and constructs its own `MessageStore`, `SummaryDAG`, and `LifecycleStateStore`. On SQLite/WAL this multiplies database and WAL file descriptors for every cached Hermes agent.

In a long-lived Hermes Gateway with a macOS soft `RLIMIT_NOFILE` of 256, clone/cache churn can therefore exhaust the process FD budget and cause unrelated operations, including subprocess creation, to fail with `Errno 24: Too many open files`.

Closing clone-local helpers after the fact is also less reliable than avoiding the connection churn: on macOS, SQLite/POSIX locking can retain main database descriptors even after WAL descriptors are released.

## Minimal reproduction

1. Create one `LCMEngine` using a SQLite database.
2. Record the process FD count.
3. Call `clone_for_agent()` 12 times while retaining the clones.
4. Record the process FD count again.
5. Inspect the descriptors for the LCM database and its `-wal`/`-shm` files.

Observed locally before remediation:

```text
12 clones: +72 file descriptors (approximately 6 per clone)
```

A representative probe can use `len(os.listdir("/dev/fd"))` before and after constructing the retained clones.

## Version and branch

`main` at `8cf25bfb5bcf03a2163c6ce15bbd4e6599a7008f`

## Environment

- macOS 26.5.2
- Python 3.11.15
- Hermes Agent v0.19.0
- SQLite WAL mode
- Gateway soft `RLIMIT_NOFILE`: 256

## Relevant config

```yaml
context:
  engine: lcm
```

## Logs, traceback, or failing test

Historic Gateway failure under FD pressure:

```text
OSError: [Errno 24] Too many open files
```

Representative live descriptor samples showed multiple handles for each shared database:

```text
lcm.db:      31
lcm.db-wal:  27
state.db:    18
state.db-wal:17
```

The LCM-specific clone probe isolated the linear increase to approximately six descriptors per clone.

## Validation attempted

A local prototype replaces per-clone physical helpers with a reference-counted shared storage bundle while preserving independent session/cursor/runtime state.

Observed with the prototype:

```text
baseline=18
after_clones=18
after_owner_first=18
after_last_clone=11
```

This demonstrates:

- 12 live clones add no physical SQLite descriptors;
- owner-first shutdown does not break surviving clones;
- the final lease closes the shared physical helpers.

Regression coverage also exercises repeated shutdown, owner/clone profile rebinding, configured database paths, cross-bundle locking during helper initialization, concurrent clone writes, and profile-local externalization homes.

Focused engine/core validation passes apart from one pre-existing timing-sensitive SQLite writer-lock threshold test.

## Proposed direction

Use a reference-counted storage bundle for the physical SQLite-backed helpers:

- clones acquire a lease on the same helper set and storage lock;
- mutable session/runtime state remains clone-local;
- shutdown releases one lease and closes helpers only after the final lease;
- profile rebinding detaches the rebinding engine to a new bundle;
- when a configured database path remains the same across bundles, retain one shared reentrant storage lock and hold it throughout helper construction/initialization.

grantjayy:
## Summary

LCM engine clones currently construct a fresh set of SQLite-backed helpers. A clone therefore pays database initialization cost again and retains another set of SQLite file descriptors even though it is serving the same logical database as its prototype.

This change gives each clone family one reference-counted storage bundle while keeping mutable engine/session/model state clone-local. Independently constructed engines still own separate bundles, but bundles resolving to the same canonical database path share one in-process reentrant lock for helper construction, operations, backup/commit, and final teardown.

Fixes #463.

## Why this is needed

Hermes creates engine clones for child agents. Before this change, ten retained clones added 60 file descriptors on the measured macOS checkout and spent roughly 100 ms constructing duplicate SQLite helper sets. That multiplication raises the risk of descriptor exhaustion and repeated schema/WAL setup under parallel child-agent workloads.

The ownership boundary is the important part of the fix:

- clones share only the storage helpers;
- session IDs, cursors, runtime counters, model/provider metadata, adaptive retrieval state, and assertion extractors remain engine-local;
- every engine owns exactly one idempotent lease;
- owner-first shutdown releases one lease without closing storage still used by clones;
- final release attempts every helper close and reports combined failures;
- abandoned clones release their lease through finalization;
- failed clone construction rolls its acquired lease back.

This is complementary to #470's cleanup work. It does not replace #486's cross-process FTS bootstrap locking; the lock introduced here is process-local.

## Implementation

- Add a reference-counted `_SharedStorage` bundle for message store, summary DAG, lifecycle state, optional assertion store, and optional query-view store.
- Pass the bundle into `clone_for_agent()` and preserve clone-local runtime/model state.
- Use a weak canonical-path lock registry so independently constructed bundles for the same database serialize helper initialization, operations, backup/commit, and teardown without sharing mutable bundle ownership.
- Allow each SQLite helper to adopt the shared reentrant lock.
- Synchronize every touched store/DAG operation, lifecycle debt cleanup, backup, close, and commit path that uses a shared connection.
- Add lifecycle, concurrency, rollback, finalization, profile-rebinding, aggregate-cleanup, and descriptor-regression tests.
- Add `scripts/measure_clone_storage.py` for reproducible before/after measurements against a selected checkout.

## Reproduced benchmark

Environment: macOS, Python from the active Hermes environment, 25 samples, 10 retained clones. The benchmark imports each selected checkout using `LCM_BENCH_REPO_ROOT`.

Commands:

```bash
LCM_BENCH_REPO_ROOT=/private/tmp/hermes-lcm-baseline-check-20260804 \
  python scripts/measure_clone_storage.py --samples 25 --clones 10 --json

LCM_BENCH_REPO_ROOT=/private/tmp/hermes-lcm-shared-storage-impl-20260804 \
  python scripts/measure_clone_storage.py --samples 25 --clones 10 --json
```

Commits:

- Base: `6b7dbb1`
- Feature: `1708241c1ce9c759348509f9148b60e0f9fc02ff`

Results:

- Retained-clone file-descriptor delta: `+60` → `0`
- Median clone setup: `5.877985 ms` → `0.198782 ms`
- Ten-clone setup: `100.084453 ms` → `2.203067 ms`
- Median initial startup: `6.731055 ms` → `7.368695 ms`

Raw local receipts:

- `/private/tmp/lcm-storage-benchmark-final-base-20260804.json`
  - SHA-256 `2bcd2a9430f2d63a5b4f7ef00a9386157a9475e27825d935c44e03d3d3c09c06`
- `/private/tmp/lcm-storage-benchmark-final-feature-20260804.json`
  - SHA-256 `ba969980fa6087ded140a811b1bcf91a2b725e9f64aae536f1484f7cc84a7626`

## Verification

Green feature/affected gates:

- `tests/test_lcm_core.py::TestLCMEngineSharedStorage`: 9 passed
- `tests/test_lcm_engine.py`: 751 passed, 1 skipped
- `tests/test_lcm_core.py` excluding one independently reproduced upstream failure: 316 passed
- `tests/test_packaging_install.py`: 38 passed in the grouped run before the known isolated baseline failure was reproduced
- `tests/test_assertion_lifecycle_tools.py tests/test_assertion_store.py tests/test_query_view_store.py`: 49 passed
- `tests/test_lcm_core.py::TestLifecycleStateStore`: 19 passed
- `tests/test_assertion_store.py`: 20 passed
- `tests/test_query_view_store.py`: 21 passed
- Ruff on every changed Python file: passed
- Python compilation on changed production/benchmark files: passed
- `git diff --check`: passed
- Independent adversarial concurrency/lifecycle review: PASS after two repair rounds

Full-suite result:

- The suite remains red on current `main` for pre-existing or load-sensitive tests. The observed failures were reproduced against pristine `6b7dbb1`, including strict wall-clock embedding deadlines, macOS `/var` versus `/private/var` containment assertions, trajectory token/chunk expectations, provider-routing behavior, and an isolated packaging registration test.
- Three doctor failures initially exposed an incompatible post-shutdown alias-clearing change. That change was reverted; all three doctor tests pass on the final feature commit.
- No remaining full-suite failure was unique to this feature branch in the bounded base comparison.

## Scope

This PR intentionally does not:

- share one bundle across independently constructed engines;
- provide cross-process locking;
- change session/runtime/model ownership;
- change public configuration;
- include the separate reasoning-control or preflight-maintenance features.

### DIFF
--- a/.plans/2026-08-04-lcm-upstream-feature-prs.md
+++ b/.plans/2026-08-04-lcm-upstream-feature-prs.md
@@ -135,9 +135,13 @@ Request:
 
 ### Existing issue
 
-Use and close #463: `fix: agent clones multiply SQLite file descriptors`.
+Link #463: `fix: agent clones multiply SQLite file descriptors`.
 
-Include these measured receipts:
+Close #463 only if the final pull request covers the issue's full accepted direction. The local clone-family bundle does not yet prove cross-bundle construction locking for independently created engines that share one database path. Either add that behavior and its concurrency tests, obtain maintainer agreement that the narrower clone-family fix resolves the issue, or leave #463 open with the remaining work stated clearly.
+
+Re-run a checked-in or pull-request-attached clone benchmark against both the pinned upstream base and the final feature branch. The benchmark must measure retained-clone file descriptors, median clone setup, ten-clone setup, and initial startup. Record its exact command, environment, raw output path, and both commit SHAs.
+
+Historical private results provide targets, not publishable claims:
 
 - Before the prototype, 12 retained clones added about 72 file descriptors.
 - With the prototype, 12 retained clones added no physical SQLite descriptors.
@@ -145,19 +149,22 @@ Include these measured receipts:
 - Ten-clone setup improved from 91.63 ms to 1.96 ms.
 - Initial plugin startup stayed effectively flat.
 
+Publish only measurements reproduced by the final benchmark run. Do not copy the historical private timings into the issue or pull request unless the new receipt confirms them.
+
 ### Pull request
 
 **Title:** `fix: share SQLite storage across LCM engine clones`
 
-**Source:** storage, locking, lifecycle, and clone-test hunks from `17df35d`; fold `5f0832b` cleanup.
+**Source:** storage, locking, lifecycle, and clone-test hunks from `17df35d`. Do not replay `5f0832b`; the commit only repairs import pollution introduced by the mixed local commit.
 
 **Files:**
 
 - Modify `engine.py`.
 - Modify `store.py`.
 - Modify `dag.py`.
-- Modify `lifecycle_state.py` only for the clean import shape.
 - Modify `tests/test_lcm_core.py`.
+
+Keep the net `lifecycle_state.py` diff empty unless current upstream analysis proves a required storage behavior change. Do not carry the unused `functools.wraps` import or unrelated comment deletions from the local range.
 
 ### Behavior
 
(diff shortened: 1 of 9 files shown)


grantjayy:
Superseded before review: the first head branch accidentally included two local planning-only commits. Reopening from a clean branch based directly on current main; production and test commit content is unchanged.

PR #506, opened 2026-08-05
fix: share LCM storage across runtime clones

jtstothard:
## Summary
- Add a reference-counted `_StorageBundle` for the SQLite helpers shared by cloned LCM runtimes.
- Adopt the shared bundle in `clone_for_agent()` and release it during shutdown, while preserving per-agent session state.
- Add FD-hygiene coverage for 50 clones and update clone identity assertions.
- Make the plugin test bootstrap locate the Hermes host module from a standalone worktree.

This is the distinct LCM storage-lifecycle half of the descriptor-lifecycle work. It is not the generic `SessionDB` fix in hermes-agent#72822.

## Tracking
- hermes-agent#79537 — descriptor-lifecycle implementation
- hermes-agent#79538 — plugin packaging
- hermes-agent#72782 — parent work item
- Distinct from hermes-agent#72822 — generic SessionDB fix

## Verification
- `python3 -m pytest -q tests/test_fd_hygiene.py` — 1 passed
- `python3 -m pytest -q tests/test_lcm_core.py` — 308 passed
- `python3 -m compileall -q engine.py tests/test_fd_hygiene.py tests/test_lcm_core.py` — passed
- `git diff --check` — passed

## Scope
Fresh branch from `origin/main`; only `engine.py`, `tests/conftest.py`, `tests/test_lcm_core.py`, and `tests/test_fd_hygiene.py` are included. No generated files or unrelated existing work are included.

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -360,6 +360,34 @@ _PRESERVED_TODO_CONTEXT_PREFIX = "[Your active task list was preserved across co
 _LCM_MESSAGE_PREFIX_FINGERPRINT_LIMIT = 8
 
 
+class _StorageBundle:
+    """Reference-counted SQLite helpers shared by engine runtime clones."""
+
+    def __init__(self, *helpers: Any) -> None:
+        self.helpers = tuple(helpers)
+        self._lock = threading.Lock()
+        self._refs = 1
+
+    def acquire(self) -> "_StorageBundle":
+        with self._lock:
+            if self._refs <= 0:
+                raise RuntimeError("cannot acquire a closed LCM storage bundle")
+            self._refs += 1
+        return self
+
+    def release(self) -> None:
+        with self._lock:
+            if self._refs <= 0:
+                return
+            self._refs -= 1
+            if self._refs:
+                return
+        for helper in self.helpers:
+            close = getattr(helper, "close", None)
+            if callable(close):
+                close()
+
+
 class LCMEngine(CompactionMixin, ResetStateMixin, ReconcileMixin, AuxiliarySessionMixin, PlaceholderLedgerMixin, BypassMixin, ContextEngine):
     """Lossless Context Management engine.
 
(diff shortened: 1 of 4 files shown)


jtstothard:
@stephenschoettler maintainer review requested. The repository owner and recent merger of the latest releases/fixes, you appear to be the appropriate reviewer for this PR.

Please review the _StorageBundle ownership/refcounting, run or enable CI, and advise whether the dependency on hermes-agent PR #79575 should be handled as-is or adjusted before merge.

jtstothard:
Pre-review finding (likely blocker): `clone_for_agent()` calls `_adopt_storage_bundle()` before constructing the optional `AdaptiveRetrievalRegistry` and `ModelAssertionExtractor`. If either constructor raises, the partially constructed clone is not shut down/released, leaving `_StorageBundle._refs` elevated and leaking the shared storage lifetime. Please wrap post-adoption clone initialization in `try/except` and call `clone.shutdown()` (or release the bundle directly) before re-raising; add a regression test that forces each optional constructor to fail and asserts the prototype refcount returns to its prior value.

jtstothard:
Additional pre-review blocker: this branch appears not to implement the explicit ContextEngine.create_runtime() / close() contract introduced by hermes-agent PR #79575. Once that core loader contract is active, the LCM engine may be rejected for lacking an isolated runtime factory, and inherited close() will not release the shared storage bundle. The plugin PR should implement the contract explicitly, likely mapping current clone/runtime construction to create_runtime() and _close_storage() to close(), with an integration test against the updated core loader.\n\nThe review also flagged a second major issue: _StorageBundle.release() stops if one helper close() raises, so later helpers can remain open. Close each helper independently while preserving and logging the first exception.

jtstothard:
Final review found one remaining blocker at head 98b259af290bfb40337e333806d53ddfd03cd39e: repeated runtime.close() is not safely idempotent. A second close can invalidate shared storage, causing a still-live prototype to fail on subsequent database use. The fix needs per-runtime one-shot release semantics plus a regression test that closes the same runtime twice and then exercises the prototype/sibling runtime. A focused repair is underway; all other lifecycle, FD, portability, and test concerns passed.

jtstothard:
Final independent approval review completed at head 95cafc4fe49a1c2ae4dca50f332b3d768cd3ec92: no remaining blockers. Verified full pytest (2799 passed, 1 skipped, 12 xfailed), focused lifecycle/core (312 passed), FD hygiene (1 passed), compileall, and diff checks. Verified repeated close is idempotent; prototype and sibling remain usable; bundle refcount and helper close counts are correct; failed-construction cleanup, rebind, concurrency, optional helpers, runtime contract compatibility with hermes-agent PR #79575, portable bootstrap, and bounded FD growth all pass. Ready for maintainer review/merge.

PR #516, opened 2026-08-08
fix: gate opportunistic maintenance compaction on token pressure

Kyzcreig:
## Summary
- Gate the two OPPORTUNISTIC maintenance-compaction arms on the divergent-replay path behind a token floor, so a media-heavy session stops requesting compaction at any size.
- Add `maintenance_min_pressure_ratio` (`LCM_MAINTENANCE_MIN_PRESSURE_RATIO`), a fraction of `threshold_tokens`. **Default `0.0` preserves current behavior exactly.**

## Why

`should_compress_preflight` reaches the leaf-candidate check two ways, and only one is gated.

Non-divergent path — gated:

```python
if self.threshold_tokens > 0 and rough >= self.threshold_tokens:
    eligible, reason = self._leaf_compaction_candidate_status(messages, ...)
    if eligible:
        return self._mark_preflight_compression_requested()
```

Divergent-replay path — the identical check, no token gate:

```python
eligible, reason = self._leaf_compaction_candidate_status(replay_messages, ...)
if eligible:
    return self._mark_preflight_compression_requested()
if self._has_ignored_backlog_outside_fresh_tail(replay_messages):
    return self._mark_preflight_compression_requested()
```

That path is entered whenever ingest rewrote the replay view — in practice an externalized payload: inline media, base64, a large tool result. A session that traffics in those requests an opportunistic maintenance compaction however small it is.

Observed against a 750,000-token threshold:

| observed tokens | % of threshold |
|---|---|
| 129,088 | 17% |
| 158,262 | 21% |
| 175,730 | 23% |
| 239,896 | 32% |

Three of those landed inside eight minutes, one on a session that had just been compacted. Those passes freed **27–47%**, where pressure-driven compactions on the same deployment freed **84–86%** — and each still costs a summarizer call plus the conversation's prompt cache.

Deliberately **not** gated: deterministic ingest-cleanup adoption returns earlier and stays available at any size (already durable, no summarizer work). Overflow recovery is checked before the floor on every path.

## Validation
- [x] Focused validation: `pytest tests/test_maintenance_pressure_floor.py -q` -> `11 passed`
- [x] RED proof: reverting both call sites -> `2 failed, 9 passed` (`test_both_opportunistic_arms_are_gated`, `test_cleanup_adoption_is_not_gated`)
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `1108 passed, 1 skipped` (incl. the new file)
  - [x] `pytest -q` -> `7 failed, 2797 passed, 2 skipped, 12 xfailed` — see Notes
  - [x] `bash -lc 'ulimit -n 1024 && pytest -q'` -> same 7, all environmental
  - [x] `python -m compileall -q .` -> exit 0
  - [ ] `python -m py_compile scripts/import_lossless_claw.py` — not run, no scripts touched
  - [ ] `bash -n scripts/install.sh scripts/update.sh` — not run, no shell scripts touched
  - [x] `ruff check .` (pinned 0.15.13) -> `All checks passed!`
- [x] Workflow validation: n/a, no workflows changed

## Notes

**The 7 remaining failures are environmental on macOS + Python 3.14 and are not caused by this change.** Verified against a pristine `upstream/main` worktree, which fails **10** in the same two families — a superset, with shifting membership run to run:

- **Wall-clock deadline assertions** — `test_embedding_provider.py`, `test_embedding_search_modes.py` assert real elapsed time (e.g. `assert elapsed < 0.06`). Membership varies between runs, so this is a flake family rather than a fixed list.
- **macOS `/private` symlink** — `test_path_containment.py`, `test_path_security.py`. `TemporaryDirectory()` returns `/var/...` while the resolved path is `/private/var/...`, so a `startswith` containment check fails. Linux CI is unaffected.

Neither family touches compaction. Claiming *their-suite-minus-known-environmental-reds*, not full green.

Two of the 11 tests exist specifically to stop the fix going half-inert: gating only one of the two arms leaves the other free to request compaction at any size, and the cleanup-adoption ordering assertion pins that deterministic cleanup keeps working below the floor. Both fail if the change is reverted.

The floor ships **off**. Operators who want it set `LCM_MAINTENANCE_MIN_PRESSURE_RATIO=0.5` (or whatever fraction suits their workload).

All values above are from synthetic/local runs — no transcripts, credentials, or `lcm.db` contents are included.

### DIFF
--- a/compaction.py
+++ b/compaction.py
@@ -66,6 +66,41 @@ class CompactionMixin:
             return False
         return tokens >= self.threshold_tokens
 
+    def _maintenance_pressure_met(self, observed_tokens: int) -> bool:
+        """Whether an OPPORTUNISTIC maintenance compaction is allowed right now.
+
+        The maintenance arms ("compactable backlog outside the fresh tail",
+        "ignored-message backlog") are tidying, not overflow protection. On the
+        divergent-replay path they run with no token gate, while the identical
+        leaf-candidate check on the non-divergent path sits behind
+        ``rough >= threshold_tokens``. Since the divergent path is entered
+        whenever ingest rewrote the replay view -- typically an externalized
+        payload (inline media, base64, a large tool result) -- a media-heavy
+        session requests compaction at any size.
+
+        Returns True (allow) when the floor is disabled, when no threshold is
+        configured, or when the session has reached the configured fraction of
+        it. Deterministic ingest-cleanup adoption returns earlier and is
+        unaffected; overflow recovery is checked before this on every path.
+        """
+        ratio = getattr(self._config, "maintenance_min_pressure_ratio", 0.0) or 0.0
+        if ratio <= 0.0:
+            return True
+        if self.threshold_tokens <= 0:
+            return True
+        floor = int(self.threshold_tokens * ratio)
+        if observed_tokens >= floor:
+            return True
+        logger.debug(
+            "LCM maintenance compaction deferred: %d tokens < %d floor "
+            "(%.0f%% of %d threshold)",
+            observed_tokens,
+            floor,
+            ratio * 100,
+            self.threshold_tokens,
+        )
+        return False
+
     def should_compress_preflight(self, messages):
         """Pre-flight check — also ingests messages into the store."""
         self._preflight_cleanup_only_due_to_boundary_cooldown = False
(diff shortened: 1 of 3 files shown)


Kyzcreig:
Hi — this PR has been open for about 38 days without anyone taking a look yet. It's not a draft and currently shows as mergeable, so it should be straightforward to open up. Would someone be able to give it a quick look when there's a spare moment? I'm not asking for a merge decision right now, just some eyes on it so it's on the radar. Happy to rebase, add context, or split it up if that would make reviewing easier — just let me know.

PR #519, opened 2026-08-09
feat: add OpenAI-compatible embedding provider

### LINKED ISSUE #419: feat: Add native openai embedding provider option
### Problem

With the rollout of the vector store and the addition of the Ollama provider, the infrastructure for handling external embedding HTTP payloads is already in place. It would be a massive workflow improvement to add an official openai provider option to target standard, openai-compatible local servers (such as llama.cpp server instances running specialized local embedding models).

### User or operator impact

This is resource saving for local inference users who also have local embedding models already deployed on local workstation or server. For example if someone is already running llama.cpp or vllm hosting an embedding model natively on a specific port (e.g. localhost:8081/v1/embeddings) for RAG or any other vector database they could use that instead of needing a completely separate instance of Ollama or background python environments just for embedding tasks for hermes_lcm.

Because hermes-lcm elegantly handles provider switching and vector store backfilling (/lcm embed backfill), allowing the plugin to natively communicate with the industry-standard openaI specification opens up immediate compatibility for a vast array of high-performance local inference backends.

### Proposed direction

Since the networking code and polling loops are already written to handle Ollama's HTTP layer, bridging to the OpenAI spec is architectural low-hanging fruit:

It simply requires routing payloads to the conventional /v1/embeddings endpoint instead of /api/embeddings.

Adjusting the POST body map from Ollama's "prompt" string key to the standard openai "input" string/array key.

### Alternatives and tradeoffs

Running all of the other options does not take advantage of resources available for cases where local openai embedding is available, while requiring maintenance and compute/memory for another embedding model of lower quality that what can be running via local openai instance (e.g. one of the Qwen3 embedding models), or even cloud OpenAI embedder.

### Backward compatibility

There is no compatibility issue as it is specifically provided within hermes_lcm backfill function.

### Validation or evidence

Rough config schema for this option:

export LCM_EMBEDDINGS_ENABLED=true
export LCM_EMBEDDING_PROVIDER=openai
export LCM_EMBEDDING_BASE_URL=http://localhost:8081/v1
export LCM_EMBEDDING_MODEL=your-chosen-embedding-model

### Implementation status

Idea only

### Additional notes

While not a developer I'm happy to help with the testing.

Vocllum:
## Summary

Adds an `OpenAICompatibleProvider` to the embedding stack, alongside the existing `voyage` / `ollama` / `fastembed` providers. It targets any OpenAI-format `/v1/embeddings` endpoint — llama.cpp server, vLLM, SiliconFlow, or a private gateway — so operators who already run an embedding service can point LCM at it instead of running a second embedding stack.

## Motivation

Issue #419 asks for exactly this: with the vector store and the Ollama provider in place, the HTTP embedding infrastructure already exists; a provider-neutral OpenAI-compatible option closes the gap for everyone with a local or cloud OpenAI-format embeddings endpoint.

## Implementation

- `embedding_provider.py`: new `OpenAICompatibleProvider(_ResilientProvider)` modeled on `OllamaProvider` — `_request` / `_embed` / `embed_documents` / `embed_document_batches` / `embed_query` / `embed_query_interactive`, dimension tracking, deadline budgets, and the same spend-guard / circuit-breaker semantics as the other remote providers.
- `resolve_provider()` accepts `openai-compatible` (aliases `openai`, `siliconflow`); unsupported-provider error text updated.
- `config.py`: two new env fields — `LCM_EMBEDDING_BASE_URL` (API root; trailing slash stripped, `/embeddings` appended) and `LCM_EMBEDDING_API_KEY` (falls back to `SILICONFLOW_API_KEY`).

## Env contract

```bash
LCM_EMBEDDING_PROVIDER=openai-compatible
LCM_EMBEDDING_MODEL=BAAI/bge-m3     # any model id your endpoint serves
LCM_EMBEDDING_BASE_URL=https://api.siliconflow.cn/v1
LCM_EMBEDDING_API_KEY=sk-...        # falls back to SILICONFLOW_API_KEY
```

Request shape is the standard OpenAI one (`{"model": ..., "input": [...]}` with a `Bearer` header). Deadline and identity semantics match the other remote providers: `LCM_EMBEDDING_QUERY_TIMEOUT_S` bounds interactive queries, `LCM_EMBEDDING_BACKFILL_TIMEOUT_S` bounds backfill batches, and vectors are stored under the full `(provider, model, ...)` identity so switching back reactivates existing vectors without re-backfilling.

## Tests & docs

- 9 new unit tests in `tests/test_embedding_provider.py` (request shape, auth header, trailing-slash handling, query/interactive paths, missing-key and fallback behavior, base-url/model validation, network-error non-retry, response-count mismatch, `resolve_provider` wiring).
- `docs/embeddings-setup.md`: new "Option 4 — OpenAI-compatible API" section plus TL;DR row.

Closes #419

### DIFF
--- a/config.py
+++ b/config.py
@@ -392,6 +392,12 @@ ENV_FIELD_SPECS: tuple[_EnvFieldSpec, ...] = (
     _EnvFieldSpec("embedding_model", "LCM_EMBEDDING_MODEL", str),
     _EnvFieldSpec("embedding_content_policy", "LCM_EMBED_CONTENT_POLICY", str),
     _EnvFieldSpec("ollama_base_url", "LCM_OLLAMA_BASE_URL", str),
+    _EnvFieldSpec(
+        "embedding_base_url", "LCM_EMBEDDING_BASE_URL", str
+    ),
+    _EnvFieldSpec(
+        "embedding_api_key_env", "LCM_EMBEDDING_API_KEY_ENV", str
+    ),
     _EnvFieldSpec("embedding_query_timeout_s", "LCM_EMBEDDING_QUERY_TIMEOUT_S", float),
     _EnvFieldSpec("recall_query_timeout_s", "LCM_RECALL_QUERY_TIMEOUT_S", float),
     _EnvFieldSpec("embedding_backfill_timeout_s", "LCM_EMBEDDING_BACKFILL_TIMEOUT_S", float),
@@ -710,6 +716,8 @@ class LCMConfig:
     # default in the chunker's normalize_content_policy.
     embedding_content_policy: str = "conversational"
     ollama_base_url: str = "http://localhost:11434"
+    embedding_base_url: str = ""
+    embedding_api_key_env: str = "LCM_EMBEDDING_API_KEY"
     embedding_query_timeout_s: float = 3.0
     # Dedicated deadline for lcm_recall. It fans out three sequential arms (FTS +
     # summary KNN + chunk KNN) plus fusion, hydration, and an optional rerank, so
--- a/docs/embeddings-setup.md
+++ b/docs/embeddings-setup.md
@@ -11,6 +11,7 @@ FTS-only exactly as before.
 | Voyage AI | free tier (200M tokens for the voyage-4 group), then ~$0.02–0.12 per million tokens | yes (no credit card) | none | best quality, zero local footprint |
 | fastembed | $0 | no | `pip install fastembed` (ONNX, no torch) | local default — no accounts, no daemon |
 | Ollama | $0 | no | Ollama app/daemon | you already run Ollama |
+| OpenAI-compatible | provider pricing | yes (per provider) | none | reuse an existing `/v1/embeddings` service (SiliconFlow, llama.cpp, vLLM) |
 
 All providers feed the same store. Switching provider/model is one config change plus a backfill;
 each full identity keeps its own vectors, so switching **back** to a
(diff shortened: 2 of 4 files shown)


Rutimka:
Hi @Vocllum — reviewed #519 (commit `a2fe642`) while building a rerank
sister feature on top of it (stacked PR incoming). Overall: clean house
style, the provider follows the `_ResilientProvider` patterns correctly,
and the SiliconFlow fallback key is a nice touch. Two things worth a look:

1. **Hard requirement on an API key for self-hosted endpoints.**
   `_request` raises when neither `LCM_EMBEDDING_API_KEY` nor
   `SILICONFLOW_API_KEY` is set — even for LAN-local endpoints that don't
   check auth at all (TEI, vLLM, Infinity). Workable today with a dummy
   key, but an "empty key → send without `Authorization` header" path
   (like Voyage's explicit auth error, inverted) would make the provider
   friction-free for self-hosted use. Not a blocker — just something we
   ended up needing for the rerank side.

2. **`uv.lock` says `requires-python = ">=3.13"`** while CI's test matrix
   runs 3.11–3.14. CI installs pytest/numpy directly (no uv), so nothing
   breaks — but the lock's floor is stricter than the declared support
   matrix. Might be worth aligning (or dropping the lock from the PR if
   it was regenerated incidentally).

Neither blocks us — the branch works as-is as a base. Happy to help with
follow-ups if useful.

Vocllum:
Closing this source PR because hermes-lcm is no longer the maintained integration path and the migration PR has landed in the successor path. Keeping this open would only leave stale implementation work against an inactive target.

Vocllum:
Closing as superseded. This repository is no longer the maintained integration target, and the migration PR on the successor path has already landed. Keeping this implementation open would only create stale parallel work. Thanks for the earlier review and feedback.

PR #520, opened 2026-08-09
feat: add LCM_DISABLED_TOOLS env filter for context-engine tools

Vocllum:
## Summary

Adds an opt-in `LCM_DISABLED_TOOLS` environment variable that removes named `lcm_*` tools from the model-facing surface, so their full JSON schemas are not injected into every turn. This is a surgical, reversible alternative to disabling the whole context-engine toolset.

## Motivation

All 15 context-engine tool schemas are injected every turn, including heavy advanced tools (`lcm_compute`, `lcm_compile_evidence`, `lcm_evidence_pack`, `lcm_retrieve`, `lcm_query_state`) that many operators never call. On small-context routes those schemas are a meaningful fixed per-turn cost (~3-4k tokens for the advanced set). Operators who want the compression engine but not the evidence/assertion surface now have a per-tool switch instead of all-or-nothing.

## Implementation

One env var, three enforcement points, all reversible:

- `engine.get_tool_schemas()` — disabled tools are excluded from the injected schema list (zero tokens per turn).
- `engine.handle_tool_call()` — disabled tools are refused **before** message ingest, returning a clear error: `LCM tool <name> is disabled via LCM_DISABLED_TOOLS`.
- plugin `register()` — disabled tools are filtered from Path A plugin-registry registration, mirroring the engine so they cost nothing on every host path.

```bash
LCM_DISABLED_TOOLS=lcm_compute,lcm_compile_evidence,lcm_evidence_pack,lcm_retrieve,lcm_query_state
```

Unset the variable to restore all 15 tools. Unknown/whitespace entries are ignored; no tool names are hardcoded into the filter itself.

## Tests & docs

- New `tests/test_disabled_tools.py` (5 tests): full 15-tool baseline without env, 10-tool filtered set with env, refusal in `handle_tool_call` before ingest, enabled tools still dispatch, and env parsing (trimming, empties, unset).
- README environment-variable table row for `LCM_DISABLED_TOOLS`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,7 @@ moved back to that assistant even when doing so exceeds a configured bound.
 | `LCM_EMPTY_LIFECYCLE_GC_ENABLED` | `true` | Master toggle for automatic pruning of lifecycle rows for sessions that never ingested any messages or summary nodes |
 | `LCM_EMPTY_LIFECYCLE_GC_THRESHOLD` | `200` | Number of lifecycle rows at which the GC pass fires |
 | `LCM_EMPTY_LIFECYCLE_GC_MAX_AGE_HOURS` | `24` | Automatic GC only deletes empty lifecycle rows at least this old; set `0` only in trusted/test environments that intentionally want immediate empty-row pruning |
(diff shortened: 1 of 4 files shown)


Vocllum:
Closing this source PR because hermes-lcm is no longer the maintained integration path and the migration PR has landed in the successor path. Keeping this open would only leave stale implementation work against an inactive target.

Vocllum:
Closing as superseded. This repository is no longer the maintained integration target, and the migration PR on the successor path has already landed. No further work is planned on this branch.

Issue #408, opened 2026-07-17
fix: LCM Externalization Fails on Windows with Errno 13 Permission Denied


azrael-229:
### Expected behavior

Externalization completing successfully, empty error logs

### Actual behavior

Error pops up in logs:

## Error Details
```
hermes_plugins.hermes_lcm.externalize: Large payload externalization skipped (non-blocking) [Errno 13] Permission denied: 'C:\\Users\\<username>\\AppData\\Local\\hermes\\lcm-large-outputs'
```

### Minimal reproduction

1. Run Hermes Agent on Windows with LCM plugin enabled
2. Trigger large payload externalization (e.g., paste multiple images, exceed context threshold)
3. Observe warning in logs: `Large payload externalization skipped (non-blocking)`

The same error is also produced by test script lcm_stress_check.py

### Version and branch

v0.19.0; main; up to date as of 17.07.26

### Environment

- **OS**: Windows 11 (build 26200+) - **Python**: 3.11+ (Hermes venv) - **LCM plugin version**: Latest from git - **Hermes Agent version**: v0.17.0 (latest)

### Relevant config

```yaml
LCM_LARGE_OUTPUT_EXTERNALIZATION_ENABLED=true

context:
  engine: lcm
plugins:
  enabled:
    - hermes-lcm
```

### Logs, traceback, or failing test

```text
hermes_plugins.hermes_lcm.externalize: Large payload externalization skipped (non-blocking) [Errno 13] Permission denied: 'C:\\Users\\jakub\\AppData\\Local\\hermes\\lcm-large-outputs'


OR from test script:

LCM ingest protection could not externalize payload at content; preserving inline content for lossless recovery
Large payload externalization skipped (non-blocking): [Errno 13] Permission denied: 'C:\\Users\\jakub\\AppData\\Local\\hermes\\plugins\\hermes-lcm\\scripts\\tests\\sandbox\\lifecycle_soak_and_profile_rebinds\\primary-home\\lcm-large-outputs'
LCM ingest payload externalization skipped (non-blocking): [Errno 13] Permission denied: 'C:\\Users\\jakub\\AppData\\Local\\hermes\\plugins\\hermes-lcm\\scripts\\tests\\sandbox\\lifecycle_soak_and_profile_rebinds\\primary-home\\lcm-large-outputs'
```

### Validation attempted

- ✅ Directory exists at target path
- ✅ User has FullControl permissions on directory
- ✅ Directory is empty (no files inside)
[...]

PR #522, opened 2026-08-10
fix(externalize): skip POSIX directory-fsync on Windows (#408)

fxhsxm:
# Fix: Externalized-payload writes fail on Windows with Errno 13 (#408)

## Root cause

`externalize._fsync_directory()` used the POSIX directory-fsync pattern:

```python
flags = os.O_RDONLY
if hasattr(os, "O_DIRECTORY"):
    flags |= os.O_DIRECTORY
dir_fd = os.open(path, flags)   # <-- PermissionError on Windows
os.fsync(dir_fd)
```

On Windows, `os.open()` cannot obtain a directory handle: opening a
directory via the underlying `CreateFile` requires
`FILE_FLAG_BACKUP_SEMANTICS`, which this `os.open()` path does not set, so
every call raises
`PermissionError: [Errno 13] Permission denied: '...\lcm-large-outputs'`.
The `hasattr(os, "O_DIRECTORY")` guard only protected the flag, not the
`os.open()` call itself.

`_fsync_directory` is invoked on **every** externalized-payload write
(`_write_externalized_payload` / `_replace_externalized_payload`) and after
directory creation, so on Windows **every** large-payload externalization
failed, the payload was silently kept inline, and the `Errno 13` pointed at
the directory path — exactly the symptom in #408.

## Fix

Skip the parent-directory fsync explicitly on Windows (`os.name == "nt"`);
all other platforms keep the original behavior, including the original
capability check on the `O_DIRECTORY` flag. On Windows the payload file
itself has already been fsynced by the caller (`os.fsync` calls
`_commit()` there); the parent-directory fsync is skipped because the API
does not expose a directory handle. Crash-durability semantics therefore
differ from the POSIX path — the payload file is durable, the directory
entry is not fsynced. On POSIX the change is byte-for-byte behavior
preserving: same flags, same open → fsync → close sequence.

## Safety

Fail-open semantics are untouched: if any write step fails, the original
payload stays inline, a warning is logged, and no partial payload file is
left behind. The fix only removes the one operation Windows cannot perform.

## Tests

- New regression tests:
  - `test_fsync_directory_noop_when_directory_handles_unsupported`
    (Windows only: asserts `os.open` is never called and no exception is
    raised; fails on Windows before this fix, passes after)
  - `test_fsync_directory_still_fsyncs_on_posix` (guards the POSIX path)
- Platform-aware adjustments to two POSIX-semantics tests that assumed
  directory fsync / POSIX permission bits always exist:
  - `test_externalized_payload_reassignment_fsyncs_replacement`
  - `test_externalized_payload_files_are_private` (skips on Windows)
- Verified on Windows 11 / Python 3.11:
  - minimal repro of #408: FAIL before (exact Errno 13 on the directory
    path), PASS after
  - engine-level externalization: payload file created, exact SHA256
    recovery, recovery after engine restart, 6 consecutive writes, reused
    directory with pre-existing payloads, fail-open under simulated write
    failure, `lcm_expand` search-prefix reader recovery — all pass
  - Active Replay Stubbing layer: old tool results stubbed to durable
    recoverable refs, protected fresh tail stays inline, fail-open — all pass
  - `lcm_stress_check.py --tier release`: the
    `redaction_and_externalization_boundaries` scenario **fails on the
    unpatched code** (0 payloads externalized, oversized payload persisted
    raw to sqlite rows) and **passes after the fix**; all 6 release-tier
    scenarios pass

## Unrelated follow-up (separate issue)

During Windows validation, six additional pre-existing replay failures were
traced to an independent persisted-output CRLF translation issue (Windows
text-mode `\n` → `\r\n` inflation vs. the marker's character count). This
is unrelated to the directory-fsync failure fixed here and will be handled
separately in #XXX.

### DIFF
--- a/externalize.py
+++ b/externalize.py
@@ -57,6 +57,14 @@ def _preview_sha256(preview_prefix: Any) -> str:
 
 
 def _fsync_directory(path: Path) -> None:
+    if os.name == "nt":
+        # Windows cannot use this POSIX directory-fsync path: opening a
+        # directory requires Windows-specific handle semantics
+        # (FILE_FLAG_BACKUP_SEMANTICS) not exposed by this os.open() call.
+        # The payload file itself has already been fsynced by the caller.
+        # Skip the parent-directory fsync on Windows; crash-durability
+        # semantics therefore differ from the POSIX path.
+        return
     flags = os.O_RDONLY
     if hasattr(os, "O_DIRECTORY"):
         flags |= os.O_DIRECTORY
--- a/tests/test_ingest_protection.py
+++ b/tests/test_ingest_protection.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 import importlib.util
 import json
+import os
 import re
 import sqlite3
 import stat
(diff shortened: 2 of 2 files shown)


azrael-229:
Applied on my deployment. Fix verified on Windows 11.

supplefrog:
Current-main follow-up: #563. It keeps this PR's direct Windows directory-`fsync` guard and credit, then adds current-main tests, Windows CI, CRLF recovery, and docs. I opened it because this draft had diverged substantially from the fast-moving main branch.

PR #528, opened 2026-08-14
fix: close finalized per-agent LCM clones

bennybuoy:
## Summary

- register the supported `on_session_finalize` plugin hook on current `main`
- resolve the exact per-agent LCM clone by durable session ID
- shut down only that finalized clone, leaving other active clones and the plugin prototype untouched
- preserve interactive CLI `/new`, which reuses its agent and context engine

## Why

Hermes deep-copies the registered context-engine prototype for each `AIAgent`. In long-lived Desktop and gateway processes, retiring an agent does not guarantee that the clone's SQLite-backed LCM helpers are closed. Finalized sessions can therefore leave Store, DAG, and lifecycle descriptors open.

The plugin already owns an exact weak session-to-engine registry and an idempotent `shutdown()` boundary. This change connects those existing pieces through Hermes' supported hard session-boundary hook; it does not patch Hermes core or sweep unrelated engines.

This refreshes #470 on current `main`. It is complementary to #501: shared clone storage reduces per-clone SQLite ownership, while this hook closes the finalized clone's own lease/resources at the host lifecycle boundary.

## Validation

- [x] Focused lifecycle and packaging validation: `HERMES_HOME=<temp> pytest tests/test_plugin_lifecycle_cleanup.py tests/test_packaging_install.py -q` -> 40 passed
- [x] Compatibility validation with #501 applied: lifecycle cleanup + shared-storage class + packaging -> 53 passed
- [ ] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py tests/test_plugin_lifecycle_cleanup.py -q` -> 1100 passed
  - [ ] `pytest -q`
  - [ ] `scripts/validate_release.sh --full --keep-going --output /tmp/hermes-lcm-release-validation-finalize-clones-v2`
  - [x] `git diff --check origin/main...HEAD && git diff --check && git diff --cached --check`
- [x] Ruff 0.15.13 on changed Python files
- [x] Python compilation on changed Python files

## Notes

- Scope is plugin-owned LCM clone cleanup only; it does not claim to close Hermes-owned `state.db` connections.
- The branch is intentionally opened as a draft while broader validation and review complete.

Refs #463
Replaces #470

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -88,6 +88,33 @@ def _ensure_engine_bound_to_session(
         )
 
 
+def _make_session_finalize_handler(prototype_engine, resolve_active_lcm_engine):
+    """Close the exact per-agent LCM clone retired by a host session boundary."""
+
+    def _on_session_finalize(**payload) -> None:
+        session_id = str(payload.get("session_id") or "")
+        platform = str(payload.get("platform") or "").strip().lower()
+        if not session_id or platform == "cli":
+            # Interactive CLI /new finalizes the old session while reusing the
+            # same agent and context engine. Process exit closes its resources.
+            return
+
+        active_engine = resolve_active_lcm_engine(session_id=session_id)
+        if (
+            active_engine is None
+            or active_engine is prototype_engine
+            or _engine_bound_session_id(active_engine) != session_id
+        ):
+            return
+
+        try:
+            active_engine.shutdown()
+        except Exception as exc:
+            logger.debug("LCM finalized-clone shutdown failed: %s", exc)
+
+    return _on_session_finalize
+
+
 def _hook_question_date(payload: dict) -> object:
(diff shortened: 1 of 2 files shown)


bennybuoy:
Closing after independent lifecycle review found a release blocker: Hermes currently invokes `on_session_finalize` before `AIAgent.shutdown_memory_provider()` calls the context engine’s required `on_session_end()` flush. This draft therefore closes SQLite helpers too early, can drop the final assistant tail, and can leave lifecycle finalization incomplete. The session-ID registry is also last-bound rather than an exact retired-agent identity, so the hook cannot safely choose a clone when bindings overlap. This needs a different host/plugin handshake rather than patching around the ordering.

PR #533, opened 2026-08-17
fix: keep overflow recovery cap independent of host estimate

### LINKED ISSUE #530: forced-overflow recovery can return an empty transcript for no-system gateway sessions
## Summary

Forced-overflow recovery can return an empty transcript for a gateway conversation with no system message when the active suffix is assistant/tool-heavy. Hermes core correctly refuses the empty result, leaving the same over-limit session unable to recover.

## Observed behavior

A production gateway session repeatedly logged this shape:

```text
LCM compaction #1: 473 messages → 0 (... forced overflow recovery)
context compression returned an empty transcript; refusing to rotate session so the parent remains resumable
Context length exceeded: 454,306+ tokens. Cannot compress further.
```

`fallback_used=false`; the SQLite database passed its integrity check. The failure repeated against the same session because the host would not commit an empty compressed transcript.

## Root cause

For a no-system gateway session:

1. `_assemble_context()` can spend the forced-overflow tail budget on derived assistant/tool traffic.
2. A previously generated `[Current user objective preserved from compacted history]` scaffold may sit behind an oversized gap and not enter the selected contiguous tail.
3. Provider-validity cleanup removes leading assistant/tool messages, allowing assembly to collapse to `[]`.
4. `_assemble_overflow_recovery_context()` falls back to only `tail_messages[-1]`; if that is an orphan tool result, `_sanitize_active_context_messages()` drops it and still returns `[]`.

Relevant code is currently in `engine.py` around `_assemble_context()` and `_assemble_overflow_recovery_context()`.

## Minimal reproduction shape

```python
result = engine._assemble_overflow_recovery_context(
    None,
    [
        {
            "role": "assistant",
            "content": (
                "[Current user objective preserved from compacted history]\n"
                "KEEP_OBJECTIVE: continue the active plan."
            ),
        },
        {"role": "assistant", "content": "oversized assistant/tool chatter " * 400},
        {"role": "tool", "tool_call_id": "orphan-call", "content": "latest tool status"},
    ],
    assembly_cap_override=120,
)
assert result  # fails before the proposed guard: result == []
```

## Expected behavior

Whenever forced-overflow recovery receives a non-empty active transcript, it should return at least one provider-valid message. For this shape it should preserve the latest real or already-scaffolded user objective as a bounded `user`-role recovery anchor, rather than returning an orphan tool result or an empty list.

## Proposed narrow fix

At the empty/minimum-candidate fallback in `_assemble_overflow_recovery_context()`:

- Prefer an existing preserved-objective scaffold.
- Otherwise build the existing preserved-objective scaffold from the newest eligible real user message.
- Bound it to the assembly cap and emit it with role `user`.
- If sanitizing the existing last-message fallback still yields nothing, return a small provider-valid recovery instruction rather than `[]`.

No schema, database, configuration, or Hermes-core changes are required.

## Validation

A focused red/green regression test reproduces the no-system + preserved objective + oversized assistant traffic + orphan tool suffix shape:

- Before guard: fails (`[]`).
- After guard: returns a bounded user anchor containing the preserved objective.
- Full plugin suite after the local patch: `2300 passed, 12 xfailed`.
- Ruff, `compileall`, and `git diff --check` pass.

the3asic:
## Problem

Forced-overflow recovery derived its assembly cap as `assembly_cap - (observed_tokens - count_messages_tokens(messages))`. `observed_tokens` is the host rough full-request estimate; it may include system prompts, tool schemas, and provider replay metadata that LCM does not count. Treating that estimator difference as fixed overhead could collapse a valid cap to one token, after which sanitization returned an empty transcript and the host retried the same poisoned session.

## Fix

Use the configured LCM message-assembly cap unchanged during overflow recovery. The host estimate remains a pressure signal for the existing force-overflow gate; it no longer rewrites a cap whose accounting basis it does not share. Non-message headroom belongs in `reserve_tokens_floor` or an explicit cap.

The regression covers the production-shaped mismatch (`340,733` host estimate vs. `<10,000` LCM message estimate with Codex replay fields) and verifies a no-system recovery still returns a user message.

Fixes #530

## Verification

- `PYTHONPATH=$HOME/Dev/hermes-agent ~/the3asic-agentkit/.venv/bin/pytest -q tests/test_lcm_engine.py --disable-warnings`
- `752 passed, 1 skipped`
- `python -m compileall -q engine.py tests/test_lcm_engine.py`
- `git diff --check`

### DIFF
--- a/engine.py
+++ b/engine.py
@@ -6139,15 +6139,19 @@ class LCMEngine(CompactionMixin, ResetStateMixin, ReconcileMixin, AuxiliarySessi
         observed_tokens: Optional[int] = None,
         messages: Optional[List[Dict[str, Any]]] = None,
     ) -> Optional[int]:
-        assembly_cap = self._effective_assembly_token_cap()
-        if assembly_cap is None:
-            return None
-        if messages is None or observed_tokens is None or observed_tokens <= 0:
-            return assembly_cap
+        """Return the cap used while rebuilding the provider-visible context.
 
-        message_tokens = count_messages_tokens(messages)
-        overhead_tokens = max(0, observed_tokens - message_tokens)
-        return max(1, assembly_cap - overhead_tokens)
+        ``observed_tokens`` is a pressure signal from the host.  It is not a
+        message-only count: Hermes may include system prompts, tool schemas,
+        and provider replay metadata that LCM's message estimator deliberately
+        does not model.  Subtracting ``observed_tokens - message_tokens``
+        therefore treats estimator skew as fixed assembly overhead and can
+        collapse a normal cap to one token.  Headroom for non-message payloads
+        belongs in ``reserve_tokens_floor`` (or an explicit cap), so recovery
+        must use the configured message-assembly cap unchanged.
+        """
+        assembly_cap = self._effective_assembly_token_cap()
+        return assembly_cap
 
     def _effective_assembly_token_cap(self) -> Optional[int]:
         """Return the active assembly cap, if any.
--- a/engine.py
+++ b/engine.py
@@ -5917,6 +5917,22 @@ class LCMEngine(CompactionMixin, ResetStateMixin, ReconcileMixin, AuxiliarySessi
         """
         result = []
 
+        # An assembly_cap_override is the forced-overflow-recovery signature:
+        # every caller that passes one is rebuilding the context under
+        # recovery pressure. Project the inputs onto canonical message keys
+        # here so the recovery guarantee holds no matter which entry point
+        # (_assemble_overflow_recovery_context, the post-leaf-compaction
+        # assembly in compaction.py) reached this method — replay metadata
+        # that count_message_tokens does not model must not survive into a
+        # payload assembled against the cap. Normal (override-less) assembly
+        # never strips: replay fields are the host's business there.
+        if assembly_cap_override is not None:
+            system_msg = _strip_replay_metadata_for_recovery_message(system_msg)
+            tail_messages = [
+                _strip_replay_metadata_for_recovery_message(msg)
+                for msg in tail_messages
+            ]
+
         # Leading anchor with optional LCM annotation. Only a true system prompt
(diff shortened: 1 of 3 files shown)


the3asic:
Addressed in 955df29. Recovery now projects the system anchor and tail onto canonical message keys only (`role`, `name`, `content`, `tool_calls`, `tool_call_id`) before assembly, on copies — the caller's dicts are never mutated, and messages that already carry only canonical keys pass through untouched so the normal path is byte-identical. The required user turn and canonical tool calls survive; `reasoning` / `codex_*` replay metadata no longer rides along, so the recovered payload's provider-visible size matches what the cap was enforced against (the 340,733-observed / 220,000-cap shape from the review cannot recur). Two regressions pin it: the recovered payload retains only canonical keys (with tool-call preservation and no-mutation assertions), and the system anchor gets the same projection. Suite: 2794 passed / 2 pre-existing macOS path-containment failures that reproduce on the parent commit untouched. @codex review — happy to have another pass.

the3asic:
Good catch — fixed in 8f91040. The canonical projection now lives inside _assemble_context itself, gated on assembly_cap_override (the forced-recovery signature), so the Step-7 post-leaf-compaction path in compaction.py is covered along with every other recovery entry point; normal override-less assembly still never strips (replay fields remain the host's business there). Idempotent with the _assemble_overflow_recovery_context strip. New regression pins the exact scenario: forced overflow + leaf compaction + replay fields on the anchor and retained fresh tail → result carries canonical keys only, caller dicts unmutated. @codex review

the3asic:
@codex review — 8f91040 addresses the post-leaf-compaction bypass; the canonical projection is now centralized in _assemble_context behind the recovery override.

the3asic:
@stephenschoettler gentle ping — both Codex-review findings are addressed (955df29: strip retained replay metadata; 8f91040: cover the post-leaf-compaction assembly path), the latest bot pass found no remaining issues, and all 6 checks are green. Ready for a human look whenever you have a minute. Fixes #530.

PR #535, opened 2026-08-19
docs: fix broken star history chart

OctoBored:
The star history chart in the README is broken. Due to GitHub stargazer API restrictions the current chart no longer loads, so I switched it to an alternative data source that requires no API token and serves the chart reliably. Updated the chart image and its link in README.md. This is a small documentation fix that restores a useful feature for visitors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -856,10 +856,10 @@ for changelogs.
 
 ## Star history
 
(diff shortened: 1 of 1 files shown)


stephenschoettler:
Thanks for flagging the broken chart. I’m not going to move the README image and click target to an unaffiliated third-party domain, so I’m closing this PR. I’ll handle the chart separately with a maintainer-controlled source.

Issue #536, opened 2026-08-19
feat: expose existing message-pin API: lcm_pin/lcm_unpin tools, /lcm pin, compaction exclusion


ruangraung:
### Problem

`hermes-lcm` already contains a message-pin storage API, but nothing in production reaches it. The `pinned` column exists in the `messages` schema, `store.pin()` and `store.unpin()` are implemented, and the garbage-collection path already skips pinned tool rows, yet there is **no agent tool, no `/lcm pin` slash command, and compaction never filters on the flag**, so a pinned row folds into a summary node exactly like any other message.

The storage layer that exists today:

- [`store.py`](https://github.com/stephenschoettler/hermes-lcm/blob/10cbb78347ec86f3004153b24767324ded9e37b4/store.py) declares `pinned INTEGER DEFAULT 0` in the `messages` schema and implements `store.pin(store_id)` / `store.unpin(store_id)`.
- The same module skips the GC placeholder-rewrite for pinned tool rows, so a pinned row is not rewritten by collection.
- `_row_to_dict` already surfaces `pinned` to callers.

The gap is confined to the wiring:

- `compaction.py` candidate selection (`_leaf_compaction_candidate_status` and the live `_compress_impl` summarizer-input path) never checks `pinned`, so pinned messages still become summary candidates.
- No `LcmPin`/`LcmUnpin` tool schemas, no `lcm_pin`/`lcm_unpin` handlers, and no  `/lcm pin` command exist.

### User or operator impact

Without an exposed pin API, there is no structural way to keep a specific message verbatim. Identity/relationship facts and other content a user wants preserved are folded into summaries as sessions grow; the summarizer is asked to fold them, and once summarized the raw row is superseded by the summary node. Pinning is intended to make that survival **structural** rather than a summarizer guarantee: a pinned message is excluded from folding, stays raw in `lcm.db`, and remains recoverable verbatim via `lcm_expand(store_id=...)`.

Today a user cannot obtain this behavior at all: the flag is unreachable from the agent, from the slash-command surface, and from compaction selection.

### Proposed direction

Stage a scoped, non-rewrite change that exposes the existing API:

1. Add `lcm_pin(store_id)` and `lcm_unpin(store_id)` agent tools with schemas, dispatched through `LCMEngine.handle_tool_call` and registered like the other  `lcm_*` diagnostic tools.
[...]

PR #537, opened 2026-08-19
feat: expose message pinning: lcm_pin/lcm_unpin tools, /lcm pin command, compaction exclusion

ruangraung:
## Summary

Exposes the already-built but otherwise unreachable message-pin API in hermes-lcm. The `pinned` column, `store.pin()`/`store.unpin()`, and the GC-skip for pinned tool rows already exist; what was missing is the wiring agent tools, slash commands, and the compaction filter that honors the flag.

This change:

- Adds `LcmPin`/`LcmUnpin` schemas to `schemas.py` and `lcm_pin`/`lcm_unpin` handlers to `tools.py`, routed through `engine.handle_tool_call()` exactly like the existing `lcm_*` diagnostic tools (the engine's message-ingest / current-turn-search step is preserved).
- Adds `/lcm pin <store_id>`, `/lcm unpin <store_id>`, and `/lcm pins` slash commands in `command.py`, backed by a new `store.list_pinned()` query.
- Excludes pinned rows from compaction on **both** the pre-flight candidate check (`_leaf_compaction_candidate_status`) and the live summarizer-input path (`_compress_impl`), so pinned content never reaches the summarizer and stays raw/recoverable via `lcm_expand(store_id=...)`.
- Splits largely shared store-id parsing into a `_parse_store_id_arg()` helper  in `tools.py`.
- Registers the tools on every surface that enforces tool-contract sync:  `plugin.yaml` `provides_tools`, `get_tool_schemas()`, capability/registration  fixtures (`EXPECTED_LCM_TOOLS`), and the engine's `handle_tool_call` handler dict.
- Updates `README.md`, `docs/retrieval-tools.md`, `docs/operator-guide.md`, and  `docs/packaging.md` (tool counts 15 → 17) and adds a "Message pinning" section to the operator guide.

## Why

Without this, a pinned row means nothing at runtime: compaction never checks the flag, so a pinned message is summarized like any other message, and there is no way for the agent or an operator to set or clear the pin. Pinning makes
survival structural, excluded from folding, never GC-rewritten, recoverable verbatim.

## Validation

- [x] **Focused pin tests:** `pytest -k pin` → `22 passed`
- [x] Pin survives compression: pinned non-tool message stays raw, content  intact, and **absent from the summarizer input**  (`test_compress_excludes_pinned_non_tool_messages_from_summary`)
- [x] Tool handlers: `lcm_pin`/`lcm_unpin` on valid + missing + unknown store_id (`test_lcm_engine.py`)
- [x] Slash commands: `/lcm pin`, `/lcm unpin`, `/lcm pins`, plus unknown store_id reporting (`test_lcm_command.py`)
- [x] **Full suite:** `pytest -q` → `2803 passed, 1 skipped, 12 xfailed`
- [x] Tool-contract sync tests pass (`test_tool_contracts.py`, `test_packaging_install.py`,  `test_host_capability.py`)
- [x] `ruff check`: clean
- [x] `python -m compileall`: no errors
- [x] `git diff --check`: no whitespace errors

**Test environment:**

- hermes-lcm base: `10cbb783` (synced with upstream/main) `plugin_version: 0.21.0-rc1`
- Python: 3.13.5
- Pytest: 9.1.1
- OS: Debian 13 (Trixie), Linux 6.12
```python
hermes version
Hermes Agent v0.20.4 (2026.8.18) · upstream 13ce0c5c
Install method: git
Python: 3.11.15
OpenAI SDK: 2.24.0
```

## Notes

- No schema migration: the `pinned` column already exists.
- No new config flags; the optional per-turn injection knob (`LCM_PINNED_INJECT_TOKENS`) is deliberately out of scope and tracked as a follow-up.
- Backwards-compatible: default behavior unchanged; pinning applies at the next compaction selection and is not retroactive.
- Pinning does not bypass `LCM_SENSITIVE_PATTERNS` redaction (redaction happens  at ingest, before storage).

Refs #536


stephenschoettler:
Thanks for the detailed implementation. I’m not going to adopt this pinning direction. hermes-lcm already keeps raw messages recoverable after compaction, while this implementation removes a pinned turn from active context without summarizing it, still records that row as summary source lineage, and exposes unscoped durable mutation and listing across sessions. Those retention, scope, and active-context semantics need an architecture contract before code, so I’m closing this PR and the linked issue rather than asking for incremental fixes on the current branch.

ruangraung:
Hi @stephenschoettler, Thank you for reading the implementation closely and for the detailed, thoughtful explanation, even though this one isn't being merged, that kind of feedback is genuinely valuable, and I really appreciate you taking the time to lay it out so clearly.

You make a fair point that `hermes-lcm` already keeps raw messages recoverable after compaction, and that pinning was reachable through the existing expansion tools. I should have started with that conversation rather than bringing a full implementation straight to the branch. And you're right on the semantics: a pinned turn dropping out of active context without being summarized creates a real hole in the conversation DAG, and the scoping of durable mutations across sessions deserved more thought than I gave it up front.

I'll take "architecture contract before code" to heart for future contributions, the right move is to sketch the design in an issue and talk it through before writing anything.

Thanks again for everything, and for #209 and #217 that did land. I look forward to contributing more in the future.

Best regards.

Issue #373, opened 2026-07-14
[Bug]: GPT-5.6 Codex OAuth live catalog rolled back to 272K; 372K fallback is stale


100yenadmin:
## Summary

PR #369 changed the GPT-5.6 Codex OAuth cap to 372,000 based on the static `openai/codex` repository catalog. The current authenticated Codex client catalog has since rolled Sol, Terra, and Luna back to:

- `context_window: 272000`
- `effective_context_window_percent: 95`

The public repository copy still says 372,000, so it is currently stale relative to the live catalog delivered to clients.

## Impact

Hermes core currently supplies 272K, so LCM remains safe when that host metadata wins. However, `_CODEX_OAUTH_CONTEXT_CAPS["gpt-5.6"] = 372_000` is now an unsafe fallback whenever host/live metadata is absent, stale, or larger than the actual Codex OAuth limit.

This can cause LCM to defer compression beyond the live route's supported window.

## Reproduction

1. Refresh the authenticated Codex model catalog/cache.
2. Inspect Sol, Terra, and Luna entries: each currently reports 272,000 with a 95% effective window.
3. Call `_codex_oauth_context_cap("gpt-5.6-sol", "openai-codex")` in current Hermes-LCM main.
4. Observe that the helper still returns 372,000.

## Suggested fix

- Restore the conservative GPT-5.6 Codex OAuth fallback to 272,000 until the live catalog re-enables a larger window.
- Prefer live/host metadata when available, but never let a static fallback raise the effective cap above a lower live value.
- Add regression coverage for a live 272K value overriding a stale 372K fallback.

Related:

- #369
- NousResearch/hermes-agent#61910 (open core-side 372K update, also based on the stale static catalog)

cresslank:
Opened PR #540: https://github.com/stephenschoettler/hermes-lcm/pull/540. It makes Hermes provider-resolved Codex metadata authoritative when available, including lower live values, while intentionally retaining the existing 372K compatibility fallback. The PR does not claim the account-independent fallback policy is settled; that tradeoff is documented for review.

PR #540, opened 2026-08-21
fix(codex): use Hermes provider context resolution

cresslank:
## Summary

- Resolve Codex context windows through Hermes Agent's supported `get_model_context_length()` API, preserving the literal model ID and forwarding the normalized provider and active credential.
- Honor the host resolver's policy, including explicit `model_overrides` before provider metadata. Retain `min(raw_host_context, resolved_window)`.
- Handle staged credentials, credential-only rotation, explicit clearing versus omission, and stale session metadata after `update_model()`.
- Keep all local fallback values, including #557's exact Terra/Sol/Luna `-900k` aliases. Missing/incompatible/failed resolvers and invalid results fall back without logging exception details.

## Why

LCM's static table can disagree with host resolution. This delegates that decision to the supported API without introducing a separate provider-overrides-config policy. A successful result can reflect configuration, cached metadata, host variant policy or fallback, not necessarily a fresh authenticated provider limit.

One commit on LCM main `8d1b1e6d3d63f5fc7b209e8d7ec1dc9b814f2e54`, preserving original authorship/date. Published head: `77739bd40aeefb5c4dd1fed3176c143c44943261`.

## Validation

Current-host checks used Hermes `e10934b03e115b036b3c38b9c6e6817039a5962d`, Python 3.11.13, isolated configuration, no inherited credentials/private plugins, and serial tests.

- **145 focused tests passed**, including ten opt-in real-host subprocess cases with synthetic HTTP and import-origin assertions. Selection: `tests/test_codex_routing.py tests/test_codex_host_resolution.py tests/test_dependency_contract.py`, with `HERMES_LCM_TEST_HOST` pointing to the pinned host. Normal minimal-host CI skips the ten real-host cases.
- **1,109 core/engine/packaging tests passed, 1 skipped**: `tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py`.
- Additional isolated host-lifecycle checks passed for staged credentials, token-cache partitioning/rotation, omission, clearing, stale-session rejection, catalog maximum handling and the lower raw-context bound.
- Whole-repository Ruff, dependency-contract validation and whitespace checks passed. Local Ruff was 0.15.10; CI pins 0.15.13.
- **Full release validation is not green.** The prior run of this identical candidate/base returned 1: normal pytest **3,078 passed / 1 failed / 2 skipped / 12 xfailed**; low-FD pytest **3,073 passed / 6 failed / 2 skipped / 12 xfailed**. All six candidate failing tests also failed on the exact base in wall-clock/thread timing tests. Other release gates passed. Those broad runs used an earlier pinned Hermes host and were not repeated for this source-identical qualification.

Tests were run with `python -m pytest -q -p no:cacheprovider` and an isolated environment. No live authenticated account/entitlement test was performed.

## Policy and scope

Explicit host model overrides may exceed authenticated catalog values. This follows the supported host policy, not an independent hard provider maximum. Raw context is not forwarded as `config_context_length`; `base_url` is omitted to retain provider-scoped lookup and older-host compatibility. Current Hermes already exempts native Codex routes from generic custom-endpoint probing.

#373's fallback-policy question and #459's default-versus-maximum/top-level override semantics remain unresolved. #573 overlaps variant handling but not the full resolver/credential lifecycle. No host, configuration schema, database schema, dependency installer or workflow changes.

Refs #373, #459, #573.

## Hosted CI

Hosted workflows for the published head ([CI](https://github.com/stephenschoettler/hermes-lcm/actions/runs/35520052243)) report `action_required`, awaiting maintainer approval. The CI run has zero jobs; this is neither passing CI nor a test failure. Please approve the workflows when ready to review.

### DIFF
--- a/codex_routing.py
+++ b/codex_routing.py
@@ -9,13 +9,17 @@ policy constants (for example the gpt-5.5 compaction threshold).
 
 from __future__ import annotations
 
+import logging
+
+logger = logging.getLogger(__name__)
+
 # ChatGPT Codex OAuth exposes provider-enforced context windows that can be
 # materially lower than the same model slug on direct OpenAI/OpenRouter routes.
 # Hermes Agent resolves these from chatgpt.com/backend-api/codex/models, with
-# this table as its fallback. LCM sees only the host-advertised context_length;
-# when that value was explicitly overridden above the real Codex OAuth window,
-# we still have to budget against the effective provider window or compaction
-# fires too late and provider requests can overflow.
+# this table as a compatibility fallback. LCM receives a host-provided
+# context_length, but that value can be an explicit override or stale metadata;
+# resolve the active provider route here as well so compaction follows the same
+# authenticated window Hermes uses for the request.
 _CODEX_OAUTH_CONTEXT_CAPS: dict[str, int] = {
     "gpt-5.1-codex-max": 272_000,
     "gpt-5.1-codex-mini": 272_000,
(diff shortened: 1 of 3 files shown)


cresslank:
@stephenschoettler — current status update

I verified this PR against the live `hermes-lcm/main` at `10cbb78347ec`. It was already based on that exact commit, remains a single commit at `7640d2f390a0`, and did not need a cosmetic rebase or head rewrite.

Issue #373 remains open and still describes the stale GPT-5.6 Codex OAuth fallback this change addresses. The current Hermes-Agent provider resolver accepts the `api_key` and `provider` arguments used here, so the integration remains aligned with the host implementation rather than hard-coding a new LCM policy.

Exact-head local validation:
- 36 Codex/context-cap/session-start/deepcopy tests passed
- `uv --locked` was not used because this checkout’s lockfile is not current and the project has no `dev` optional dependency; the direct `pytest` run was read-only and passed

The GitHub workflow suite is still `action_required` with no check jobs reported. That is the remaining maintainer admission gate, not a demonstrated implementation failure. Please approve the fork workflow when convenient so the hosted CI can run.

cresslank:
Current-main follow-up: #557's exact 900K aliases have landed and must be preserved; #573 overlaps the host-predicate side of routing. Neither makes this branch's broader provider-aware context objective obsolete, but this head now conflicts with the new routing policy/tests. The August test results above are historical.

Before refreshing the implementation, can we confirm the intended precedence between explicit host model_overrides and authenticated provider metadata? Hermes's public get_model_context_length resolver honors explicit overrides first, so calling it does not by itself guarantee that lower provider metadata overrides an explicit higher value.

My proposed port would honor the host's documented resolution policy, retain #557's exact-alias fallback when host resolution is unavailable, preserve the existing lower-host-context cap, and test those boundaries against the actual host resolver rather than only mocking it. If provider metadata must always win even over explicit overrides, that needs an explicit host contract instead. I will leave workflow admission until the refreshed candidate and this policy are settled.

cresslank:
Published the refresh at `77739bd40aeefb5c4dd1fed3176c143c44943261` on LCM main `8d1b1e6d3d63f5fc7b209e8d7ec1dc9b814f2e54`, preserving original authorship and #557's exact aliases.

This implements the proposed supported-host-resolver policy, including explicit model-override precedence, the lower raw-context bound and unchanged local fallbacks. It does not claim authenticated metadata overrides explicit configuration or that #373/#459 are settled. Staged/rotated/cleared credentials and stale session metadata are covered.

Against Hermes `e10934b03e115b036b3c38b9c6e6817039a5962d`: **145 focused passed; 1,109 core/engine/packaging passed, 1 skipped**. Ruff, dependency contract and diff checks passed. Real-host tests used isolated configuration and synthetic HTTP, not account access.

Full release validation is **not green**: the prior exact-candidate run had one standard-suite and six low-FD timing failures, all reproduced on exact base. The updated body separates that historical broad run from current-host qualification; no unrelated timing cleanup was included.

Hosted workflows for the published head ([CI](https://github.com/stephenschoettler/hermes-lcm/actions/runs/35520052243)) report `action_required`, awaiting maintainer approval. The CI run has zero jobs; this is neither passing CI nor a test failure. Please approve the workflows when ready to review.

Issue #552, opened 2026-08-29
fix: lifecycle frontier never advances on ingest — frozen frontier hides raw backlog (deferred maintenance never fires)


emiltsoi:
### Expected behavior

The lifecycle frontier (`current_frontier_store_id`) should advance as messages are ingested, so the engine's raw-backlog/debt computation stays honest between compactions. `MAX(store_id) - current_frontier_store_id` should reflect the actual unconsumed backlog.

### Actual behavior

The lifecycle frontier is only advanced at session bind (`on_session_start`) and after a successful compaction (`_persist_frontier_marker`). It is NEVER advanced during normal per-turn ingest. When the compression preflight is gated off (or no compaction fires for a long time), the frontier stays frozen at the last compaction point while `MAX(store_id)` grows every turn. The engine then computes `raw backlog = MAX(store_id) - frontier = 0` → no debt is recorded → deferred maintenance never fires → the conversation runs hot to the context backstop with zero compactions.

Observed fleet-wide on long-lived conversations: one profile had `current_frontier_store_id = 0` with `MAX(store_id) = 283,913` (283K messages of hidden backlog); others had 18-28K lags. Context climbed to ~380-500K tokens with no maintenance events for days.

### Minimal reproduction

1. Enable hermes-lcm with `context.engine: lcm` + deferred maintenance (`LCM_DEFERRED_MAINTENANCE_ENABLED=true` or the v2.1 env regime).
2. Start a session, ingest messages, let one compaction advance the frontier to F (> 0).
3. Disable the compression preflight path (e.g. `agent.compression.enabled: false`) OR simply let the conversation continue without any further compaction firing.
4. Ingest more messages (they store fine — `MAX(store_id)` grows to M).
5. Observe: `current_frontier_store_id` stays at F, `raw backlog = M - F` computes as 0 backlog (because the refresh path only runs inside the gated preflight), `debt_kind = null`, no maintenance ever fires.

### Version and branch

main @ 10cbb78 (v0.21.0-rc2 lineage), observed on a long-lived fleet deployment

### Environment

Linux, Python 3.12, Hermes Agent gateway with hermes-lcm plugin, long-lived Telegram DM sessions (weeks), deferred-maintenance regime (200000/0.50/128/0/true)

### Logs, traceback, or failing test

The failure is silent (no error) — the symptom is a frozen frontier + absent debt/maintenance. A regression test that captures the class:
[...]

PR #553, opened 2026-08-29
fix: advance lifecycle frontier on ingest (frozen-frontier hides raw backlog)

emiltsoi:
## Summary

The lifecycle frontier (`current_frontier_store_id`) now advances with per-turn ingest, not only at session bind and after compaction. A new `advance_frontier_to_store_max()` advances the frontier to the conversation's `MAX(store_id)` after each successful ingest (monotonic SQL `MAX`, guarded on `current_session_id`).

## Why

The frontier was compaction-driven. When the compression preflight was gated off or no compaction fired for a long time, the frontier stayed frozen while `MAX(store_id)` grew every turn. The engine computed `raw backlog = MAX(store_id) - frontier = 0`, recorded no debt, and deferred maintenance never fired — long-lived conversations ran hot to the context backstop with zero compactions. Observed fleet-wide: one profile had frontier 0 with 283K messages of hidden backlog; others had 18-28K lags.

## Validation

- [x] Focused validation: `pytest "tests/test_lcm_engine.py::TestEngineABC::test_ingest_advances_frontier_to_store_max_without_compaction" tests/test_fresh_tail.py -q` -> `10 passed`
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> `1099 passed` (6:56)
  - [x] `pytest -q` -> 2805 passed (fleet-tested version; full-suite re-run on this branch takes ~11 min)
  - [x] `bash -lc 'ulimit -n 1024 && pytest -q'` -> 1099 passed (low-FD focused set, 7:13)
  - [x] `python -m compileall -q .` -> OK
  - [x] `python -m py_compile scripts/import_lossless_claw.py` -> OK
  - [x] `bash -n scripts/install.sh scripts/update.sh` -> OK
  - [x] `git diff --check` -> OK
- [ ] Workflow validation (no workflows changed): N/A

## Notes

- The regression test (`test_ingest_advances_frontier_to_store_max_without_compaction`) captures the class: ingest messages with no compaction → frontier must equal the store max.
- `test_rotate_uses_effective_token_bounded_tail` was updated: it asserted the OLD frontier-lag behavior (rotate preview `noop=False` because the frontier lagged). With the frontier correct, the rotate properly no-ops within the fresh tail — the messages are all current/consumed.
- The advance is one cheap SQL UPDATE per turn (guarded on the current session row), monotonic under concurrency via `MAX()` in SQL.

Refs #552

### DIFF
--- a/compaction.py
+++ b/compaction.py
@@ -504,6 +504,19 @@ class CompactionMixin:
         leaf_compacted_this_turn = False
         dropped_replayed_scaffold_messages = False
         leaf_passes = 0
+        # Total-summary-output bound for the WIDENED leaf path (agent0 2026-08-30
+        # edge): with a large store backlog split into many chunks, per-chunk
+        # summaries (12K cap each) can MERGE into a total LARGER than the
+        # original window -> the anti-growth guard refuses (double work + cache
+        # break). Cap the accumulated summary tokens so a widened drain stops
+        # once the summary prefix would exceed the budget; the rest of the
+        # backlog waits for future compactions.
+        widened_summary_budget = (
+            int(self._config.summary_prefix_target_tokens)
+            if self._config.summary_prefix_target_tokens > 0
+            else int(self._config.leaf_chunk_tokens)
+        )
+        widened_summary_used = 0
         estimated_active_tokens = (
             observed_prompt_tokens
             if observed_prompt_tokens is not None and observed_prompt_tokens > 0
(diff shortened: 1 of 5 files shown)


emiltsoi:
Closing this PR — the underlying design question (how the summarizer should handle a widened working set over its chunk budget) is tracked in #554. This branch remains in our fork for our deployment; happy to share details there if maintainers want them.

Issue #554, opened 2026-08-29
compaction: widened store backlog is summarized as one giant LLM call (400) with default config — design question: chunk the widened set?


emiltsoi:
## Summary

When the compression working set is widened from the session store (a large unsummarized backlog), the leaf summarization path sends the ENTIRE widened set to the LLM in a single call when the engine is running with the default config (no full-sweep, no dynamic leaf chunking). For a multi-million-token backlog this exceeds the model's context window and the summarization call fails with HTTP 400 (`requested N tokens > max context`), aborting compaction entirely.

## Expected behavior

The widened working set should be summarized in bounded chunks (e.g. by `leaf_chunk_tokens`), not as one giant call, so a large backlog can be drained incrementally. The chunking machinery already exists for the dynamic/full-sweep paths (`_select_oldest_leaf_chunk`); the default path bypasses it.

## Actual behavior

```
LLM summarization failed: Error code: 400 — This model's maximum context length is 1048576 tokens.
However, you requested 2952216 tokens (2952216 in the messages, 0 in the completion)
```

Compaction aborts, no reduction happens, and the context stays at its pre-compaction size.

## Repro environment

- hermes-lcm with the store-backlog widen (compress() pulls messages from the store when the passed window is materially smaller than the session)
- Default config: `dynamic_leaf_chunk_enabled=false`, `threshold_full_sweep_enabled=false`
- A session store holding a large unsummarized backlog (multi-megabyte class)

## Design question (the core of this issue)

How SHOULD the summarizer handle a widened working set that exceeds the leaf chunk budget? Options discussed:

1. **Chunk always** (route the default path through `_select_oldest_leaf_chunk` bounded by `leaf_chunk_tokens`, multiple leaf passes) — drains the backlog incrementally, but changes the default path's "whole candidate in one pass" contract (preservation behavior differs).
2. **Chunk only when widened** (gate on a widen signal, e.g. the working set differs from the passed window) — minimal blast radius, normal windows keep exact behavior.
3. **Cap the widen** (pull only up to N messages / M tokens from the store) — bounded memory, but the backlog drains slowly and the leaf pass may no-op repeatedly.

This issue is a defect report + a request for the maintainers' design call. No patch is attached — we're running a local variant that works around this for our deployment, and we're deliberately not attaching it so the design conversation stays open. Happy to share details if useful.

## Version

hermes-lcm at/after the store-widen change (post-v0.21.0-rc2 line).

PR #561, opened 2026-09-01
fix(escalation): acceptance gate must enforce the token budget, not just < source_tokens

emiltsoi:
## Summary

- Tighten `summarize_with_escalation`'s acceptance gate: the accepted summary must be **≤ the token budget**, not merely `< source_tokens`.
- L1 gate: `count_tokens(result) <= token_budget` (was `< source_tokens`)
- L2 gate: `count_tokens(result) <= l2_budget` (was `< source_tokens`)
- Adds a regression test: a source-inclusive LLM echo (100K tokens vs 12K budget) is rejected → escalates to L3 deterministic truncation (bounded).

## Why

With a large source chunk, `< source_tokens` is nearly vacuous — the model's max output is always smaller than a multi-hundred-K source, so any model output is accepted regardless of the budget the prompt asked for. The stored summary can be ~an order of magnitude larger than intended (observed at scale: 2.2M-token "summaries" that grew a compaction 10x and stalled it via the anti-growth guard). The budget is already what the L1/L2 prompts communicate ("Target ~{token_budget} tokens"), so this makes the acceptance gate consistent with the prompt contract.

## Validation

- [x] Focused validation: `pytest tests/test_lcm_engine.py -q` -> 1102 passed (focused set: test_lcm_core + test_lcm_engine + test_packaging_install)
- [x] Default validation:
  - [x] `pytest tests/test_lcm_core.py tests/test_lcm_engine.py tests/test_packaging_install.py -q` -> 1102 passed
  - [ ] `pytest -q` (full suite; see Notes)
  - [ ] `scripts/validate_release.sh --full --keep-going --output /tmp/hermes-lcm-release-validation-<topic>` (see Notes)
  - [x] `git diff --check origin/main...HEAD && git diff --check && git diff --cached --check` -> clean
- [ ] Workflow validation: not applicable (no workflow changes)

## Notes

- Full `pytest -q` and `scripts/validate_release.sh --full` were not run in this PR's CI environment (time); the focused set above covers the touched surface (`escalation.py` + one new test in `test_lcm_engine.py`). Happy to run the full suite if the maintainer wants it before merge.
- The fix is deliberately minimal (2 predicates + 1 regression test). The broader widened-backlog compaction behavior is tracked separately in #554 (design question) — this PR only fixes the acceptance gate.

Refs #554

### DIFF
--- a/escalation.py
+++ b/escalation.py
@@ -574,7 +574,12 @@ def summarize_with_escalation(
         timeout=timeout,
         circuit_breaker=circuit_breaker,
         spend_guard=spend_guard,
-        accepts_result=lambda result: count_tokens(result) < source_tokens,
+        # ACCEPTANCE GATE (2026-09-01, Clara stall fix): the accepted summary
+        # must be ≤ the token BUDGET (12K cap), not merely < source_tokens.
+        # With a 2.2M-token widened chunk, '< source_tokens' accepted a
+        # source-inclusive ~2.2M echo → 1000→9,223 message blowup (10× tokens).
+        # Enforcing the budget rejects the echo → escalates to L2/L3 → bounded.
+        accepts_result=lambda result: count_tokens(result) <= token_budget,
     )
 
     if l1_result:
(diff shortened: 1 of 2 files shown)


emiltsoi:
Fixes the defect described in #562.

PR #567, opened 2026-09-01
fix(storage): remove clock uniqueness assumptions

### LINKED ISSUE #566: fix: remove clock-uniqueness assumptions from storage
## Expected behavior
Persisted identities and ordering should not assume the wall clock advances between adjacent operations:

- every message inserted by `append_batch()` should receive the distinct timestamp promised by its existing regression test;
- repeated externalization of the same content should not collide merely because `time.time_ns()` repeats.

## Actual behavior
On current `main`, the existing 50-row timestamp test fails on Windows with only 2 distinct timestamps. Freezing `time.time_ns()` also makes the second same-content externalization reuse the first filename; the exclusive create fails and externalization falls back inline.

## Minimal reproduction
At `22f6089b6e0f7ef43a787ae91713b6875ee21bb5`:

```text
python -m pytest -q -o 'addopts=' \
  tests/test_lcm_core.py::TestMessageStore::test_append_batch_timestamps_are_unique_per_row

Expected 50 unique timestamps, got 2
```

A deterministic companion regression freezes `time.time()` and `time.time_ns()`; both timestamp uniqueness and filename uniqueness fail on unpatched `main`.

## Proposed scope
- use a compact random suffix rather than `time_ns()` for externalized and temporary payload filenames;
- advance duplicate/non-increasing per-batch timestamps with `math.nextafter()`;
- make both regressions deterministic across platforms.

## Environment
Windows 11, Python 3.11. The frozen-clock reproductions are platform-independent.

## Validation attempted
- Unpatched current-main deterministic regressions: **2 failed**.
- Candidate implementation: **2 passed**; Ruff, compilation, and diff checks passed.

supplefrog:
## Summary
- replace `time.time_ns()` filename suffixes with compact UUID-derived suffixes for externalized and temporary payload files
- guarantee strictly increasing message timestamps within each `append_batch()` call even when the clock repeats or moves backward
- make the existing timestamp regression deterministic and add a deterministic filename-collision regression

## Why
`time.time()` and `time.time_ns()` report clock values, not uniqueness guarantees. Current `main` already has a test requiring distinct per-row batch timestamps, but on this Windows host 50 rows received only 2 distinct values. When `time_ns()` repeats, same-content externalizations select the same path; the exclusive create fails and the second payload falls back inline.

This fixes the whole adjacent clock-identity class without changing persisted schema or filename parsing.

## Validation
- [x] Unpatched current-main deterministic regressions: the two focused tests -> **2 failed**
- [x] Candidate focused validation: the two focused tests -> **2 passed**
- [x] Ruff on changed Python files -> **passed**
- [x] `python -m compileall -q .` -> **passed**
- [x] `git diff --check` / staged diff check -> **passed**
- [ ] Default Linux validation is delegated to existing CI on this independent branch.

## Notes
- Timestamp monotonicity is guaranteed within each batch, matching the existing contract and test; this does not introduce a global timestamp sequence across processes or restarts.
- The 16-hex-character suffix keeps existing placeholder sizes stable while removing clock resolution as an identity source.
- A broader local Windows run recorded **361 passed, 83 failed, 3 skipped** because this branch intentionally does not include the independent parent-directory `fsync` fix in #563; those failures are not reported as green.

Fixes #566

### DIFF
--- a/externalize.py
+++ b/externalize.py
@@ -17,6 +17,7 @@ import os
 import re
 import stat
 import time
+import uuid
 from pathlib import Path
 from typing import Any, BinaryIO, Dict
 
@@ -48,6 +49,11 @@ def _safe_stub(value: str, fallback: str) -> str:
 
 def _content_digest_prefix(content: str) -> str:
     return hashlib.sha256((content or "").encode("utf-8")).hexdigest()[:12]
+
+
+def _unique_file_suffix() -> str:
+    # Keep references compact while avoiding any clock-resolution assumption.
+    return uuid.uuid4().hex[:16]
 
 
 def _preview_sha256(preview_prefix: Any) -> str:
@@ -167,7 +173,7 @@ def _write_externalized_payload(path: Path, payload: Dict[str, Any]) -> None:
 
 
 def _replace_externalized_payload(path: Path, payload: Dict[str, Any]) -> None:
-    tmp_path = path.with_name(f"{path.name}.{time.time_ns():x}.tmp")
+    tmp_path = path.with_name(f"{path.name}.{_unique_file_suffix()}.tmp")
     try:
         _write_externalized_payload(tmp_path, payload)
         tmp_path.replace(path)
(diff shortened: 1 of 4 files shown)


supplefrog:
@codex review

PR #575, opened 2026-09-04
feat(amr): Level-1 quote anchors + four-outcome citation verifier

Rutimka:
## Summary

Implements the AMR (auditable-memory-records 0.1) **Level-1 slice** for the V4 assertion family, per the public spec: https://github.com/phasespace-labs/auditable-memory-records (SPEC.md §4.2/§5/§6).

- `lcm_assertions` gains `source_quote_hash` (algorithm-prefixed AMR §5 sha256; column CHECK enforces the exact format **including lowercase hex** via `substr` prefix check + negated GLOB on the suffix — note: `GLOB 'sha256:[0-9a-f]*'` alone would NOT enforce case, since SQLite GLOB char classes match a single char followed by any chars) and `epistemic` (closed AMR §4.2 vocabulary: `fact`/`inference`/`open_question`/`unverified`; NULL is the unmarked state — *absent is not fact*, values are case-sensitive, unknown values are **rejected, never coerced**).
- `lcm_assertion_relations` gains `source_quote_hash`.
- **Hash binding is post-digest:** assertion/relation identity digests are unchanged, so existing ids remain stable across the migration and idempotent re-publication stays a zero-write no-op.
- Idempotent in-place migration rebuilds only the two rebuildable quote tables (insert guards temporarily down so already-invalidated generations survive the copy; the re-entrant call of `ensure_assertion_tables` inside the migration is documented in-code and terminates because `_needs_rebuild` is false on re-entry), backfills hashes via AMR §5 normalization (smart punctuation → ASCII, whitespace runs collapsed, trimmed), and preserves every `lcm_assertion_sources` row **including invalidation history**. The exact-shape schema verifier passes for fresh and migrated databases alike.
- New `assertion_amr.py`: read-only verifier reporting the four AMR §6 outcomes (`ok` / `anchor_tampered` / `source_drifted` / `source_missing`) with the §6.1 partial-check signal when no hash is stored. It never mutates state; the existing fail-closed query path is untouched.
- `lcm_query_state` exposes `epistemic` and per-anchor `quote_hash` in `source_ref`; absent values stay explicitly `null` so the unmarked state is visible to consumers.

## Test plan

- 14 new tests (`tests/test_assertion_amr.py`): vocabulary closure, normalization idempotence, anchor format (schema-level defense-in-depth incl. **case enforcement at correct length**), case sensitivity, all four verifier outcomes, migration preservation (invalidated generation + hash backfill), relation anchors, zero-write idempotency.
- Full suite in a clean worktree off `origin/main`: **3043 passed, 1 skipped, 12 xfailed** (numpy-needing int8 suite included via extra dep).
- `ruff check` clean on all touched files.

Reviewed pre-publication: constraint case-gap found in internal review, fixed (substr + negated GLOB), regression tests added.

### DIFF
--- a/assertion_amr.py
+++ b/assertion_amr.py
@@ -1,0 +1,313 @@
+"""AMR (auditable-memory-records 0.1) citation verification for assertions.
+
+Read-only companion to :mod:`assertion_store`: it verifies stored quote
+anchors against the authoritative message content and reports the four AMR
+§6 outcomes. It never mutates state and never resolves disagreement between
+records — verification answers only "does this citation still hold?".
+
+Outcome semantics (AMR §6 — collapsing these into a boolean is the classic
+mistake; ``anchor_tampered`` and ``source_drifted`` look identical to a
+naive "citation invalid" check and mean opposite things):
+
+- ``ok`` — stored hash matches the stored quote, and the quote is still at
+  its span in the unchanged source.
+- ``anchor_tampered`` — the record's own citation is internally
+  inconsistent (stored hash disagrees with the stored quote, or the
+  span/quote pair disagrees with a provably unchanged source).
+- ``source_drifted`` — the record is honest, but the source changed
+  underneath it.
+- ``source_missing`` — the cited source cannot be resolved.
+
+Level-1 marking is carried by the closed ``epistemic`` vocabulary; ``None``
+is the unmarked state and is never coerced to ``fact`` (absent is not
+fact). This module also exposes the per-implementation AMR declaration.
+"""
+
+from __future__ import annotations
+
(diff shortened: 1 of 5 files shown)


Rutimka:
Known non-blocking nits from pre-publication review, for transparency:

- `verify_assertion_citations` / `verify_relation_citations` share ~80% of their logic — happy to unify into a shared helper if preferred.
- `_quote_hash` / `_normalize_quote_for_hash` are underscore-prefixed helpers imported cross-module from `db_bootstrap.py` (no existing precedent in the repo). Can promote them to public helpers if that fits conventions better.

Both are easy follow-ups; say the word.

PR #576, opened 2026-09-04
feat(sufficiency): explicit sufficiency verdict and disclosure at the preanswer finalize seam

Rutimka:
## Summary

Implements the **is_sufficient gate** half of the 2026-09-03 scout synthesis decision #2 (docs/research/2026-09-03-scout-synthesis.md, Decision Question on evidence sufficiency): a single-place policy that classifies every finished preanswer-evidence result into an explicit sufficiency vocabulary and maps it to one answer-contract policy action.

- New `sufficiency_gate.py`: `answer_sufficient` / `computation_sufficient` / `finite_coverage` → **answer**; `partial` → **answer_with_disclosure** (rendered exclusively from fields already stored on the result — no provider, no fabricated content); `unknown` / `conflicted` → **annotate** instead of staying silent.
- States where the pipeline performed no evidence work (feature boundaries, unsupported questions, ordinary routes) receive **no claim at all**: unmarked stays unmarked — absent is not fact, unknown is not refuse.
- The requirements compiler's blanket `unknown` state is refined per reason code (`ambiguous_*` / `state_conflicted` → `conflicted`; grounded evidence with no delivered context → `partial`); **delivered states are never overridden**.
- The gate never calls a provider and never mutates evidence, computation, or the delivered context of sufficient results.
- Wired as `LCM_PREANSWER_EVIDENCE_MODE=sufficiency_v1` through the `pre_llm_call` hook (requirements-compiler route and legacy selective route both gate fail-open). Default-off: every other mode stays byte-identical to its previous behavior.
- Module API: `build_preanswer_evidence(sufficiency_gate=True)` for the direct path; the disclosure block is emitted as the ephemeral context only for non-sufficient states with empty context.

## Test plan

- 7 module-level tests (`tests/test_sufficiency_gate.py`): byte-identity without the flag, gate additivity on sufficient results, disclosure rendering from stored fields only, annotate-on-no-hit, no-claim paths stay unmarked, evidence/computation immutability.
- 4 hook-level tests (`tests/test_sufficiency_hook.py`) through the packaging harness: answer verdict on a validated fact, annotate on zero hits, conflicted annotation on two contradictory facts, byte-identical no-claim routing.
- Full suite in a clean checkout off `main`: **3040 passed, 1 skipped, 12 xfailed**.
- `ruff check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,12 @@ This repo also publishes GitHub Releases. This file is the repo-root release sur
   performed no evidence work (feature boundaries, unsupported questions,
   ordinary routes) receive no claim at all: unmarked stays unmarked.
   The gate never calls a provider and never mutates evidence, computation,
-  or the delivered context of sufficient results.  Default-off; every other
-  mode is byte-identical to its previous behavior.
+  or the delivered context of sufficient results.  Gate application is
+  atomic (a gate crash can never leave a partially marked result; the
+  wrapper fails open to the legacy result), delivered compiler states are
+  never overridden by reason-code classification, and the compiler's trace
+  digest remains re-derivable from every gated result.  Default-off; every
+  other mode is byte-identical to its previous behavior.
 
 ## v1.0.0-rc.1 - 2026-09-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,20 @@ This repo also publishes GitHub Releases. This file is the repo-root release sur
 
 ## Unreleased
 
-No additional changes yet.
+### Added
+
+- Sufficiency gate (`sufficiency_gate.py`, mode
+  `LCM_PREANSWER_EVIDENCE_MODE=sufficiency_v1`): the preanswer evidence
+  pipeline now delivers an explicit sufficiency verdict with every result —
+  `answer_sufficient` / `computation_sufficient` / `finite_coverage` map to
+  *answer*, `partial` to *answer with disclosure* (rendered exclusively from
+  fields already stored on the result), and `unknown` / `conflicted` to
+  *annotate* (disclose rather than stay silent).  States where the pipeline
+  performed no evidence work (feature boundaries, unsupported questions,
+  ordinary routes) receive no claim at all: unmarked stays unmarked.
+  The gate never calls a provider and never mutates evidence, computation,
+  or the delivered context of sufficient results.  Default-off; every other
+  mode is byte-identical to its previous behavior.
 
 ## v1.0.0-rc.1 - 2026-09-03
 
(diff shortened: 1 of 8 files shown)


Rutimka:
Independent pre-merge review round (two external read-only reviewers, different models, run on a frozen worktree copy before committing fixes) reported **SHIP-WITH-FIXES**. All confirmed findings are addressed in af2c32c:

1. **Delivered-state precedence** — a recognized delivered `state` now outranks the reason-code no-claim tables, so a stale/future no-claim code can never strip the verdict from a delivered result. The compiler's base `unknown` state keeps its documented exception (refusal paths carrying it with no-claim codes stay unmarked).
2. **Atomic gate application** — the verdict is assembled on a copy and committed only on success; a classification/render crash can no longer leave a partially marked result. `build_preanswer_evidence` now also fails open (matching the hook-path contract) and logs a warning.
3. **Trace digest integrity** — gate latency and disclosure size moved into the gate-owned `sufficiency` section; the gate no longer writes foreign `metrics` keys, so `trace.digest_sha256` remains re-derivable from every gated result (verified by a re-derivation regression test).
4. Removed the unreachable legacy-selective hook call (sufficiency_v1 routes through the requirements compiler only, as documented).

New regression tests cover all four; full suite 3043 passed / 1 skipped / 12 xfailed, ruff clean.

Rutimka:
Second independent review round on the repair commit af2c32c (same protocol: two external read-only reviewers, different models, frozen detached worktree, scope = repair delta only):

- **Reviewer A (sandboxed read-only CLI agent): SHIP-WITH-FIXES** — no CRITICAL/MAJOR; confirmed all three production repairs. One MINOR: the render-crash atomicity regression was a false pass (the classifier monkeypatch from part (a) was still active, so the expected RuntimeError came from the classifier, and the render path was never exercised; the original missing-trace failure mode was also not recreated).
- **Reviewer B (whitelisted read-only tools): SHIP** — all three repairs independently confirmed correct and complete, including that the hook path's thin-forward to the requirements compiler means the digest fix covers the real production route; no regressions; dead-code removal verified genuinely unreachable. Two non-blocking observations (a docstring edge-case overclaim that cannot occur in either production route; the atomicity guarantee relies on an unenforced-but-currently-true read-only contract for the classifier/render helpers).

The MINOR is addressed in f744b85 (tests only, no production change): classifier/render patches now live in separate `monkeypatch.context()` scopes with exception-message matching, plus a new missing-trace (b2) regression asserting the gate marks the result, skips the trace update, and never partially mutates. The strengthened test is mutation-proven: reverting the gate to the pre-fix non-atomic shape makes it fail; the restored gate passes with an identical file hash. Full suite 3043 passed / 1 skipped / 12 xfailed, ruff clean.

PR #589, opened 2026-09-14
fix: resolve macOS APFS concurrency corruption, increase lock timeouts, and add FTS self-healing

ben-wf:
## Summary

This PR addresses database corruption (`database disk image is malformed`) and concurrency contention under multi-process environments on macOS (APFS), adds automatic self-healing for corrupted FTS5 shadow indexes, increases connection busy timeouts, and isolates test runs from user production databases.

## Problem Statement & Root Cause Analysis

In real-world setups running Hermes with concurrent processes (e.g., the supervised Hermes Gateway, the desktop Electron backend `hermes serve`, subagent runners, and CLI invocations), the SQLite database `lcm.db` periodically suffered corruption (`DatabaseError: database disk image is malformed`) during concurrent turn compactions.

Investigation identified four distinct failure modes:

1. **APFS Page Cache Incoherence via `PRAGMA mmap_size` on Darwin:**
   `db_bootstrap.py` configured `PRAGMA mmap_size = 268435456` (256 MiB) unconditionally. On macOS APFS, SQLite memory-mapped reads (`mmap`) combined with multi-process writes cause known page-cache incoherencies between processes sharing the same WAL file, leading SQLite to read out-of-sync B-tree frames and report `database disk image is malformed`.
2. **Hardcoded 5-Second Connection Timeouts:**
   Several stores (`store.py`, `dag.py`, etc.) hardcoded `sqlite3.connect(..., timeout=5.0)`. During deep compactions or heavy write bursts, concurrent readers/writers timed out after only 5 seconds, precipitating transaction rollback races.
3. **Missing FTS Self-Healing in Summary DAG & Shadow Index Desynchronization:**
   While `store.py` had partial handling, `dag.py` had no catch-and-retry logic for FTS index corruption during summary node insertion. Additionally, `repair_external_content_fts()` skipped deep structural rebuilds if external content table row counts matched.
4. **Test Suite Pollution of Production Databases:**
   Running `pytest` without explicit `HERMES_HOME` redirection defaulted to writing to and altering `~/.hermes/lcm.db`, risking production data integrity during local testing.

### 1. Darwin APFS Hardening (`db_bootstrap.py`)
- Default `PRAGMA mmap_size = 0` on `sys.platform == "darwin"` to force POSIX `pread`/`pwrite` through the unified kernel buffer cache, eliminating APFS memory-map cache tearing.
- Allow environment override via `LCM_MMAP_SIZE`.
[...]

PR #607, opened 2026-09-18
fix(lcm): bound session-end lock waits with a real wall clock

trevornk:
## Problem

`_temporary_sqlite_busy_timeout` exists to bound lock waits on gateway-critical paths — `engine.on_session_end()` wraps raw-message ingest in it with `_SESSION_END_BUSY_TIMEOUT_MS = 50`. It sets `PRAGMA busy_timeout` and trusts SQLite to honor it.

Measured against a held WAL writer lock (macOS 27 / CPython 3.11), it does not:

| budget | actual wait | overshoot |
|---|---|---|
| `busy_timeout=50` | ~645–1005 ms | **13–20x** |
| `busy_timeout=300` | ~1856–2289 ms | 6–8x |

A single autocommit `INSERT` overshot identically, so this isn't transaction or engine complexity.

SQLite's busy handler sleeps in a fixed escalating pattern and only compares total elapsed time *between* those sleeps. When short sleeps cost far more than requested — measured here, `1ms -> ~42-54ms`, `2ms -> ~101-116ms` — it sails past a small budget before it next looks at the clock. Whatever the precise cause, the empirical result is what matters: **the PRAGMA does not bound a sub-100ms window**, so a session-end hook meant to yield in 50ms can block the better part of a second on the gateway's critical path.

## Fix

Bound the wait in Python. Poll for the write lock against a `time.monotonic()` deadline with `busy_timeout=0` (each probe fails instantly), then install the caller's `timeout_ms` once the lock looks free.

If the deadline passes with the lock still held, the context manager **does not raise** — it enters the block with `busy_timeout=0` so the caller's own first statement fails immediately with `database is locked`. That keeps the failure attributable to the operation the caller was actually running, preserving its specific diagnostics and recovery branches instead of collapsing every lock loss into one generic error at the wrong layer.

Measured after:

```
50ms budget   -> ~75.8ms   (1.52x)
300ms budget  -> ~370.6ms  (1.24x)
uncontended write -> ~0.99ms
original busy_timeout restored
```

Real `on_session_end` under a held writer lock: **~55.8ms against a 300ms budget**.

## Two approaches measured and rejected

Recorded so they aren't retried:

1. **A progress handler** does not fire while SQLite sleeps in its busy handler — still ~856ms / ~1.89s for the 50ms / 300ms settings.
2. **Retrying `_ingest_messages()` at the Python level** leaks one externalized payload file per failed attempt: a failed attempt rolls back message rows but not the payload write. The correct layer is bounding the lock wait, not retrying the ingest.

## Tradeoffs (called out deliberately)

- The poll interval is **0.0** on purpose. `sleep(0)` yields without arming a timer; any nonzero sleep can blow a 50ms budget in a single poll on this platform. That makes this a bounded busy-spin, suitable only for the small explicitly-budgeted windows this helper serves — noted in the code so it isn't reused for long waits.
- A connection already inside a transaction is left alone, so probing can't disturb caller scope.
- Residual race: the probe releases the lock before yielding, so another writer can take it in between. `timeout_ms` stays installed as a backstop for that window.

## Test Plan

`tests/test_bounded_sqlite_wait.py` — 9 cases: bounded return under a held lock, busy-timeout restoration, uncontended writes, transaction-scope safety, and a negative control. Fails on current `main` (the helper doesn't exist).

With the fix: `tests/test_bounded_sqlite_wait.py tests/test_lcm_engine.py tests/test_db_bootstrap_fts.py` → **808 passed, 1 skipped**.

Wall-clock bounds in the tests use wide tolerances consistent with observed scheduler behavior, so they shouldn't flake on a loaded runner.

## Note on overlap

This touches the same general area as #589 (macOS lock timeouts / APFS hardening) but is a different mechanism: #589 raises timeouts and adds FTS self-healing, while this makes an explicitly-budgeted short wait actually honor its budget. They're complementary — happy to rebase behind #589 if you'd prefer it land first.


trevornk:
Pushed two additional commits to this branch.

**This PR also fixes `test_compaction_telemetry_lock_contention_is_nonblocking`**, which I found failing 3/3 deterministically on pristine `main` while preparing a separate test-reliability PR. That is not a flaky test — it is this exact bug caught from the other side.

`increment_compaction_telemetry` (store.py:1178) wraps its read-modify-write in `_temporary_sqlite_busy_timeout([conn], 100)`, with a comment stating it wants "only a tightly bounded overlap before skipping instead of inheriting the connection's 30s wait". The helper could not deliver that bound, so the telemetry write inherited a long wait on a contended lock and the test's 0.25s assertion blew. With the wall-clock bound it passes in ~0.2-0.3s.

So this closes two independent symptoms of one root cause: the session-end flush and the telemetry hot path. Worth noting the telemetry call site already documented the behavior it expected — the helper just never delivered it.

The second commit moves a one-line tokenizer warm-up into `test_on_session_end_returns_quickly_under_real_sqlite_writer_lock`. Without it the test passes only when sibling files happen to warm the tokenizer first, and **fails 3/3 in isolation**: the first `count_message_tokens` call costs ~254ms cold versus ~0ms warm, most of the 300ms budget, and measures interpreter warm-up rather than the lock wait. The budget is unchanged.

Verification: `test_bounded_sqlite_wait.py + test_lcm_engine.py + test_db_bootstrap_fts.py + test_compaction_telemetry.py` -> **839 passed, 1 skipped**.

